PR #27, opened 2017-04-10
add onMessage props,  postMessage  methods

qingyangmoke:


src/native-common/WebView.tsx
```diff
@@ -65,6 +66,13 @@ export class WebView extends RX.WebView<{}> {
             webView.goForward();
         }
     }
+    
+    postMessage(message: string, targetOrigin?: string) {
+        const webView : RN.WebView = this.refs[WEBVIEW_REF] as RN.WebView;
+        if (webView) {
+            webView.postMessage(message,targetOrigin);
```
erictraut: Please add space after comma for coding consistency.

src/native-common/WebView.tsx
```diff
@@ -41,6 +41,7 @@ export class WebView extends RX.WebView<{}> {
                 scalesPageToFit={ this.props.scalesPageToFit }
                 onError={ this.props.onError }
                 onLoadStart={ this.props.onLoadStart }
+                onMessage={this.props.onMessage}
```
erictraut: Please add spaces within the curly brackets to match other code.

erictraut: This prop is not defined on the WebViewProps object within Types.ts. It's also not clear how this same prop would be supported in the web implementation. In general, all props that are supported in ReactXP should be implemented in all platforms. All props should also be documented in the components/docs directory (in this case, in the components/docs/webview.md file).

ms-deregtd:
onMessage/postMessage is bidirectional on both web and webviews.  The inner and outer sites need to ensure that their origins match for it to be bidirectional, but it's quite doable.  onMessage should be fine on web unless I'm missing something.

src/native-common/WebView.tsx
qingyangmoke: sorry  ,  updated

Possum8:
THANKS

src/native-common/WebView.tsx
erictraut: There's still no space after the comma. Should be "message, targetOrigin".

src/native-common/WebView.tsx
```diff
@@ -41,6 +41,7 @@ export class WebView extends RX.WebView<{}> {
                 scalesPageToFit={ this.props.scalesPageToFit }
                 onError={ this.props.onError }
                 onLoadStart={ this.props.onLoadStart }
+                onMessage={ this.props.onMessage }
```
erictraut: If we're going to support onMessage, then this needs to be added to WebViewProps in Types.ts. And a corresponding implementation is needed for the web implementation of WebView.tsx.

PR #33, opened 2017-04-11
Fix typo in comment in AccessibilityUtil.ts

p01:


src/web/AccessibilityUtil.ts
```diff
@@ -81,9 +81,9 @@ export class AccessibilityUtil extends CommonAccessibiltiyUtil {
         }
         
         // Here we are returning undefined if the above condtion is not met 
-        // as we dont want to poluute the dom with "aria-selected = false" for every falsy condition
+        // as we dont want to polute the dom with "aria-selected = false" for every falsy condition
```
erictraut: Thanks for noticing the typo, but the correct spelling is "pollute". I'll fix it directly. Thanks!

erictraut:
I changed it directly rather than incorporating this PR to provide the correct spelling. Thanks for the contribution!

src/web/AccessibilityUtil.ts
p01: Thanks! Sorry about the typo.

Issue #34, opened 2017-04-11
Why `RX.Component`?


SukantGujar:
Please see my original question on SO: http://stackoverflow.com/questions/43342570/what-features-does-rx-component-class-provide-over-react-component.

george-cz:
You can tell by the sources, that RX.Component is just React.Component. You can use both. This way you don't have to import React.

https://github.com/Microsoft/reactxp/blob/master/src/web/ReactXP.ts#L131

SukantGujar:
Thanks @george-cz. IMO this should be mentioned in the docs as it's an important detail. Are you taking PRs for the documentation?

SukantGujar:
FYI, added a tag for [reactxp on SO](http://stackoverflow.com/questions/tagged/reactxp), please review and add tag-info as appropriate.

george-cz:
@SukantGujar thats not up to me, however feel free to create a PR :)

SukantGujar:
Done! https://github.com/Microsoft/reactxp/pull/36

erictraut:
I merged your PR. Thanks for the contribution!

PR #36, opened 2017-04-11
Add RX.Component and React.Component relationship.

SukantGujar:
https://github.com/Microsoft/reactxp/issues/34


pronebird:
Ehhh what? You do have to import `React` specifically because otherwise you get the following error when using JSX:

>'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.

erictraut:
If you're using TypeScript, you can avoid the need to include React by adding the following line to your tsconfig.json file:
"reactNamespace": "RX"

PR #40, opened 2017-04-12
Fix docs/react_concepts layout bug

599316527:
image

The missing line break leads into incorrect rendering of the manual page.


erictraut:
Thanks for the fix!

PR #43, opened 2017-04-12
Use TypeScript for webpack config in hello world example

mohsen1:
Using TypeScript for Webpack config makes it easier to extend. And yes, this works out of the box!


ms-deregtd:
@mohsen1 Interesting.  I was not aware you could do this. :)  Thanks for helping out!

PR #48, opened 2017-04-16
add a simple sample for using resub and docs for it.

bidanjun:
add a hello-resub sample,for using resub with reactxp,and how to using a custom component.
It's copied from hello-world.


berickson1:
@bidanjun, your example doesn't highlight any of the autoSubscribe functionality provided by ReSub.  Much like the HelloWorld example, it will render static content once.  I'd propose adding something like an auto-incrementing counter that's driven by a ReSub Store (maybe with a start/stop button on the page).  This would be a simple demonstration of ReSub's auto-subscribe functionality in action.  If you'd like help pulling together, I'm willing to assist, just let me know :)

bidanjun:
@berickson1 thanks,and what can I do now?

berickson1:
That would be ok. Ideally the example would utilize store triggering. A show/hide would be a simple example that could demonstrate that.

To have a counter you could use setInterval/clearInterval inside the store (triggered by button onClick functions) that increments a counter and triggers.  I'm happy to add to your example if you like

bidanjun:
@berickson1  I've just add a button, to show/hide the "hello resub" string.,and merged the newest reactxp. thanks for your help ...

bidanjun:
@berickson1  "I'm happy to add to your example if you like", sure,you could do anything as you like ~_~
, I found there is't any example in resub project...even a simple hello-world.

samples/hello-resub/src/tutorials.md
```diff
@@ -0,0 +1,77 @@
...
+                </RX.Text>
+                <Hello >                    
+                </Hello>
+
+
+
+## build it
+    D:\lib\reactxp\samples\hello-resub>yarn run web-watch
+    a compile error here:
+    ERROR in [at-loader] ./node_modules/resub/dist/src/StoreBase.d.ts:13:20
```
Theadd: What about suggesting a viable workaround for the ones that just stepped into reactxp/typescript? (like me)

After taking a look at [tsconfig.json compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html) I got rid of it by setting this options in tsconfig.json
```diff
+    "noImplicitAny": false,
+    "skipLibCheck": true,
```
> I've not looked into which one of the above two changes is the error not being checked or ignored.

**BUT** I'd like to have those checks being applied in my own code sooner or later. Is there a better workaround / correct solution?

PR #58, opened 2017-04-19
Fix RX.Modal sample usage example.

9Y5:
Getting this typescript error when using `<RX.Modal>`:
```
JSX element type 'RX.Modal' does not have any construct or call signatures.
```


erictraut:
Thanks for the fix!

PR #59, opened 2017-04-20
Fixes for React Native 0.42.0

rigdern:
These are the changes I made to get ReactXP working with React Native 0.42.0.


src/native-common/Button.tsx
```diff
@@ -110,6 +134,12 @@ export class Button extends RX.Button<{}> {
         }
     }
 
+    public setNativeProps(nativeProps: RN.ViewProps) {
```
erictraut: No need to explicitly mark it public. It is assumed to be so. We don't use public elsewhere in ReactXP.

src/native-common/ScrollView.tsx
```diff
@@ -40,9 +40,11 @@ export class ScrollView extends ViewBase<Types.ScrollViewProps, {}> implements R
         // TODO: #737970 Remove special case for UWP when this bug is fixed. The bug
         //   causes you to have to click twice instead of once on some pieces of UI in
         //   order for the UI to acknowledge your interaction.
-        const keyboardShouldPersistTaps = RN.Platform.OS === 'windows' ?
-            true :
-            this.props.keyboardShouldPersistTaps;
+        const keyboardShouldPersistTaps = (
+            RN.Platform.OS === 'windows' ? 'always'
```
erictraut: Please include parens so the order of evaluation is clear.

src/native-common/Styles.ts
```diff
@@ -20,6 +20,13 @@ const forbiddenProps: string[] = [
     'cursor'
 ];
 
+// React Native styles that ReactXP doesn't expose.
```
erictraut: I'd prefer not to expose these. They needlessly complicate the model and introduce ambiguity if both flex and these are presented within a combined style. They also slow down style adaptation, which is already a measurable source of overhead.

src/typings/react-native.d.ts
```diff
@@ -242,8 +242,8 @@ declare module 'react-native' {
 
         contentContainerStyle?: StyleRuleSet | StyleRuleSet[];
         horizontal?: boolean;
-        keyboardDismissMode?: string; // enum( 'none', 'interactive', 'on-drag' )
-        keyboardShouldPersistTaps?: boolean;
+        keyboardDismissMode?: 'none' | 'interactive' | 'on-drag';
+        keyboardShouldPersistTaps?: string; //'always' | 'never' | 'handled';
```
erictraut: Why not use strongly-typed values here as well?

src/native-common/Styles.ts
rigdern: ReactXP doesn't expose `flexGrow`, `flexShrink`, or `flexBasis` to consumers of the library which means this change doesn't introduce any flex ambiguity issues.

However, this change does map `flex` to these three properties exactly the same way as ReactXP does on web. You are right that this will cause style adaptation overhead although I'm not familiar with the cost.

The reason that this mapping was introduced is that React Native changed the meaning of `flex`. The new meaning broke a bunch of layouts so I decided to map `flex` to grow, shrink, and basis the same way ReactXP does on web in order to maintain layout consistency between React Native and web. I can provide more details if you are interested.

src/native-common/ScrollView.tsx
steida: https://github.com/prettier/prettier?

rigdern: @steida that's a good suggestion but I don't think prettier supports TypeScript yet (prettier/prettier#13).

PR #61, opened 2017-04-20
Fix #56 - fadeDuration should be a prop, not a style

### LINKED ISSUE #56: fadeDuration added by default on Android for RX.Image is resulting in error
```
ExceptionsManager.js:71 Warning: Failed prop type: Invalid props.style key `fadeDuration` supplied to `RCTImageView`.
Bad object: {
  "width": 360,
  "height": 360,
  "overflow": "hidden",
  "flex": 0,
  "borderRadius": 5,
  "fadeDuration": 0
}
```

Here's the code I'm creating my ImageStyle with

```
artistImageStyle =  RX.Styles.createImageStyle({
            borderRadius: 5,
            height: width,
            width: width
        });
```

And here's my Image element.

```
<RX.Image source={track.getHighResImage()} style={artistImageStyle}/>
```

Looking through the source, I believe this is where fadeDuration is being set

https://github.com/Microsoft/reactxp/blob/0e195c9cac45b15bcee324945720b91d1249b6d7/src/android/Image.tsx

I'm running the build through Android Studio and testing on my Samsung Galaxy S6.

Cheers

berickson1:


src/native-common/Image.tsx
```diff
@@ -46,6 +46,10 @@ export class Image extends RX.Image<{}> {
     private _nativeImageWidth: number;
     private _nativeImageHeight: number;
 
+    protected _getAdditionalProps(): { [key: string]: string|number } {
```
rigdern: Should we make the return value of type `ImageProps` with everything optional? I think that would be expressed as `Partial<ImageProps>`.

rigdern:
I left one comment. Looks good to me.

PR #68, opened 2017-04-23
Support long press callback for the Link component

hccoelho:
This change allow to pass a longpress callback with a url as parameter (e.g. it can be used for menu options of the url)

Please review @erictraut @berickson1 @deregtd

### DIFF
--- a/src/web/Link.tsx
+++ b/src/web/Link.tsx
@@ -90,9 +90,7 @@ export class Link extends RX.Link<void> {
     private _onClick = (e: React.MouseEvent) => {
         e.stopPropagation();
 
-        if (this._ignoreClick) {
-            this._ignoreClick = false;                        
-        } else if (this.props.onPress) {
+        if (this.props.onPress) {
             e.preventDefault();
             this.props.onPress(e, this.props.url);
         }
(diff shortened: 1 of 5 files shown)


src/web/Link.tsx
```diff
@@ -84,10 +92,28 @@ export class Link extends RX.Link<void> {
...
+                if (this.props.onLongPress) {
+                    this.props.onLongPress(e, this.props.url);
+                }
+            }, _longPressTime);
+        }
+    }    
+
+    private _onMouseUp = (e: Types.SyntheticEvent) => {
+        if (this._longPressTimer) {
+            window.clearTimeout(this._longPressTimer);
```
erictraut: It's good form to set this._longPressTimer to undefined after clearing it.

src/web/Link.tsx
```diff
@@ -37,7 +37,14 @@ const _styles = {
     }
 };
 
+const _longPressTime = 1000;
+
 export class Link extends RX.Link<void> {
+    
+    private _lastMouseDownEvent: Types.SyntheticEvent;
```
erictraut: This doesn't appear to be used.

src/web/Link.tsx
```diff
@@ -37,7 +37,14 @@ const _styles = {
     }
 };
 
+const _longPressTime = 1000;
+
 export class Link extends RX.Link<void> {
+    
+    private _lastMouseDownEvent: Types.SyntheticEvent;    
+    private _longPressTimer: number;    
+    private _ignoreClick = false;
```
erictraut: Same here.

src/web/Link.tsx
hccoelho: Good call.

src/web/Link.tsx
hccoelho: Grrrr. Some left overs from another thing I wanted to include but not right now.

hccoelho:
@erictraut Thanks for the good suggestions and for catching some leftovers. Updated the PR. Let me know if there is anything else.

PR #69, opened 2017-04-24
Expose setAccessibilityFocus from Native-Common AccessibilityUtil and expose setting accessibility focus on RX.TextInput.

eulphean:
This change exposes a way to set accessibility focus on RX.TextInput. This also changes the way we had AccessibilityUtil implementations individually for each platform. That becomes an overhead for each component that wants to leverage those platform helpers. We need to create a platform specific copy of each component and leverage the platform specific AccessibilityUtil. 

With this change, we let Native-Common AccessibilityUtil do all the work for calling any platform specific methods.


src/native-common/AccessibilityUtil.ts
```diff
@@ -54,6 +60,28 @@ const componentTypeMap = {
 };
 
 export class AccessibilityUtil extends CommonAccessibilityUtil {
+    // Native platform specific instance for AccessibilityUtil. 
+    private _instance: typeof AndroidAccessibilityUtil | typeof iOSAccessibilityUtil;
```
berickson1: These should both be the same type. You should define an interface (or abstract base class) that both iOS and Android need to implmenet

src/native-common/AccessibilityUtil.ts
```diff
@@ -54,6 +60,28 @@ const componentTypeMap = {
...
+            case 'android':
+                this._instance = AndroidAccessibilityUtil;
+                break;
+
+            case 'ios':
+                this._instance = iOSAccessibilityUtil;
+                break;
+
+            default: 
+                assert(false, 'Unknown platform.');
```
berickson1: uwp isn't an unknown platform inside ReactXP

src/native-common/AccessibilityUtil.ts
```diff
@@ -53,7 +60,38 @@ const componentTypeMap = {
     [Types.AccessibilityTrait.Radio_button_unchecked]: 'radiobutton_unchecked'
 };
 
+// Platform specific helpers exposed through Native-Common AccessibilityUtil. 
+export abstract class NativeHelpers {
```
sevachon: Consider giving a more specific name to this class to be derived by each platform. Maybe someone like CommonNativeAccessibilityUtil which would be in line with the pattern used currently or NativeAccessibilityHelpers. NativeHelpers is very generic and makes it's difficult to the reader to understand what it's about on its own.

src/android/AccessibilityUtil.ts
```diff
@@ -11,9 +11,9 @@ import React = require('react');
 import RN = require('react-native');
 
 import Accessibility from '../native-common/Accessibility';
-import { AccessibilityUtil as CommonAccessibilityUtil } from '../native-common/AccessibilityUtil';
+import { NativeHelpers } from '../native-common/AccessibilityUtil';
```
berickson1: I think you have a potential circular dependency here. native-common/AccessibilityUtil -> android/AccessibilityUtil -> native-common/AccessibilityUtil

src/ios/AccessibilityUtil.ts
```diff
@@ -0,0 +1,22 @@
+/**
+* iOSAccessibilityUtil.ts
```
erictraut: The comment should match the file name.

src/native-common/AccessibilityUtil.ts
```diff
@@ -9,8 +9,15 @@
 */
 
 import _ = require('./lodashMini');
+import assert = require('assert');
+import React = require('react');
+import RN = require('react-native');
 
 import { AccessibilityUtil as CommonAccessibilityUtil } from '../common/AccessibilityUtil';
+import AndroidAccessibilityUtil from '../android/AccessibilityUtil';
```
erictraut: Why are you importing platform-specific files within a common file? This is an anti-pattern and will create dependency cycles. Platform-specific files should import native-common, not the other way around.

src/native-common/AccessibilityUtil.ts
erictraut: I agree with Seb. NativeHelpers is too generic a name to export.

src/native-common/AccessibilityUtil.ts
berickson1: While this is an anti-pattern, it's beneficial to us.  The idea behind this is that pretty much every UI component needs to talk to AccessibilityUtils in some way, shape or form.  That leaves a few options:

1) Extend every UI component (that needs accessibility) for every platform so that it imports the platform specific AccessibiilityUtils and calls the function
2) Provide some sort of override/shim at the RX platform creation (something AccessibilityUtils.setPlatformHelperFunctions that pipes to this._impl that's in this PR)
3) Have native-common/AccessibilityUtils decide which impl to use on init and import platform specific implementations

#1 adds a bunch of extra code and another layer of abstract classes
#2 is somewhat obscure and non-obvious to someone working in the codebase
#3 isn't ideal but seems somewhat better than the other options.

Do you have any alternative solutions?

eulphean:
@erictraut @berickson1 

I have a new iteration addressing the comments.

src/android/AndroidAccessibilityUtil.ts
```diff
@@ -1,5 +1,5 @@
 /**
-* AccessibilityUtil.ts
+* AndroidAccessibilityUtil.ts
```
erictraut: Why did you rename this file to start with "Android"? It's already in the Android directory, so adding this is redundant. It's also inconsistent with all of the other code in ReactXP. I'd prefer if you went back to the previous name for consistency.

src/ios/iOSAccessibilityUtil.ts
```diff
@@ -0,0 +1,22 @@
+/**
+* iOSAccessibilityUtil.ts
```
erictraut: Once again, the file name shouldn't contain iOS since that's redundant. It's already in a folder called iOS. Same with the class name it exports.

src/native-common/AccessibilityUtil.ts
```diff
@@ -9,8 +9,15 @@
 */
 
 import _ = require('./lodashMini');
+import assert = require('assert');
+import React = require('react');
+import RN = require('react-native');
+
+import { AccessibilityUtil as CommonAccessibilityUtil, PlatformAccessibilityHelpers } from '../common/AccessibilityUtil';
+import AndroidAccessibilityUtil from '../android/AndroidAccessibilityUtil';
```
erictraut: This is still inverted. The "native common" implementation should not be importing the platform-specific implementations. It should be the other way around. The platform-specific implementations should be importing the common and overriding the parts that are necessary. Doing it the other way is inconsistent with all other parts of ReactXP and has the potential for creating dependency cycles.

src/native-common/AccessibilityUtil.ts
```diff
@@ -54,6 +61,32 @@ const componentTypeMap = {
...
+    // Specific native platform instance for AccessibilityUtil. 
+    private _instance: PlatformAccessibilityHelpers;
+
+    constructor() {
+        super();
+
+        // Set AccessibilityUtil instance based on the platform. 
+        switch (RN.Platform.OS) {
+            case 'android':
+                this._instance = AndroidAccessibilityUtil;
```
erictraut: Instantiation of platform-specific modules should happen only in the ReactXP.ts files that are specific to each platform. You shouldn't be doing it here in this native-common module. By doing so, you're pulling in all platform implementation on all platforms. The iOS version of ReactXP shouldn't contain any Android-specific code and vice versa. By instantiating the platform-specific code in ReactXP.ts, we are able to strip out the code for other platforms. By doing it here, you defeat that mechanism.

src/windows/WindowsAccessibilityUtil.ts
```diff
@@ -0,0 +1,20 @@
+/**
+* WindowsAccessibilityUtil.ts
```
erictraut: See note above. This file and its exported class shouldn't start with "Windows".

eulphean:
Thanks for the feedback, guys. I just pushed a latest iteration that should address the concerns around this. The problem with using a flat inheritance model how RX.Linking component does is that platform specific implementations need to be initialized inside Native-Common AccessibilityUtil in order for it to consume it. This brings up the concern of importing platform specific code into Native-Common instance that defeats the purpose of ReactXp's message. 

By letting initialization path in ReactXp poke the platform specific instance inside native-common AccessibilityUtil, we solve that problem. 

I used Seb's suggestion for naming the platform specific accessibility helper. Goal was to keep the suffix 'Util' to bear resemblance and use the term 'Native' to drive the difference between the two classes.

erictraut:
Looking better. I realize this approach is still a bit convoluted, but I think it's much better than the previous approach. Once @berickson1 reviews and approves the change, he or I can merge it in.

src/android/AccessibilityNativeUtil.ts
```diff
@@ -1,5 +1,5 @@
 /**
-* AccessibilityUtil.ts
+* AccessibilityNativeUtil.ts
```
berickson1: Can't this still be called AccessibilityUtil (Native is redundant since it's in the android folder

src/android/ReactXP.ts
```diff
@@ -55,6 +58,11 @@ const _defaultViewStyle = StylesImpl.createViewStyle({
 });
 ViewBase.setDefaultViewStyle(_defaultViewStyle);
 
+// Initialize Android implementation of platform accessibility helpers inside the singleton
+// instance of native-common AccessibilityUtil. This is to let native-common components access
+// platform specific APIs through native-common implementation itself. 
+AccessibilityUtil.setAccessibilityNativeUtil(AccessibilityNativeUtil);
```
berickson1: Use AccessibilityUtilImpl

PR #73, opened 2017-04-26
Displaying non-selectable texts as pseudo elements

mshoho:
Displaying non-selectable texts as pseudo elements to avoid copying them to clipboard on web.

### DIFF
--- a/src/web/Text.tsx
+++ b/src/web/Text.tsx
@@ -15,6 +15,14 @@ import RX = require('../common/Interfaces');
 import Styles from './Styles';
 import Types = require('../common/Types');
 
+if (typeof document !== 'undefined') {
+    const textAsPseudoElement = '[data-text-as-pseudo-element]::before { content: attr(data-text-as-pseudo-element); }';
+    const style = document.createElement('style');
+    style.type = 'text/css';
+    style.appendChild(document.createTextNode(textAsPseudoElement));
+    document.head.appendChild(style);
+}
+
 const _styles = {
     defaultStyle: {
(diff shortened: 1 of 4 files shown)


src/web/Text.tsx
```diff
@@ -15,6 +15,14 @@ import RX = require('../common/Interfaces');
 import Styles from './Styles';
 import Types = require('../common/Types');
 
+if (typeof document !== 'undefined') {
```
erictraut: Please add a comment that explains what this does. It's sufficiently non-obvious that a comment is needed. :)

src/web/Text.tsx
```diff
@@ -15,6 +15,14 @@ import RX = require('../common/Interfaces');
...
+    const textAsPseudoElement = '[data-text-as-pseudo-element]::before { content: attr(data-text-as-pseudo-element); }';
+    const style = document.createElement('style');
```
erictraut: This dynamic modification of the style is clever. I think we should use this same technique to eliminate the manual "box-sizing" override step that's currently needed for all ReactXP apps: https://microsoft.github.io/reactxp/docs/using-reactxp.html. If you feel like it, you're welcome to make this change as well -- and remove the step from the documentation.

mshoho: Is RootView.tsx the best place to add it?

erictraut:
Yeah, that makes sense.

src/web/Text.tsx
mshoho: Done.

PR #77, opened 2017-04-30
[BUGFIX] Typo in ReactXP.ts for Windows

jackTheRipper:
**Do**

When trying to build my application for Windows with ReactXP I face the following issue:

err-reactxp

I found out that it happens because of the path. 

So here is a PR for helping. Now having my App up and running on the web, Android and Windows :D


erictraut:
Thanks for the fix, @jackTheRipper.

PR #78, opened 2017-05-01
Exposing the maxContentSizeMultiplier prop for RN.Text and RN.TextInput.

rumit91:
Whenever a user changes the system font size to its maximum allowable setting most React Native apps that allow font scaling become unusable. Ideally React Native apps would mimic the behavior of the iMessage (iOS) app where the font size used for non-body text (header, navigational elements, etc) is capped while the body text (text in the message bubbles) is allowed to grow. In this PR I expose the `maxContentSizeMultiplier` prop on `<Text>`, `<TextInput>`, and `<Link>`. Developers can overwrite the default maxContentSizeMultiplier by calling the static method `UserInterface. setMaxContentSizeMultiplier()` or they can just choose to set the `maxContentSizeMultiplier` on the desired elements.


package.json
```diff
@@ -20,7 +20,7 @@
     "react-addons-perf": "^15.4.1",
     "react-addons-test-utils": "^15.4.1",
     "react-dom": "^15.4.1",
-    "react-native": "^0.42.0",
+    "react-native": "git+https://skype.visualstudio.com/DefaultCollection/TPS/_git/react-native#timale/fontScalingLimit",
```
erictraut: You can't change the public reactxp to point to a private, internal react-native version. Just leave the version as-is. Internally, the Skype repo can refer to the private react-native branch.

src/android/Text.tsx
```diff
@@ -40,6 +40,7 @@ export class Text extends CommonText {
                 importantForAccessibility={ importantForAccessibility }
                 numberOfLines={ this.props.numberOfLines === 0 ? null : this.props.numberOfLines }
                 allowFontScaling={ this.props.allowFontScaling }
+                maxFontSizeMultiplier={ this.props.maxFontSizeMultiplier }
```
erictraut: Why wouldn't we also expose a minFontSizeMultiplier?

src/android/Text.tsx
```diff
@@ -52,6 +53,10 @@ export class Text extends CommonText {
     focus() {
         AccessibilityUtil.setAccessibilityFocus(this);
     }
+
+    static setDefaultMaxFontSizeMultiplier(maxFontSizeMultiplier: number): void {
```
erictraut: This shouldn't be exposed as a static method on Text. It should be a method on the UserInterface namespace.

src/android/Text.tsx
erictraut: For consistency with UserInterface (and react native) terminology, this should be called maxContentSizeMultiplier, not minFontSizeMultiplier.

erictraut: Do we really need to expose an override of the default max on a per-Text basis? This seems like it might be overkill.

src/native-common/TextInput.tsx
```diff
@@ -220,6 +221,10 @@ export class TextInput extends RX.TextInput<TextInputState> {
     setValue(value: string): void {
         this._onChangeText(value);
     }
+
+    static setDefaultMaxFontSizeMultiplier(maxFontSizeMultiplier: number): void {
```
erictraut: I don't think we need a different default for Text and TextInput. As mentioned above, this should be collapsed to a single API - UserInterface.setDefaultMaxContentSizeMultiplier.

src/android/Text.tsx
erictraut: The documentation for Text, TextInput and UserInterface need to be updated as well.

package.json
rumit91: Yep will remove for sure. Was just using this for testing.

src/android/Text.tsx
rumit91: I think we need this to be able to override the default in specific instances. For example, we may want the `maxContentSizeMultiplier` to be much higher for text inside message bubbles vs the text in the navigational elements.

rumit91: I didn't add a min, because the user changing the system font to the minimum didn't seem to cause any issues in RN apps, but do you think it would still be valuable for me to add?

rumit91: Changed the var name.

src/android/Text.tsx
rumit91: Moved to UserInterface.

src/native-common/TextInput.tsx
rumit91: Agreed. Moved to UserInterface.

src/android/Text.tsx
rumit91: Thanks for reminding me. I added the docs.

docs/docs/apis/userinterface.md
```diff
@@ -57,5 +67,10 @@ dismissKeyboard(): void;
 // app is running
 contentSizeMultiplierChangedEvent: SubscribableEvent<
     (multiplier: number) => void>();
+
+// Triggered when the max content size multiplier changes while
+// the app is running.
+maxContentSizeMultiplierChangedEvent: SubscribableEvent<
```
erictraut: We don't need this event. The max content size multiplier is under the control of the app, so it can track this change itself. The reason I added an event for contentSizeMultiplierChanged is that it's under control of the OS (and ultimately the user), so the app needs to be informed when it changes.

docs/docs/components/text.md
```diff
@@ -29,12 +29,15 @@ accessibilityTraits: AccessibilityTrait | AccessibilityTrait[] = undefined;
...
-// Keyboard tab order
-tabIndex: number = undefined;
-
 // Should fonts be scaled according to system setting?
 allowFontScaling: boolean = true; // Android and iOS only
 
+// Should the scale multiplier be capped when allowFontScaling is set to true?
+// The default of 0 indicates that the compoent should obey the global setting 
+// in UserInterface which by default is uncapped.
+// Note: Most versions of React Native don’t support this interface.
```
erictraut: See above.

docs/docs/components/textinput.md
```diff
@@ -43,6 +43,12 @@ keyboardAppearance: 'default' | 'light' | 'dark';
 // On-screen keyboard type to display
 keyboardType: 'default' | 'numeric' | 'email-address' | 'number-pad';
 
+// Should the scale multiplier be capped when allowFontScaling is set to true?
+// The default of 0 indicates that the compoent should obey the global setting 
+// in UserInterface which by default is uncapped.
+// Note: Most versions of React Native don’t support this interface.
```
erictraut: See above.

docs/docs/components/link.md
```diff
@@ -11,6 +11,15 @@ This component displays a hyperlink. On the web, it translates to an &lt;a&gt; t
 
 ## Props
 ``` javascript
+// Should fonts be scaled according to system setting?
+allowFontScaling: boolean = true; // Android and iOS only
+
+// Should the scale multiplier be capped when allowFontScaling is set to true?
+// The default of 0 indicates that the compoent should obey the global setting 
+// in UserInterface which by default is uncapped.
+// Note: Most versions of React Native don’t support this interface.
```
erictraut: Rather than saying "Most versions of React Native", please change to "Older versions...".

src/android/Text.tsx
```diff
@@ -48,10 +49,6 @@ export class Text extends CommonText {
             </RN.Text>
         );
     }
-
```
erictraut: Did you intend to delete this, or was it a merge error? Amay recently added this code.

rumit91: I deleted it because it caused merge conflicts. I believe Amay removed it from master recently with this commit: https://github.com/Microsoft/reactxp/commit/cff4b51bd88c6c636135ecbd73b4ccaf0ed2c1cd

docs/docs/apis/userinterface.md
rumit91: Removed from the code and the docs.

docs/docs/components/link.md
rumit91: Done.

PR #80, opened 2017-05-05
Expose className property for text input (for web purpose)

marketaha:
Expose className property for text input to improve element locating on the web that is needed in some cases 
(e. g. to style pseudo-elements like a placeholder, because placeholderTextColor property does not work on the web)


erictraut:
This change violates the design principles of ReactXP. If you want to write web-specific code, you should stick with React. ReactXP is specifically for modules that work cross-platform.

marketaha:
Good point, I can actually workaround this web-specific problem by using React. Thanks, will close this PR.

PR #88, opened 2017-05-11
Add “display” to forbiddenProps

Iamsamwhoami:
In order to center align everything inside a `<ScrollView />` (which has a default of `display: block;`) I need it to be `display: flex;`, this makes sure that we never expose `display` to RN clients (which makes it crash).


erictraut:
This change shouldn't be necessary. The display:block is an internal implementation detail of ScrollView. I'd like to understand what problem you've faced and find another way to address it.

PR #89, opened 2017-05-11
Pass “secureTextEntry” down to RN

Iamsamwhoami:
`secureTextEntry` didn't work and then I looked at the code and it seem's like it's never passed down to RN, this fixes that.

### DIFF
--- a/src/native-common/TextInput.tsx
+++ b/src/native-common/TextInput.tsx
@@ -77,6 +77,7 @@ export class TextInput extends RX.TextInput<TextInputState> {
                 onBlur={ this._onBlur }
                 onScroll={ this._onScroll }
                 selection={{ start: this._selectionStart, end: this._selectionEnd}}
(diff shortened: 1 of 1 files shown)


src/native-common/TextInput.tsx
```diff
@@ -77,6 +77,7 @@ export class TextInput extends RX.TextInput<TextInputState> {
                 onBlur={ this._onBlur }
                 onScroll={ this._onScroll }
                 selection={{ start: this._selectionStart, end: this._selectionEnd}}
+                secureTextEntry={this.props.secureTextEntry}
```
erictraut: Please use spacing consistent with the other code in ReactXP.

erictraut:
Good find. Thanks for fixing. I ended up checking in my own fix - a small variation of yours that uses spacing that's consistent with the other code in ReactXP.

PR #90, opened 2017-05-11
[Accessibility] Expose the implementation for setting accessibility focus programmatically on iOS. Also, expose .focus() API for RX.Button.

eulphean:
This change enables the react-native work to programmatically set accessibility focus for iOS. Since that change will take some time to be contributed back to open source React-Native, this API will not work as expected. 

This change also exposes the .focus() behavior for RX.Button.

### DIFF
--- a/src/ios/Accessibility.ts
+++ b/src/ios/Accessibility.ts
@@ -63,7 +63,9 @@ export class Accessibility extends NativeAccessibility {
 
         // Some versions of RN don't support this interface.
         if (RN.AccessibilityInfo) {
-            RN.AccessibilityInfo.announceForAccessibility(announcement);
+            if (RN.AccessibilityInfo.announceForAccessibility) {
+                RN.AccessibilityInfo.announceForAccessibility(announcement);
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/ios/Accessibility.ts
```diff
@@ -63,7 +63,9 @@ export class Accessibility extends NativeAccessibility {
 
         // Some versions of RN don't support this interface.
         if (RN.AccessibilityInfo) {
```
berickson1: merge the two if blocks

src/ios/AccessibilityUtil.ts
```diff
@@ -15,7 +15,9 @@ import { AccessibilityPlatformUtil } from '../common/AccessibilityUtil';
 
 export class AccessibilityUtil extends AccessibilityPlatformUtil {
     setAccessibilityFocus(component: React.Component<any, any>): void {
-        // NO-OP
+        if (RN.AccessibilityInfo && Accessibility.isScreenReaderEnabled()) {
```
berickson1: You can merge these two checks

PR #92, opened 2017-05-12
Switch PropTypes import from React to 'prop-types' package

psegalen:
This is a try since I'm quite new to TypeScript, I may have done something wrong but it compiles.

### DIFF
--- a/src/typings/prop-types.d.ts
+++ b/src/typings/prop-types.d.ts
@@ -1,7 +1,33 @@
 declare module 'prop-types' {
-    import React = require('react');
+    // Type definitions for prop-types 15.5
+    // Project: https://github.com/reactjs/prop-types
+    // Definitions by: DovydasNavickas <https://github.com/DovydasNavickas>
+    // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+    // TypeScript Version: 2.2
 
-    export const bool: {
-        isRequired: React.Validator<any>
+    export type Validator<T> = (object: T, key: string, componentName: string, ...rest: any[]) => Error | null;
+
+    // psegalen: modified original declaration because of
+    // "error TS2312: An interface may only extend a class or another interface."
+    export interface Requireable<T> {
+        isRequired: Validator<T>;
     }
+
+    export type ValidationMap<T> = {[K in keyof T]?: Validator<T> };
+
+    export const any: Requireable<any> & Validator<any>;
+    export const array: Requireable<any> & Validator<any>;
+    export const bool: Requireable<any> & Validator<any>;
+    export const func: Requireable<any> & Validator<any>;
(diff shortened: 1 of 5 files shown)


src/typings/prop-types.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module 'prop-types' {
+    import React = require('react');
+
+    export const bool: {
+        isRequired: React.Validator<any>
```
george-cz: This is probably insufficient and will strictly work only for `PropTypes.bool`, any other prop type will fail

psegalen: I've exported bool only because it's the only PropType used in the project for now but you're right, I'll use what you mentioned in the EDIT.

berickson1: You should actually be able to add @types/prop-types to package.json and get typing that way

psegalen: Since the whole project is kind of huge, I didn't wish to load everything for this.
Is the way I did it in my second commit acceptable?

psegalen: Plus: I ran into a TS compiler error when trying to tsc "error TS2312: An interface may only extend a class or another interface." so I had to modify the code a little.

ms-deregtd:
Thanks for your submission!

Can you please elaborate on what the advantage of this change is?  We're trying to keep the footprint of ReactXP as small as possible, and this seems to be bringing in an extra package to do the same thing that's currently available built into React, so I'm trying to understand what the tradeoffs are here.  Thanks!

bmavity:
PropTypes is deprecated as of 15.5.0.

https://github.com/facebook/react/blob/master/CHANGELOG.md

ms-deregtd:
Aha, gotcha.  It looks like it's designed to be mostly pulled out in production builds as well, so it doesn't actually add much size.

psegalen:
Indeed, the purpose of this PR is to get rid of the deprecated warning!

psegalen:
I forgot to copy / paste the body of the issue I filed earlier in the PR, sorry!

src/typings/prop-types.d.ts
berickson1: FYI, Limiting the typings doens't actually limit the scope of what's imported, it just limits what the TS compiler will let you use.
I suspect if you remove all references to React.PropTypes internally in react.d.ts the warning will go away.
Ideally we would be consuming all our typings files directly from @types repo

psegalen: I tried to npm install **@types/prop-types** but it won't compile because of error TS2312, that's why I modified the code to convert the interface extending the type for an implementation of both the interface and the type on each property.

If I remove all references to React.PropTypes in react.d.ts, I suspect that the references in Image.tsx, Text.tsx and View.tsx won't compile, am I wrong?

PR #101, opened 2017-05-18
Fix and update web Navigator component

RyanThomas73:
The web Navigator component will error when used because presentedIndex and transitionFromIndex are not correctly initialized.

This pull request also updates the web Navigator component to honor the initialRouteStack or initialRoute props if they are provided and to pass the navigator instance as a second parameter when calling the renderScene(...) handler.

### DIFF
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -967,7 +967,7 @@ export type CustomNavigatorSceneConfig = {
 };
 
 export interface NavigatorProps extends CommonProps {
-    renderScene?: (route: NavigatorRoute) => JSX.Element;
+    renderScene?: (route: NavigatorRoute, navigator: RX.Navigator<any>) => JSX.Element;
     navigateBackCompleted?: () => void;
     // NOTE: Arguments are only passed to transitionStarted by the experimental navigator
     transitionStarted?: (progress?: RX.AnimatedValue,
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -976,6 +976,8 @@ export interface NavigatorProps extends CommonProps {
     transitionCompleted?: () => void;
     cardStyle?: ViewStyleRuleSet;
     children?: ReactNode;
+    initialRouteStack?: Array<NavigatorRoute>;
+    initialRoute?: NavigatorRoute;
 }
 
 //
--- a/src/web/Navigator.tsx
+++ b/src/web/Navigator.tsx
@@ -139,7 +139,7 @@ export class Navigator extends RX.Navigator<NavigatorState> {
 
         // handle optional initialRouteStack or initialRoute props
         let { initialRouteStack, initialRoute } = initialProps;
-        if(!Array.isArray(initialRouteStack)) {
+        if (!Array.isArray(initialRouteStack)) {
             initialRouteStack = [];
         }
 
--- a/src/web/Navigator.tsx
+++ b/src/web/Navigator.tsx
@@ -137,10 +137,21 @@ export class Navigator extends RX.Navigator<NavigatorState> {
     constructor(initialProps?: Types.NavigatorProps) {
         super(initialProps);
 
+        // handle optional initialRouteStack or initialRoute props
+        let { initialRouteStack, initialRoute } = initialProps;
+        if(!Array.isArray(initialRouteStack)) {
+            initialRouteStack = [];
+        }
+
+        if(initialRouteStack.length === 0 && initialRoute) {
+            initialRouteStack.push(initialRoute);
+        }
+        
+
         // Default navigator state
         this.state = {
             sceneConfigStack: [],
-            routeStack: [],
+            routeStack: initialRouteStack,
             transitionQueue: []
         };
     }
--- a/src/web/Navigator.tsx
+++ b/src/web/Navigator.tsx
@@ -146,10 +146,6 @@ export class Navigator extends RX.Navigator<NavigatorState> {
         if(initialRouteStack.length === 0 && initialRoute) {
             initialRouteStack.push(initialRoute);
         }
-
-        // need to initialize these or it will error 
-        (this.state as any).presentedIndex = 0;
-        (this.state as any).transitionFromIndex = null;
         
 
         // Default navigator state
--- a/src/web/Navigator.tsx
+++ b/src/web/Navigator.tsx
@@ -364,7 +364,7 @@ export class Navigator extends RX.Navigator<NavigatorState> {
                 ref={ 'scene_' + index }
                 style={ styles }
             >
-                { this.props.renderScene(route) }
+                { this.props.renderScene(route, this) }
             </View>
         );
     }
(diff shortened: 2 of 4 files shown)


src/web/Navigator.tsx
```diff
@@ -137,10 +137,23 @@ export class Navigator extends RX.Navigator<NavigatorState> {
     constructor(initialProps?: Types.NavigatorProps) {
         super(initialProps);
 
+        // handle optional initialRouteStack or initialRoute props
+        let { initialRouteStack, initialRoute } = initialProps;
+        if(!Array.isArray(initialRouteStack)) {
```
erictraut: Please add spaces after "if" for consistency with rest of code in ReactXP.

src/common/Types.ts
```diff
@@ -967,7 +967,7 @@ export type CustomNavigatorSceneConfig = {
 };
 
 export interface NavigatorProps extends CommonProps {
-    renderScene?: (route: NavigatorRoute) => JSX.Element;
+    renderScene?: (route: NavigatorRoute, navigator: RX.Navigator<any>) => JSX.Element;
```
erictraut: This is a change in the public interface — something that will force existing apps to change their code to avoid getting TypeScript compile errors. As such, we need a very good justification for making this change. I'm not sure what you're trying to accomplish by passing the navigator back to the renderScene callback. The user should already have access to the navigator (through refs) if it wants it.

src/common/Types.ts
```diff
@@ -976,6 +976,8 @@ export interface NavigatorProps extends CommonProps {
     transitionCompleted?: () => void;
     cardStyle?: ViewStyleRuleSet;
     children?: ReactNode;
+    initialRouteStack?: Array<NavigatorRoute>;
```
erictraut: Why do you want to pass the initial route stack as props? In general, I'm not a fan of props that specify an "initial" value because they go against the React philosophy of updating existing components. It's not clear, for example, what happens if the caller changes the initialRouteStack and initialRoute on subsequent renders. Should they be honored or not?

If we do keep these, they need to apply to both the native and web implementations. It looks like you've implemented it only on the web currently.

src/web/Navigator.tsx
```diff
@@ -137,10 +137,23 @@ export class Navigator extends RX.Navigator<NavigatorState> {
...
+        if(initialRouteStack.length === 0 && initialRoute) {
+            initialRouteStack.push(initialRoute);
+        }
+        
+
         // Default navigator state
         this.state = {
             sceneConfigStack: [],
-            routeStack: [],
+            presentedIndex: 0,
```
erictraut: The initialization of presentedIndex is a good fix. Initialization of transitionFromIndex isn't necessary because it's set to undefined if not initialized, and this is a fine value (effectively the same as null).

RyanThomas73:
@erictraut
Hi Eric. Thanks for the feedback so far. I've the space on the if statement per your recommendation.

In regards to your comment about initializing the transitionFromIndex to null. Without this the navigator fails to transition for me because of the code at line 428. I'm open to an alternative fix for this if you would prefer such as changing the if statement to `!= null` etc.

```
// undefined !== null so this block is entered when it should not be
if (this.state.transitionFromIndex !== null) {                         
```


In regards to the other two comments you posted regarding my changes to `renderScene(...)` and the support for `initialRoute` and `initialRouteStack`. At the time I began expermienting with using the `reactxp <Navigator>` component there was not documentation available for the component. Because of this I used the `react-native <Navigator>` component and documentation as a starting point. These components support all three of the changes that I've made `initialRoute`, `initialRouteStack`, and the 2nd parameter for `renderScene(...)`. Looking at the `reactxp <Navigator>` documentation that has since been added it appears `reactxp` has deviated from supporting these in a few ways. I'm not sure of the degree to which `reactxp` intends to maintain support for the syntax of the `react-native` components. 

In lieu of supporting the `initialRoute` property you're example code uses the following logic.
It appears this example would work for the `initialRouteStack` property as well.
```
componentDidMount() {
        // Now that the app is mounted, specify the initial
        // navigator route.
        this._navigator.immediatelyResetRouteStack([{
            routeId: NavigationRouteId.MainPanel,
            sceneConfigType: RX.Types.NavigatorSceneConfigType.Fade
        }]);
}
```

In lieu of the `renderScene(...)` being passed the navigator instance as a second parameter, the `reactxp` example code does the following:
```
render() {
        return (
            <RX.Navigator
                ref={ this._onNavigatorRef }
                renderScene={ this._renderScene }
            />
        );
    }

    private _onNavigatorRef = (naviator: RX.Navigator) => {
        // Stash away a reference to the mounted navigator
        this._navigator = navigator;
    }
```

Now that there is documentation I can see how `reactxp` has handle these scenarios differently.
Let me know if you would like me to remove the support for these outright or if you would like me to investigate updating the non web components to include these since the underlying `react-native` components did support them prior to the `<Navigator>` component being removed.

src/web/Navigator.tsx
RyanThomas73: Without this the navigator fails to transition for me because of the code at line 428. I'm open to an alternative fix for this if you would prefer such as changing the if statement to `!= null` etc.

```
// undefined !== null so this block is entered when it should not be
if (this.state.transitionFromIndex !== null) {                         
```

src/common/Types.ts
RyanThomas73: At the time I began experimenting with `reactxp <Navigator />` there was no documentation available so I based my code on the `react-native <Navigator/>` component which included this second parameter. Looking at the `reactxp` documentation that has since been added I see how you've handled this differently.


In lieu of supporting the `initialRoute` property you're example code uses the following logic.
It appears this example would work for the `initialRouteStack` property as well.
```
componentDidMount() {
        // Now that the app is mounted, specify the initial
        // navigator route.
        this._navigator.immediatelyResetRouteStack([{
            routeId: NavigationRouteId.MainPanel,
            sceneConfigType: RX.Types.NavigatorSceneConfigType.Fade
        }]);
}
```

src/common/Types.ts
RyanThomas73: In lieu of supporting the `initialRoute` property you're example code uses the following logic.
It appears this example would work for the `initialRouteStack` property as well.
```
componentDidMount() {
        // Now that the app is mounted, specify the initial
        // navigator route.
        this._navigator.immediatelyResetRouteStack([{
            routeId: NavigationRouteId.MainPanel,
            sceneConfigType: RX.Types.NavigatorSceneConfigType.Fade
        }]);
}
```

erictraut:
@RyanThomas73: OK, thanks for the explanation. I think we should leave the existing interface in place rather than change it. I do want to incorporate your other bug fixes though (the state initialization issue). I'd like to fix a few additional issues (mostly just inconsistencies) while I'm in there, so I'm going to create a separate PR. I'll include you on it for review. Thanks again for raising this issue.

PR #102, opened 2017-05-19
adds alignContent prop to FlexboxStyle

ofirgeller:
I tested that this works on the web ReactXP. the docs from react native indicate it is also supported.
https://facebook.github.io/react-native/docs/layout-props.html#aligncontent

It also seems that we are missing 'space-between', 'space-around' on all props but I don't know if that a ReactEx specific limitation.


erictraut:
Thanks @ofirgeller. I'll also update the documentation to reflect this addition.

PR #103, opened 2017-05-19
FocusManager to provide a better keyboard navigation.

mshoho:


docs/docs/components/view.md
```diff
@@ -40,8 +40,16 @@ blockPointerEvents: boolean = false; // iOS and Android only
 // components behind to receive them
 ignorePointerEvents: boolean = false; // web only
 
-// Can the component accept keyboard focus?
-focusable: boolean = false; // web only
```
mshoho: This on seems to not be used anywhere.

src/web/View.tsx
```diff
@@ -66,21 +67,33 @@ if (typeof document !== 'undefined') {
...
     };
 
+    private focusManager: FocusManager;
+
     private resizeDetectorAnimationFrame: number;
     private resizeDetectorNodes: { grow?: HTMLElement, shrink?: HTMLElement } = {};
 
+    constructor(props: Types.ViewProps) {
+        super(props);
+        if (this.props.restrictFocusWithin) {
```
erictraut: This should be "props", not "this.props".

erictraut: What happens if the "restrictFocusWithin" prop is updated? Looks like this case isn't handled. Probably requires adding a componentDidUpdate or componentWillUpdate method.

src/web/View.tsx
```diff
@@ -66,21 +67,33 @@ if (typeof document !== 'undefined') {
...
     };
     context: ViewContext;
 
     static childContextTypes: React.ValidationMap<any> = {
-        isRxParentAText: PropTypes.bool.isRequired
+        isRxParentAText: PropTypes.bool.isRequired,
+        focusManager: PropTypes.object
     };
 
+    private focusManager: FocusManager;
```
erictraut: Name should include underscore for consistency with coding conventions used throughout ReactXP.

src/web/View.tsx
```diff
@@ -66,21 +67,33 @@ if (typeof document !== 'undefined') {
...
+
     private resizeDetectorAnimationFrame: number;
```
erictraut: These should start with underscores as well. (I didn't catch that in the previous review.)

src/web/utils/FocusManager.ts
```diff
@@ -0,0 +1,198 @@
...
+let _lastComponentId: number = 0;
+
+interface StoredFocusableComponent {
+    component: React.Component<any, any>;
+    restricted?: boolean;
+    origTabIndex?: number;
+}
+
+export class FocusManager {
+    private static rootFocusManager: FocusManager;
```
erictraut: Private variables should start with underscore.

src/web/utils/FocusManager.ts
```diff
@@ -0,0 +1,198 @@
...
+    private static restrictionStack: FocusManager[] = [];
+    private static currentRestrictionOwner: FocusManager;
+
+    private static allFocusableComponents: {[id: string]: StoredFocusableComponent} = {};
```
erictraut: Opening curly brace should have space after it; similar for closing brace. Same comment below.

mshoho:
Dynamically changing restrictFocusWithin has required a relatively complex change (I now maintain a tree of FocusManagers because React doesn't have a public API to get the actual child component instances, so I can't traverse them and I'm building a FocusManager instance for each View).

src/web/View.tsx
```diff
@@ -82,19 +82,17 @@ export class View extends ViewBase<Types.ViewProps, {}> {
...
-    private resizeDetectorNodes: { grow?: HTMLElement, shrink?: HTMLElement } = {};
+    private _resizeDetectorAnimationFrame: number;
+    private _resizeDetectorNodes: { grow?: HTMLElement, shrink?: HTMLElement } = {};
 
     constructor(props: Types.ViewProps) {
         super(props);
-        if (this.props.restrictFocusWithin) {
-            this.focusManager = new FocusManager();
-        }
+        this._focusManager = new FocusManager();
```
erictraut: I'm a bit concerned about the perf and memory implications of this. We allocate a LOT of Views, and most of them don't need a FocusManager. Could we allocate this on demand or conditionally, only when restrictFocusWithin is true? If not, could we allocate it only when mounted?

mshoho: The problem is that getChildContext() which provides the children with appropriate FocusManager instance is invoked before the component is mounted (so, there would be no possibility to maintain the proper relations between instances). And React doesn't have a public API to traverse the child component instances (props.children provides properties and some other things, but not the exact instances and even no state).

What we can do is to force the `restrictFocusWithin` property to be readonly (and tell about it in the documentation) — it's sort of ugly, but makes sense, I think. I could go back to the initial implementation, but log an error in case restrictFocusWithin property is being changed during the component life cycle.

erictraut: I think that's a reasonable restriction. Let's go back to your previous implementation and document it as a read-only prop.

mshoho: Done.

mshoho:
In the name of performance I've implemented one more thing: isConditionallyFocusable property for FocusableComponentMixin. Views are focusable only when their tabIndex >= 0, accounting for them conditionally dramatically reduces the number of component references (and the number of tabIndex changes) in FocusManager.

PR #104, opened 2017-05-22
Virtuallistview extension missing index files added

4ndr1u5:
reactxp-virtuallistview npm package is missing index files - I've added them here so that example at https://microsoft.github.io/reactxp/docs/extensions/virtuallistview.html works (given you update import line to
`import { VirtualListView, VirtualListViewItemInfo } from 'reactxp-virtuallistview';`)


erictraut:
Thanks @4ndr1u5!

PR #105, opened 2017-05-24
Fixed state initialization in web implementation of Navigator. 

erictraut:
Also switched from null to undefined for consistency.


erictraut:
@RyanThomas73 Here's my updated PR. For some reason, github isn't allowing me to add you as a reviewer. If you're able to see this branch, please review and let me know if you see any problems with my changes.

PR #107, opened 2017-05-24
Icon support (web-only) for Alert component + Traditional alert() fallback

psegalen:
Hi all,
I'd like to set the icon for my web notifications and I also would like a traditional JS alert() triggered as a fallback if the user's browser doesn't support web notifications.
What do you think?


psegalen:
I wonder why it shows 5 commits, maybe because I didn't sync my fork since my previous PR?

PR #115, opened 2017-05-31
Diagnostic code to raise an event when the JS layer handling of a native touch event is delayed by more than a specified threshold 100ms or more.

epatey:
@msfterictraut, this is the change that we discussed last week.

### DIFF
--- a/src/native-common/UserInterface.ts
+++ b/src/native-common/UserInterface.ts
@@ -18,6 +18,8 @@ import RX = require('../common/Interfaces');
 import Types = require('../common/Types');
 
 export class UserInterface extends RX.UserInterface {
+    private _touchLatencyThresholhdMs: number;
+
     constructor() {
         super();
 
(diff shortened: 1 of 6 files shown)


src/native-common/UserInterface.ts
```diff
@@ -143,6 +145,19 @@ export class UserInterface extends RX.UserInterface {
...
         // Nothing to do
     }
+
+    enableTouchLatencyEvents(latencyThresholdMs: number): void {
+        this._touchLatencyThresholhdMs = latencyThresholdMs;
+    }
+
+    evaluateTouchLatency(e: Types.SyntheticEvent) {
+        if (this._touchLatencyThresholhdMs) {
+            const latency = Date.now() - e.timeStamp.valueOf();
```
berickson1: GestureView.tsx has the following code, do we need similar here?
```typescript
        var timestamp = e.timeStamp;
        // Work around a bug in some versions of RN where "timestamp" is
        // capitalized differently for some events.
        if (!timestamp) {
            timestamp = e.timestamp;
        }
        if (!timestamp) {
            return 0;
        }
```

epatey: Oh good catch. Blech, but I'll add it.

PR #116, opened 2017-05-31
display flex removed from root view default styles

sorenrehkopf:
Display was being set to flex on the root element for some reason. Causing unnecessary layout issues. Doesn't seem like there would be a necessary functionality reason to have that style set especially since it's specific to web.

### DIFF
--- a/src/web/RootView.tsx
+++ b/src/web/RootView.tsx
@@ -233,7 +233,6 @@ export class RootView extends React.Component<RootViewProps, RootViewState> {
         let rootViewStyle = {
             width: '100%',
             height: '100%',
-            display: 'flex',
             cursor: 'default'
         };
 


erictraut:
@sorenrehkopf The root view is intentionally set to use flex layout. All views in ReactXP are assumed to do so, since that provides compatibility with React Native. Can you provide more details about what problem you were seeing?

sorenrehkopf:
Hey Eric, I figured it was intentional, but the problem with that is when
building for web it overrides some of the default functionality for DOM
elements since anything inside of the root element that's being bound is
inside of that flex container. IE, any block element that's a descendent of
that root element no longer behaves like a block element in terms of
sizing.I just had a nav that I wanted positioned at the top of my page. As
a block element, default functionality should make it take up 100% of the
window width, but due to the display flex, it sizes more like an inline
element, only taking up as much space as is needed by the content on the
page.

Obviously there are hacks to get around this, ie, overriding the inline
style for that root element with !important, or setting block elements to
100vw (has some problems of it's own), but it seems like overriding default
functionality for all block descendants of the root element seems like a
bad thing to have hard-coded.

Maybe a better solution would be to allow for a config option to turn that
functionality off for web in the app.setMainView function? I think my pr
only turns it off for web though. Should still have the display flex for
native builds I think.

Won't be offended if you reject my pr but you guys should think about
handling it differently.

-Soren

RangerMauve:
@sorenrehkopf You could add `alignSelf: "stretch"` to your nav to make it take up the full horizontal width

erictraut:
If you want to host web-specific content that depends on block layout in any subview within ReactXP, you're going to run into this same problem. It's not specific to the root view. I think the correct solution is for you to create a web-specific container div that establishes a "display: block" context within the ReactXP parent view (in this case, the root view). 

Your proposed solution will cause problems for other ReactXP apps because they expect subviews to follow flex layout.

sorenrehkopf:
That doesn't solve the problem because setting display block on an element
inside the parent view still means that element is a flex item which
doesn't default to full window for width. The whole problem is that setting
the ReactXP parent view to flex causes display:block to not behave like
expected for any descendants, without also setting the child to 100% width
or using some other ad-hoc fix. Not a huge deal, but definitely unexpected,
and not desirable in a web context.

Anyway, a related issue may be that I'm using react-router instead of the
built in navigator api cause the navigator api seems to not support urls?
Was unable to find any docs on it and my app requires distinct url's for
different routes on web.

-Soren

erictraut:
As RangerMauve mentioned, you can add `alignSelf: "stretch"` to the nav container to get it to expand to the width of its parent. That's the flex equivalent of "width: 100%" when the flex direction is column (the default in ReactXP).

I don't think it's true that display:flex "causes display:block not behave like expected for any descendants". The "display" CSS property isn't cascading, so it shouldn't affect descendants. This is consistent with what I'm seeing in my testing.

sorenrehkopf:
I know display doesn't cascade, but it does affect the descendants because
display block and manually setting width 100% are both relevant to the
parent. If you don't manually set all direct children of the ReactXP parent
view to 100% width than they, and any children, will not size to the width
of the window when set to display:block, nor when set to width:100% for
non-direct descendants.

Like I said, I understand what's going on and how to work around it, I'm
just saying it makes for a non-intuitive layout experience for web. More of
a design recommendation than a bug-fix.

Your project though so as you will.

RangerMauve:
> setting width 100%

Have you read the comment I posted about using `alignSelf: "stretch"`? You don't need to set the width.
If the root component is bothering you, add a view that has `display: block` and `flex: 1` which should make it fill the entire space and allow it's children to not worry about flexbox at all.

RangerMauve:
As well, if you're planning on using ReactXP, you should be making use of flexbox and View components for layout regardless of whether you're building for web or native. Otherwise you lose most of the benefits of using it.

sorenrehkopf:
I did read the comment about alignSelf:stretch and I know that fixes it
this specific problem. My suggestion is just to make it so that people who
may not want to apply that to every single different view they have,
doesn't have to just find the work-around that works best for them.

Just opened one more pr that keeps the default functionality but allows for
a config option in the setMainView function to specify different styles.
This was the first thing I looked for when I realized what was breaking my
layout. Also put a config option to let the user specify the element they
want to bind to cause that just seems like a pretty basic thing that wasn't
there already.

Think this will let it behave the way you guys want as a default, but also
let others who have the same problem I was having work around it without
having to put ad-hoc css hacks into their styles just for this issue.

-Soren

erictraut:
What you're trying to do is not really in keeping with the philosophy of ReactXP. I understand your particular need, but it's not going to be a common scenario. Since we've already identified a viable workaround that doesn't involve a change to the public interface of ReactXP, I don't think we should make this change.

sorenrehkopf:
Like I said your project so your call.

As I understand it, one of the main points of ReactXP is to allow for one
react code-base for web in addition to mobile. As such, I strongly believe
that having a default layout style like that set on the root view element,
and not giving the user an option to override it using the api itself makes
it less usable for web. You can call it the philosophy of ReactXP if you
want, but while opinionated yet flexible is usable, opinionated and rigid
is less so. The pr I just opened is non-obtrusive and I believe makes it
more adaptable for web.

Do with it what you will though.

Thanks for taking the time to discuss it at least.

-Soren

PR #122, opened 2017-06-05
Exposing clear function in textinput

RamyaSenk:
In some android devices, the auto suggestion of keyboards are not getting cleared when the text in text input is cleared. Currently we can clear text in text input by setting inputValue to ''. Trying to see if calling clear() clears the auto suggestion.


erictraut:
@RamyaSenk Is this just an experiment, or are you sure this fixes the problem? If it's just an experiment, can you do a local build to confirm before we make this change to ReactXP?

RamyaSenk:
@erictraut this is an experiment. Let me abandon the pull request, till i am sure if fixes the problem.

PR #125, opened 2017-06-08
Avoid "error TS2322: Type '{}' is not assignable to type 'null'." when "tsc" on hello-world sample

psegalen:
Hello,
After cloning the hello-world sampe, `npm install` and `tsc`, I have an error saying:
> error TS2322: Type '{}' is not assignable to type 'null'.

I think it's due to a recent version of TypeScript, f7627da is a workaround.


erictraut:
Thanks for the fix.

psegalen:
You're welcome! Thanks for the lib! :smile:

PR #147, opened 2017-06-23
setFocusRestricted() and setFocusLimited() methods and nested focus restrictions fix

mshoho:
Three things:
1. setFocusRestricted() and setFocusLimited() methods for View to change the focus restrictions during the component life cycle.
2. Fix for handling nested focus restrictions.
3. subscribe(), unsubscribe() and isComponentFocusRestrictedOrLimited() methods for FocusManager to be aware of the restriction state.

### DIFF
--- a/src/web/UserInterface.ts
+++ b/src/web/UserInterface.ts
@@ -18,8 +18,6 @@ import RX = require('../common/Interfaces');
 import Types = require('../common/Types');
 
 export class UserInterface extends RX.UserInterface {
-    private _layoutChangeAnimationFrame: number;
-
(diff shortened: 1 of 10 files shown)


src/web/UserInterface.ts
```diff
@@ -18,8 +18,6 @@ import RX = require('../common/Interfaces');
 import Types = require('../common/Types');
 
 export class UserInterface extends RX.UserInterface {
-    private _layoutChangeAnimationFrame: number;
```
mshoho: An unused leftover.

PR #149, opened 2017-06-27
Upgrade typescript compiler, move to @types for lodash, use public SubscribableEvent

berickson1:
New lib.d.ts has fetch types and es6-promise types, so remove from project
Fix some type incompatibilities between RN & Web Animated implementation


extensions/imagesvg/src/typings/react-native.d.ts
```diff
@@ -857,7 +857,7 @@ declare module 'react-native' {
          class Value {
             constructor(val: number);
             setValue(value: number): void;
-            addListener(callback: any): number;
```
berickson1: Looked at RN code, this is actually a string return, not a number

Issue #181, opened 2017-07-21
RN initialProps don't get passed from RootView to mainView


Fitzpasd:
React native has the ability to pass initial properties from native code to components, [see here](https://facebook.github.io/react-native/docs/communication-ios.html#passing-properties-from-native-to-react-native)

ReactXP's `RootView` receives these, but doesn't pass them down to the `mainView`.

I'll submit a PR for this shortly.  Opening this bug to track/link to.

PR #182, opened 2017-07-21
Pass initialProps from RootView to inner mainView

Fitzpasd:
Fix for issue #181.

This is the simplest fix imaginable, simply pass all the `RootViews` props to the `mainView` if they exist.  This seems fine right now, since `RootView` doesn't have any props, and I don't think that's likely to change in the future.  However, if it does, then we should guard against passing down props that are RootView-specific.  There isn't a nice way to pre-guard against that now IMO, since we'd have to work around naming conflicts.  I'm of the opinion that this simple fix is appropriate for now, and we shouldn't future-proof any further.  Open to opinions though.

(I'm pretty new to TS and React, so comments or alternative solutions are very welcome).

### DIFF
--- a/src/native-common/RootView.tsx
+++ b/src/native-common/RootView.tsx
@@ -113,8 +113,14 @@ export class RootView extends React.Component<{}, RootViewState> {
     }
 
     private _getStateFromStore(): RootViewState {
+        let mainView = MainViewStore.getMainView();
+
+        if (this.props) {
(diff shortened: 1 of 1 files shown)


erictraut:
Thanks @Fitzpasd, looks good. I wasn't even aware that RN passed initialProps to its top-level component. We'll publish this as part of our next round of changes.

PR #193, opened 2017-08-01
Add support for custom accessibility actions on iOS.

ewdavison:
@deregtd

### DIFF
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -341,7 +341,7 @@ export interface CommonAccessibilityProps {
 
     // iOS only.
     accessibilityActions?: string[];
-    onAccessibilityAction?: Function;
+    onAccessibilityAction?: (e: SyntheticEvent) => void;
 }
 
 // Auto, Yes, No - iOS & Android.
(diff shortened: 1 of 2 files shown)


src/common/Types.ts
```diff
@@ -338,6 +338,10 @@ export interface CommonAccessibilityProps {
 
     // Desktop only.
     tabIndex?: number;
+
+    // iOS only.
+    accessibilityActions?: string[];
+    onAccessibilityAction?: Function;
```
erictraut: This function should have a well-defined type (including incoming and return params) rather than a generic Function type.

src/common/Types.ts
```diff
@@ -338,6 +338,10 @@ export interface CommonAccessibilityProps {
 
     // Desktop only.
     tabIndex?: number;
+
+    // iOS only.
+    accessibilityActions?: string[];
```
erictraut: These new parameters need to be added to the documentation in accessibility.md.

ewdavison: Added information to accessibility.md.

PR #196, opened 2017-08-04
Making gesture views accessibile

RamyaSenk:
Currently we cannot recognise swipe and pan gestures when screen reader is enabled. Making gesture views implement accessibility properties. Also making gesture views accessible by default because it is associated with actions done on the view - which should almost always be accessible. Similar to button behaviour. 

You can test these gestures by long pressing on the gesture view and performing the gesture like swipe and pan.


src/native-common/GestureView.tsx
```diff
@@ -509,12 +513,21 @@ export abstract class GestureView extends ViewBase<Types.GestureViewProps, {}> {
...
     render() {
+        const importantForAccessibility = AccessibilityUtil.importantForAccessibilityToString(this.props.importantForAccessibility,
+            _defaultImportantForAccessibility);
+        const accessibilityTrait = AccessibilityUtil.accessibilityTraitToString(this.props.accessibilityTraits);
+        const accessibilityComponentType = AccessibilityUtil.accessibilityComponentTypeToString(this.props.accessibilityTraits);
+
         return (
             <RN.View
-                style={ this._getStyles(this.props) }
+                style={this._getStyles(this.props)}
```
erictraut: Please add spaces within the curly brackets to be consistent with all other code in ReactXP.

RamyaSenk:
Thanks @erictraut 

I have updated the document and fixed spaces. I missed web, have implemented accessibility props in there as well.

src/native-common/GestureView.tsx
RamyaSenk: Done. Updated it.

Issue #169, opened 2017-07-13
RX.Alert doesn't support buttons on web


berickson1:
If I create an alert that has buttons using RX.Alert, it displays a browser toast without buttons. ON web I can never invoke the callback of buttons.

erictraut:
This is a limitation of the web Notification standard. I don't see any way around this. Do you have suggestions? The web standard does provide a single onClick event that is invoked when the user taps on the notification. We could somehow pass that back through ReactXP, but I don't think that's what you're looking for.

berickson1:
I'm not really sure what the proper behaviour should be, but the component doesn't really conform to the ReactXP idea of cross-plat compat.

Some browsers (firefox/safari) also automatically close these alerts after a few seconds, meaning the user may never see them

Potential Options:
1) Update the docs to indicate that RX.Alert isn't properly supported on web
2) Change RX.Alert to be a custom styled modal on web with buttons (more like native)

erictraut:
@apotra I think you wrote that module originally, right? Do you have a strong opinion one way or another?

psegalen:
Hi you two,
I personally think that web notifications are pretty useful as, like the push notifications on a mobile OS, it can reach users even if they are outside the app (for example if the browser has been minified).
But I agree 100% with @berickson1 about the cross-platform side.
Plus, web notifications aren't alerts (like js `alert()`) so I would go for @berickson1's option 2 for `RX.Alert` and I would create a `RX.Notification` sending a web notification on web and a local notification on Android (I think local notifs aren't supported by iOS), what do you think?

MovingGifts:
I think option 2 would be best for web.

MovingGifts:
I think it's best for the RX.Alert to be written once and work cross platform from the ReactXP side. For web, the RX.Alert with functional buttons (which work on mobile) should be transformed to a modal with functional buttons.

Any chance of this getting implemented as part of core ReactXP?

erictraut:
The challenge with this approach is that we'd need to pick some style for the buttons, and it's possible (likely?) that this style won't match the style in the rest of your app. In all other cases, ReactXP is completely style-agnostic. I guess we could provide style overrides for the various components of the alert dialog so callers can "skin" it to match the rest of their app.

MovingGifts:
I totally agree with the skinning approach here; as long as the developer can override the styles and get alert to modal functionality on web, that would make it cross platform :)

MovingGifts:
[...]

PR #219, opened 2017-08-22
Convert the web Alert from notification to Modal

psegalen:
First attempt to convert current web specific Alert component from a web notification to a Modal.
Related to issue #169


erictraut:
@psegalen - thanks for the submission. Sorry for the slow response. I've been on vacation for the past two weeks. I just pushed an updated PR to your branch. Please review and let me know if it looks good. If so, I can merge it into master.

psegalen:
@erictraut - no problem about the delay, we all need vacation from time to time!
Your update seems good to me, it shows me how much my code needed to be cleaned... If I submit another PR someday, I'll try to remember this.

MovingGifts:
Hmmm, I don't see the RX.Modal working, it still prompts the notification way (no change), I think something with the version? I have this atm:
```json
"reactxp": "^0.42.1",
```
installed with:

`npm i -S reactxp`

Wondering how I can install the latest reactxp package that contains it?

erictraut:
I assume you mean RX.Alert. The updated behavior is available only in the master branch, not the 0.42.x branch. You can use version "0.46.0_rc.1".

MovingGifts:
Yup, sorry I meant the RX.Alert's modal :)

I tried this to install the version "0.46.0_rc.1":
```
npm i -S reactxp@0.46.0_rc.1
```
Then I got an error like so:
```
npm ERR! code ETARGET
npm ERR! notarget No matching version found for reactxp@0.46.0_rc.1
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
```
Not sure why I got that error, am I installing it incorrectly?

MovingGifts:
I tried it with a dash instead like so: `0.46.0-rc.1`

Which installed but I still don't see the `src/AlertModalContent.tsx`

MovingGifts:
The following seemed to work for the time being, but once `0.46` is out, I will change it to that.
```
"reactxp": "git+https://github.com/Microsoft/reactxp.git"
```

MovingGifts:
Sorry to keep posting, but I get an error in web (only checked in web) with:

`"reactxp": "git+https://github.com/Microsoft/reactxp.git"`

screen shot 2017-09-08 at 7 04 01 pm
The error is because there isn't a `dist` folder at all with this version and `node_modules/reactxp/index.js` has this line:
```js
// Export web by default. Other platforms have custom index.[platform].js files
module.exports = require('./dist/web/ReactXP.js');
```
The `0.46.0-rc.1` version on npm install creates the `dist` directory correctly at `node_modules/reactxp/dist` so there is no error and it loads the web app, but the web alert is not available in that version.

Not sure what to do :(

erictraut:
If you are using a published version of reactxp, the dist directly will be populated already. If you're directly checking out the reactxp code from git, you'll need to build reactxp yourself. Use "npm install" and "npm run build" to populate the dist directory.

MovingGifts:
Yup, that did it :) Thank you so much!

PR #226, opened 2017-08-25
Fix typo in navigator documentation

rajington:
seems like an over-deletion occurred in [this PR](https://github.com/Microsoft/reactxp/commit/e73b2b6a31572ee84b3755724cab3310e72e3b0b##commitcomment-23871642)

### DIFF
--- a/extensions/navigation/README.md
+++ b/extensions/navigation/README.md
@@ -1,4 +1,5 @@
-d: components/navigator
+---
+id: components/navigator
 title: Navigator
 layout: docs
 category: Components


dryganets:
Thank you.
I refactored the documentation little bit differently.
My bad.
https://github.com/Microsoft/reactxp/pull/227

rajington:
np at all @dryganets, but i think you might've missed moving it to extensions here: https://github.com/Microsoft/reactxp/blob/master/docs/_data/nav_docs.yml#L35:L36

PR #229, opened 2017-08-26
reactxp-video uses newer version of react-native-video

dryganets:
This fixes the samples using reactxp-video component.


dryganets:
Closed to avoid accidental merge.
Need to wait till we get the reactxp-video published to npm first.

Issue #235, opened 2017-08-28
when navigator back,Component not call  componentWillUnmount (reactxp-navigation)


wangfaxi1985:
when navigator back,Component not call  componentWillUnmount -- (web)
i want a callback when navigator back?  how can i do 

reactxp-navigation:1.0.6

dryganets:
I see the root cause of the bug - we have direct state manipulations in navigator for the web.
Trying to figure out why it is done this way as I'm not an author of this code.

Thank you for reporting this.

wangfaxi1985:
i just read reactxp-navigation source ,()
because when pop back complete ,method :_cleanScenesPastIndex ,if condition　dissatisfaction ( i navigate to root Component)

PR #239, opened 2017-08-29
reactxp-navigator 1.0.7. Web Navigator's pop command unmount previous scene

dryganets:


dryganets:
Fixes: https://github.com/Microsoft/reactxp/issues/235

PR #252, opened 2017-09-05
Making url for the links mandatory to prevent misusing links as inaccessible buttons.

mshoho:


src/web/Link.tsx
```diff
@@ -62,9 +66,22 @@ export class Link extends React.Component<Types.LinkProps, {}> {
                 onMouseEnter={ this.props.onHoverStart }
                 onMouseLeave={ this.props.onHoverEnd }
                 onMouseDown={ this._onMouseDown }
+                tabIndex={ this.props.tabIndex }
             >
                 { this.props.children }
             </a>
+        ) : (
+            <button // When url is not set <a> will be unaccessible, so we make it a button.
```
erictraut: This imposes a new limitation on the use of an RX.Link. It can no longer be used inside of an RX.Button because of the nesting rules of <buttons> in HTML. That's probably OK, but it's something we should be aware of.

src/web/Link.tsx
```diff
@@ -47,11 +51,11 @@ export class Link extends React.Component<Types.LinkProps, {}> {
...
-       // SECURITY WARNING:
-       //   Note the use of rel='noreferrer'
-       //   Destroy the back-link to this window. Otherwise the (untrusted) URL we are about to load can redirect OUR window.
-       //   See: https://mathiasbynens.github.io/rel-noopener/
-        return (
+        // SECURITY WARNING:
+        //   Note the use of rel='noreferrer'
+        //   Destroy the back-link to this window. Otherwise the (untrusted) URL we are about to load can redirect OUR window.
+        //   See: https://mathiasbynens.github.io/rel-noopener/
+        return this.props.url ? (
```
erictraut: Is there really a need for RX.Link to support the case where there's no provided URL? Why not just make 'url' a required prop?

mshoho: Yes, valid point. I should have addressed it from the other direction.

src/web/Link.tsx
mshoho: As for this part, the spec says that both `<a>` and `<button>` can't have interactive content:

https://w3c.github.io/html/sec-forms.html#the-button-element
https://w3c.github.io/html/textlevel-semantics.html#the-a-element

It's not that we check for a validity given the dynamic DOM generation, though...

mshoho:
Pushed the change and renamed the PR to reflect the change.

Issue #203, opened 2017-08-14
Expose an AppRegistry.registerComponent equivalent to allow multiple different root views in RN apps


Fitzpasd:
ReactXP is currently limited to one root component, `RXApp`, and doesn't allow for developers to register and use others.  This is a major limitation when integrating ReactXP into an existing iOS/Android application.  

Consider the scenario where a native app wants to have 2 react-based screens and the rest of the application is written in native code.  This isn't easily done with the current implementation of ReactXP due to the reliance on `UserInterface.setMainView` and the internal use of `RootView`.

Proposed solution:

- Expose an equivalent interface to `AppRegistry.registerComponent `
- Internally wrap that registered component in a higher-order component that:
    - uses `componentWillMount` to call `UserInterface.setMainView` with the original component
    - renders `RootView`.

I have a local (hacked) implementation of the above that makes me believe it is a feasible approach.  I'm happy to clean that up and send out a PR if this is something the team wants to implement.

dryganets:
Hi @Fitzpasd,
The feature you want to add, exist in react-native only and we are trying to keep the core of the reactxp as slim as possible.
You could make a reactxp extension which exposes this API.
Feel free to use any of https://github.com/Microsoft/reactxp/tree/master/extensions/ as a template.
Remeber, you need to have a stubs/implementation for not react-native platforms to make it cross-platform.

erictraut:
Hi @Fitzpasd, thanks for the submission. This is good work.

I understand @dryganets perspective, but I'm going to overrule him in this case. I think this is a worthwhile addition to ReactXP. However, I think we should reconsider the interface. Rather than exposing a new AppRegistry namespace, which is an RN-specific concept, I'd like to simply add a new method (perhaps "registerRootView"?) to the existing RX.UserInterface namespace that allows callers to register alternative root views.

I'd like to avoid complex or non-deterministic interactions between setMainView and registerRootView. Let's think about how to simplify that.

Fitzpasd:
I like the idea of merging this interface into `RX.UserInterface`, I think that fits nicer.

The coupling with `setMainView` is concerning.  How about introducing a property on `RootView` that can be used to set the initial mainView in the ctor?

Fitzpasd:
I opened another PR that implements the above approach.  Useful for discussion on how best to implement this.

neelimabhukya:
how to implement datepicker or calander in reacrtxp?

erictraut:
[...]

PR #257, opened 2017-09-07
UserInterface.registerRootView method, mimicing RN's AppRegistry.regi…

Fitzpasd:
…sterComponent functionality.

See issue #203 for discussion.


src/native-common/RootView.tsx
```diff
@@ -115,14 +126,22 @@ export class RootView extends React.Component<{}, RootViewState> {
...
 
-        if (mainView && this.props) {
-            mainView = React.cloneElement(mainView, this.props);
+        if (mainView && !_.isEqual(mainView.props, this._mainViewProps)) {
+            mainView = React.cloneElement(mainView, this._mainViewProps);
         }
 
         return {
-            mainView: mainView
+            mainView: mainView || this.state.mainView
```
Fitzpasd: Do we want to allow `mainView` to be set to null via the store?  The reason I'm not allowing it here is to prevent the `componentWillMount` setState call overriding the mainView setup from the ctor.  If we do want to allow null from store, this can be refactored to prevent the overriding in a different manner.

erictraut:
Your code is rendering a ReactXP RootView instance for not only the main view but all alternate roots. The ReactXP RootView is responsible for maintaining the popup and modal layer in the app, and there should be only one of those. I was assuming that alternate roots would not display modals or popups. The way it's coded currently, a modal or popup is going to appear on the main view and all registered alternate views. Was that your intention? I don't think the exist modal and popup logic will work if there are multiple RootViews instantiated.

Fitzpasd:
That was my intention, I think that the full API should be available from any registered component.  To make ReactXP feasible for integrating into existing native apps ([RN supports this](https://facebook.github.io/react-native/docs/integration-with-existing-apps.html), and is a [common ](https://www.youtube.com/watch?v=tWitQoPgs8w) [scenario](https://engineering.instagram.com/react-native-at-instagram-dd828a9a90c7)), it has to be possible to register and use independent components.  Typically there would be two scenarios:

1. Acting as a full page Screen.  This is what this PR implements, and I hosted them inside a new instance of `RootView` to allow use of the Modal, Popup and Accessibility APIs.
2. Acting as a single view object on a screen.  I just thought of this use-case now, and there should probably also be a new API for this.

It would be much easier to only support scenario 2, but then we'd have to somehow make it very clear that there are some APIs that won't work in that instance.  That's something our team could live with though, if you think that's best for ReactXP itself.

Fitzpasd:
@erictraut - After some more testing of this, you're right that the Modal layer goes into unexpected states with this current implementation.  Popups appear to work.

How would you feel about changing this PR to not wrap the registered components in a `RootView` and opening a new bug to address the Modal API then not working?  I haven't dug into the source of that API yet (and won't have time for a week or two), but to fully support this scenario we'd need to make changes to have each `RootView` have it's own instances of the various Modal components/managers, or vice-versa.

Fitzpasd:
To summarize the options as I see them:

1. Fully support this scenario.
2. Partially support this scenario, with clear messaging that some RXP APIs won't work.
3. Don't support this scenario.

I'm a believer in option `1`, although I can see how that might not be a small amount of work given the singleton nature of some of the components that the Modal API uses.

I don't like option `2`, the messaging around broken APIs will be tough.  

I don't like option `3` because I think there's value in supporting this.  If that's the route we do go, our team could use the RN API directly for a workaround and never use the broken APIs (fine for the short-term, but would make me uneasy the more we adopt RXP).

erictraut:
@Fitzpasd, sorry for the slow reply. I've been focused on other things. 

Here's how I'm thinking about this. Multiple root views can be used in a variety of ways. Each root view is meant to represent a somewhat-independent React program, even though they share the same JS environment in the current implementation of RN. In some use cases, they can be placed side-by-side on the screen, each embedded in a separate native control region. In other cases, they can cover the entire screen.

Based on this, I don't think it makes sense to present popups or modals on all root views. They should either appear in the main view only or we should allow users to specify which root view they are targeting (with the default being the main view).

As to your point about making the code resilient to displaying multiple instances of modals and popups... I'm worried not only about the ReactXP framework code but also the callback logic that app writers need to implement such as renderPopup, onDismiss, onAnchorPressed, etc. Callers are not expecting these to be called multiple times for a single popup.

My proposal:
1. Add a new optional parameter to the ```Modal.show``` method that has a type of ```ModalOptions```. This object would have one optional field (with room for additional fields in the future) called ```rootViewId```. 
2. Likewise, add an optional ```rootViewId``` field to the existing ```PopupOptions```.
3. Based on the value of the rootViewId, present the popup or modal within the specified root view. If rootViewId is not specified, assume the main view.

Thoughts?

Fitzpasd:
Sorry for late response, also been busy with other stuff.

I like the proposal, and can start working on it now.  Should we include the `Alert` API though?  Those (at least on mobile) are typically overlaid on the entire screen, and not attached to any view.

Fitzpasd:
I like this proposal, and can start working on it today.

I don't think we want to include the `Alert` API here.  Those are typically set on the overall window, and not associated with any view.

Fitzpasd:
@erictraut This is now updated with your suggestions.

Issue #142, opened 2017-06-21
[RN 0.45] Update getContentSizeMultiplier consumption


rigdern:
While submitting a pull request to Facebook, a couple of APIs got renamed. Currently ReactXP:
- Calls `UIManager.getContentSizeMultiplier` [(usage in ReactXP)](https://github.com/Microsoft/reactxp/blob/7c4ef262777211b9974436abe2ae5de81a57742f/src/native-common/UserInterface.ts#L93-L95)
- Listens to `didUpdateContentSizeMultiplier` event [(usage in ReactXP)](https://github.com/Microsoft/reactxp/blob/7c4ef262777211b9974436abe2ae5de81a57742f/src/native-common/UserInterface.ts#L26-L28)

These APIs no longer exist in React Native 0.45. Instead, the following APIs should be used:
- Read the font scale from `Dimensions.get('screen').fontScale`
- Listen to the `didUpdateDimensions` event and read the new font scale off of the event handler's argument.

PR #262, opened 2017-09-08
API cleanup before move to react-native-46

dryganets:


dryganets:
Fixes https://github.com/Microsoft/reactxp/issues/142

PR #263, opened 2017-09-09
Web Alert Improved Customization + Destructive Button Integration

MovingGifts:
I added more customization to the web alert and integrated destructive button. First pull request :)

### DIFF
--- a/docs/docs/apis/alert.md
+++ b/docs/docs/apis/alert.md
@@ -25,6 +25,9 @@ interface AlertButtonSpec {
 }
 
 interface AlertModalTheme {
+    // Overlay background style
+    overlayStyle?: StyleRuleSet<ViewStyle>;
+
     // Modal background style
     bodyStyle?: StyleRuleSet<ViewStyle>;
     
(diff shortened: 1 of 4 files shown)


erictraut:
I don't think the additional complexity is needed. This component is already more complex than it should be. If you really want that much customization, I recommend writing your own custom modal dialog. The reason we added a cancel button type is that it's supported in the native alerts. There's no such analog of a "destructive" button, so I don't think it makes sense to add such functionality only in the web implementation.

MovingGifts:
I thought it would be useful for others as well, but I understand what you mean. I just wanted to be able to use the RX.Alert cross platform syntax. For web, since we have a hard coded overlay bg color, I wasn't able to change the overlay color via the theme, but there was no possible way without adding it in the core. I was also trying to get a different button color for destructive from the theme, but couldn't. This just enabled doing both of those very easily from the theme if you wanted to (only for web since this is only a web component). It's useful if added to core as an option as it adds flexibility to the themeing, as it's optional and doesn't hurt but helps.

Feel free to close or respond :)

MovingGifts:
Hey, sorry to open this again, but it's about functionality of pressing an alert button and I added a code fix in my pull request (that would be the only functional change needed). When you click a  default button with an onPress, or a cancel button, they both should dismiss the alert. Currently, only the cancel button dismisses the alert. In native, any alert button dismisses the alert.

MovingGifts:
Should be:
```jsx
private _onPressButton(btnSpec: Types.AlertButtonSpec) {
  Modal.dismiss(this.props.modalId);
  if (btnSpec.onPress) {
      btnSpec.onPress();
  }
}
```
Instead of:
```jsx
private _onPressButton(btnSpec: Types.AlertButtonSpec) {
  if (btnSpec.style === 'cancel') {
    Modal.dismiss(this.props.modalId);
  } else if (btnSpec.onPress) {
    btnSpec.onPress();
  }
}
```

erictraut:
Yes, good point. I've checked in this change. Thanks!

PR #279, opened 2017-09-18
Added context menu support for text on web.

plisy:
Context menu support is needed when you want to react to contextmenu event within nested text events.

### DIFF
--- a/docs/docs/components/text.md
+++ b/docs/docs/components/text.md
@@ -48,8 +48,8 @@ numberOfLines: number = 0;
 selectable: boolean = false;
 
 // Mouse & Touch Events
-onPress?: (e: SyntheticEvent) => void;
-onContextMenu?: (e: SyntheticEvent) => void;
+onPress?: (e: SyntheticEvent) => void = undefined;
+onContextMenu?: (e: SyntheticEvent) => void = undefined;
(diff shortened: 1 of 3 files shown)


erictraut:
@plisy, why do we need an onContext on a Text node? Can't you just place it on a surrounding View node? I'm trying to keep our primitive types in ReactXP as lightweight as possible.

erictraut:
If there is truly no current workaround, we can add it, but your PR should include updates to the Text.md documentation in that case.

plisy:
Unfortunately no, since I have to render it within another text element. To do that I'd have to know size before rendering. I do understand why we require that and rendering text instead is already a workaround.

docs/docs/components/text.md
```diff
@@ -47,6 +47,10 @@ numberOfLines: number = 0;
 // Is the text selectable (affects mouse pointer and copy command)
 selectable: boolean = false;
 
+// Mouse & Touch Events
+onPress?: (e: SyntheticEvent) => void;
```
erictraut: Both of these have default values of "undefined". Please add "= undefined" at the end to indicate so. (The value after the "=" in the documentation indicates the default value.)

PR #280, opened 2017-09-18
Enabling various Typescript Compile Options (strict null checks, etc.)

deregtd:
+        "noImplicitReturns": true,
+        "noImplicitThis": true,
+        "noUnusedLocals": true,
+        "strictNullChecks": true,
+        "forceConsistentCasingInFileNames": true,


src/common/Types.ts
```diff
@@ -1103,10 +1103,10 @@ export interface KeyboardEvent extends SyntheticEvent {
 // ----------------------------------------------------------------------
 export var Children: React.ReactChildren;
 
-interface Element<P> {
-    type: React.ComponentClass<P>;
-    props: P;
-}
+// interface Element<P> {
```
ms-markda: Why not remove?

src/macos/Input.ts
```diff
@@ -7,7 +7,7 @@
 * MacOS implementation of Input interface.
 */
 
-import RN = require('react-native');
+//import RN = require('react-native');
```
ms-markda: Why not remove?

src/common/Types.ts
deregtd: It was probably wanted by someone originally.  Gets compiled away.

src/macos/Input.ts
deregtd: When they build this component, they'll want this line back.

src/common/Types.ts
```diff
@@ -488,7 +488,7 @@ export interface ImagePropsShared extends CommonProps {
     resizeMethod?: 'auto' | 'resize' | 'scale'; // Android only
     title?: string;
 
-    onLoad?: (size: Dimensions) => void;
+    onLoad?: (size: OptionalDimensions) => void;
```
erictraut: Why are these fields optional? Are there circumstances where RN can't hand back valid dimensions? I'd prefer if these are non-optional. Otherwise it complicates the logic in all components that use this callback.

src/common/Types.ts
```diff
@@ -938,7 +938,7 @@ export interface PopupOptions {
     // anchor has been pressed.
     // IMPORTANT NOTE: This handler may be called when the component is
     // already unmounted as it uses a time delay to accommodate a fade-out animation.
-    onAnchorPressed?: (e: RX.Types.SyntheticEvent) => void;
+    onAnchorPressed?: (e?: RX.Types.SyntheticEvent) => void;
```
erictraut: Why is this parameter optional? Seems like it should be required.

src/common/Types.ts
erictraut: If this is unused, you can delete it.

src/common/Types.ts
```diff
@@ -1115,6 +1115,10 @@ export type Dimensions = {
     width: number;
     height: number;
 };
+export type OptionalDimensions = {
```
erictraut: See note above. I don't think these fields should be optional.

src/common/Types.ts
```diff
@@ -1141,7 +1145,7 @@ export enum LinkingErrorCode {
 
 export interface LinkingErrorInfo {
     code: LinkingErrorCode;
-    url: string;
+    url: string|undefined;
```
erictraut: We recently changed links so their URL was required, so this should just be "string".

src/macos/Input.ts
erictraut: If this isn't needed, you can remove it.

src/native-common/FrontLayerViewManager.tsx
```diff
@@ -169,13 +169,14 @@ export class FrontLayerViewManager {
                                 bottom: y + height, width: width, height: height };
 
                         // Find out if the press event was on the anchor so we can notify the caller about it.
-                        if (touchEvent.pageX >= anchorRect.left && touchEvent.pageX < anchorRect.right
-                            && touchEvent.pageY >= anchorRect.top && touchEvent.pageY < anchorRect.bottom) {
+                        if (touchEvent.pageX && touchEvent.pageY &&
```
erictraut: These checks will not do the right thing if the values happen to be defined and equal to zero. I think you need to do explicit isDefined checks here.

src/native-common/GestureView.tsx
```diff
@@ -298,7 +298,8 @@ export abstract class GestureView extends ViewBase<Types.GestureViewProps, {}> {
         }
 
         // Has the user started to pan?
-        const panThreshold = this.props.panPixelThreshold > 0 ? this.props.panPixelThreshold : _panPixelThreshold;
+        const panThreshold = (this.props.panPixelThreshold && this.props.panPixelThreshold > 0) ?
```
erictraut: The logic happens to work OK if the value is defined and equal to zero, but it would be better to make the check for undefined explicit.

src/native-common/GestureView.tsx
```diff
@@ -308,7 +309,8 @@ export abstract class GestureView extends ViewBase<Types.GestureViewProps, {}> {
...
```
erictraut: See comment above.

src/native-common/GestureView.tsx
```diff
@@ -323,7 +325,8 @@ export abstract class GestureView extends ViewBase<Types.GestureViewProps, {}> {
...
```
erictraut: See comment above.

src/native-common/GestureView.tsx
```diff
@@ -207,13 +207,13 @@ export abstract class GestureView extends ViewBase<Types.GestureViewProps, {}> {
...
-        const initialPageX = this._lastGestureStartEvent.pageX;
-        const initialPageY = this._lastGestureStartEvent.pageY;
+        const initialPageX = this._lastGestureStartEvent.pageX || 0;
+        const initialPageY = this._lastGestureStartEvent.pageY || 0;
 
         const timeStamp = this._getEventTimestamp(e);
 
         return (timeStamp - initialTimeStamp <= _tapDurationThreshold &&
-            this._calcDistance(initialPageX - e.pageX, initialPageY - e.pageY) <= _tapPixelThreshold);
+            this._calcDistance(initialPageX - (e.pageX || 0), initialPageY - (e.pageY || 0)) <= _tapPixelThreshold);
```
erictraut: For these types of events, we're guaranteed that pageX and pageY are defined. So you could keep the code as is or change it to e.pageX!!!, etc.

src/web/AlertModalContent.tsx
```diff
@@ -131,9 +132,9 @@ export class AlertModalContent extends RX.Component<AppModalContentProps, AppMod
                         onPress={ e => this._onPressButton(btnSpec) }
                         onHoverStart={ () => this.setState({ hoverIndex: i }) }
                         onHoverEnd={ () => this.setState({ hoverIndex: -1 }) }
-                        style={ buttonStyle }
+                        style={ _.compact(buttonStyle) }
```
erictraut: Why are you compacting these styles? It's perfectly legit for style arrays to contain undefined values.

src/web/GestureView.tsx
```diff
@@ -188,7 +188,7 @@ export class GestureView extends RX.ViewBase<Types.GestureViewProps, {}> {
     }
 
     private _getPanPixelThreshold = () => {
-        return this.props.panPixelThreshold > 0 ? this.props.panPixelThreshold : _panPixelThreshold;
+        return (this.props.panPixelThreshold && this.props.panPixelThreshold > 0) ? this.props.panPixelThreshold : _panPixelThreshold;
```
erictraut: Zero values work OK in this logic, but it would be better if the check for undefined was more explicit.

src/web/RootView.tsx
```diff
@@ -462,9 +463,9 @@ export class RootView extends React.Component<RootViewProps, RootViewState> {
     private _startHidePopupTimer() {
         if (this.props.autoDismiss) {
             // Should we immediately hide it, or did the caller request a delay?
-            if (this.props.autoDismissDelay > 0) {
+            if (this.props.autoDismissDelay && this.props.autoDismissDelay > 0) {
```
erictraut: Zero values work OK with this logic, but an explicit undefined type would be better.

src/web/Storage.ts
```diff
@@ -14,7 +14,7 @@ import RX = require('../common/Interfaces');
 export class Storage extends RX.Storage {
     getItem(key: string): SyncTasks.Promise<string|undefined> {
         const value = window.localStorage.getItem(key);
-        return SyncTasks.Resolved<string|undefined>(value);
+        return SyncTasks.Resolved<string|undefined>(value || undefined);
```
erictraut: Why do you want to turn empty strings into undefined here? I don't think that's the right semantic.

src/common/Types.ts
deregtd: talked offline

PR #282, opened 2017-09-20
Exposing attributes to make accessible menus and listboxes

piotrowskiLuc:
Added trait 'Option' in order to have role=option for listboxes.
Exposing attributes:
id, aria-labelledBy ,aria-haspopup, aria-controls to make accessible menus.

### DIFF
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -547,6 +547,8 @@ export interface TextPropsShared extends CommonProps {
     elevation?: number;
 
     onPress?: (e: SyntheticEvent) => void;
+
+    id?: string; // Web only. Needed for accessibility.
 }
 
 export interface TextProps extends TextPropsShared {
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -370,6 +370,10 @@ export interface AccessibilityHtmlAttributes extends React.HTMLAttributes {
     'aria-hidden'?: boolean;
     'aria-disabled'?: boolean;
     'aria-selected'?: boolean;
+    'aria-checked'?: boolean;
+    'aria-haspopups'?: boolean;
+    'aria-controls'?: string;
+    'aria-labelledby'?: string;
 }
 
 // Android & Desktop supported prop, which allows screen-reader to inform its users when a
@@ -427,6 +431,7 @@ export enum AccessibilityTrait {
     Log,
     Status,
     Dialog,
+    HasPopup,
 
     // Desktop & mobile. This is at the end because this
     // is the highest priority trait.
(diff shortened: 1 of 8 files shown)


src/common/Types.ts
```diff
@@ -463,6 +469,10 @@ export interface ButtonProps extends CommonStyledProps<ButtonStyleRuleSet>, Comm
     disableTouchOpacityAnimation?: boolean;
     activeOpacity?: number;
     underlayColor?: string;
+
+    // Web only.
+    id?: string; // Needed for accessibility to be able to use labelledBy attribute.
```
erictraut: Please correct spelling. Should be "labeledBy".

src/common/Types.ts
```diff
@@ -562,6 +574,8 @@ export interface ViewPropsShared extends CommonProps, CommonAccessibilityProps {
     limitFocusWithin?: boolean; // Web-only, make the view and all focusable subelements not focusable when isFocusLimited state is true
 
     importantForLayout?: boolean; // Web-only, additional invisible DOM elements will be added to track the size changes faster
+    id?: string; // Web-only. Needed for accessibility.
+    ariaLabelledby?: string; // Web-only. Needed for accessibility.
```
erictraut: Please correct spelling. Should be "ariaLabeledBy".

src/web/AccessibilityUtil.ts
```diff
@@ -95,6 +95,17 @@ export class AccessibilityUtil extends CommonAccessibiltiyUtil {
         // as we dont want to pollute the dom with "aria-checked = false" for every falsy condition
         return undefined;
     }
+
+    accessibilityTraitToAriaHasPopup(traits: Types.AccessibilityTrait | Types.AccessibilityTrait[]|undefined) {
```
erictraut: Nit: For consistency, add spaces around "|" operator.

src/common/Types.ts
```diff
@@ -547,6 +547,8 @@ export interface TextPropsShared extends CommonProps {
     elevation?: number;
 
     onPress?: (e: SyntheticEvent) => void;
+
+    id?: string; // Web only. Needed for accessibility.
```
erictraut: Please add to appropriate documentation (".md") file. Props are listed in alphabetical order in the documentation. Docs also explain the default value. I think the default value in this case is undefined.

src/common/Types.ts
piotrowskiLuc: I've addressed the comments and added documentation to respective *.md files

mshoho: In the standard it's double «l» — https://www.w3.org/TR/wai-aria/states_and_properties#aria-labelledby

piotrowskiLuc: Lets wait with merging it. I think I still need to expose aria-expanded attribute

piotrowskiLuc: Actually lets go without aria-expanded for now. We can add it later if we see it's needed.

PR #298, opened 2017-10-06
Adding a timeout before focusing previous component.

piotrowskiLuc:
Adding this timeout, because of screen readers keeping reference to old aria-label and not reading correct state of components when the label changes dynamically.


piotrowskiLuc:
Hi Eric, can we please backport this one? One ship blocker depends on this.

PR #302, opened 2017-10-09
Update Popup API docs to include  missing method and return type of s…

Fitzpasd:
…how()


Fitzpasd:
**NOTE:**  This is likely to need to be updated again based on #304

PR #303, opened 2017-10-09
Adding the Android 'mode' property to Picker

Fitzpasd:
This is a very widely used property for Android.  In fact, the default of 'dialog' is much less common in my experience.

Without this addition, there is no way to have the common dropdown spinner on Android.

### DIFF
--- a/docs/docs/components/picker.md
+++ b/docs/docs/components/picker.md
@@ -29,6 +29,11 @@ selectedValue: string;
 onValueChange: (itemValue: string, itemPosition: number) => void;
 
 style: PickerStyleRuleSet | PickerStyleRuleSet[] = [];
+
+// Android only.
+// 'dialog': Show a modal dialog. This is the default.
+// 'dropdown': Shows a dropdown anchored to the picker view
+mode: string;
 ```
 
 ## Styles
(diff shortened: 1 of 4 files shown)


docs/docs/components/picker.md
```diff
@@ -29,6 +29,11 @@ selectedValue: string;
 onValueChange: (itemValue: string, itemPosition: number) => void;
 
 style: PickerStyleRuleSet | PickerStyleRuleSet[] = [];
+
+// Android only.
+// 'dialog': Show a modal dialog. This is the default.
+// 'dropdown': Shows a dropdown anchored to the picker view
+mode: string;
```
erictraut: Please define this as a 'dialog'|'dropdown' rather than a string. This provides better type checking.

src/common/Types.ts
```diff
@@ -490,6 +490,7 @@ export interface PickerProps extends CommonProps {
     selectedValue: string;
     onValueChange: (itemValue: string, itemPosition: number) => void;
     style?: StyleRuleSetRecursive<PickerStyleRuleSet>;
+    mode?: string;
```
erictraut: Same as above. Please define as 'dialog'|'dropdown'.

Issue #301, opened 2017-10-09
Why does Popup.show() return false for invalid args, rather than throw an Error?


Fitzpasd:
When invoking `Popup.show()`, the logic returns a `false` if the input args are incorrect.  

```
// FrontLayerViewManager.tsx

if (!popupId || popupId === '') {
  console.error('FrontLayerViewManager: popupId must be valid!');
  return false;
}

if (!popupOptions.getAnchor()) {
  console.error('FrontLayerViewManager: getAnchor() must be valid!');
  return false;
}
```

Why is this preferred rather than throwing an `Error`?  I personally prefer the 'fail-fast' approach for libraries like this, both as an author and a consumer, so would like to understand why this approach is used instead.

erictraut:
Yeah, I agree with you. A throw in this case would be preferable given that these conditions are never valid and are easily verified through simple testing. For more subtle or dynamic conditions, we try to avoid throwing exceptions because it forces callers to always include try/catch statements to protect against app crashes.

Fitzpasd:
Thanks for the clarification.  Didn't want to create a PR without knowing whether or not this is a conscious design decision cross project.  

I'll create one tomorrow.

PR #304, opened 2017-10-10
Moving argument validation from FrontLayerViewManager to public Popup…

Fitzpasd:
… and Modal APIs.

Issue #301. 

This moves (and adds some) argument validation from the `FrontLayerViewManager` to the `Popup` and `Modal` API entry points.  This is a better place for these IMO, since we are failing quicker on arguments we know are invalid, and providing a nicer stacktrace for consumers. 

I choose these validations (rather than validate every possible value/field) because they are the ones that the API will hard-fail on if not satisfied.

### DIFF
--- a/src/web/Popup.ts
+++ b/src/web/Popup.ts
@@ -12,16 +12,27 @@ import RX = require('../common/Interfaces');
 import Types = require('../common/Types');
 
 export class Popup extends RX.Popup {
-
     show(options: Types.PopupOptions, popupId: string, delay?: number): boolean {
+        if (!popupId || popupId === '') {
+            throw new Error(`popupId must be a non-empty string. Actual: ${popupId}`);
+        }
+
         return FrontLayerViewManager.showPopup(options, popupId, delay);
     }
 
     autoDismiss(popupId: string, delay?: number): void {
+        if (!popupId || popupId === '') {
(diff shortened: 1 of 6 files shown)


src/web/Popup.ts
```diff
@@ -12,16 +12,27 @@ import RX = require('../common/Interfaces');
 import Types = require('../common/Types');
 
 export class Popup extends RX.Popup {
-
     show(options: Types.PopupOptions, popupId: string, delay?: number): boolean {
+        if (!popupId || popupId === '') {
```
a-tarasyuk: Maybe would be better move this logic to separate method? And will be one method with the same logic. What do you think, does it make sense?

```
private isEmpty(popupId: string) {
   if (!popupId || popupId === '') {
      throw new Error(`popupId must be a non-empty string. Actual: ${popupId}`);
   }
}
```

Fitzpasd: In this use case (parameter validation), I think the better stack trace is worth the small code duplication.

If we were to start extracting these into separate functions, then I'd advocate for introducing a lib to handle that consistently across the project, e.g. [validate](https://github.com/mdvorscak/validate-params), [check-preconditions](https://github.com/pnann/check-preconditions)

Issue #289, opened 2017-09-25
RX.TextInput default value always returns error/warning


MovingGifts:
When trying to set a default value on page load, I always get this error:
![screen shot 2017-09-25 at 6 00 21 pm](https://user-images.githubusercontent.com/1202909/30833087-8ebf60fa-a21b-11e7-88b3-cf946deb0b17.png)

This is the simplest example that produces the error:
```jsx
import React from 'react'
import RX from 'reactxp'

export default class TextInputValue extends RX.Component {
  constructor(props) {
    super(props);

    this.state = {
        inputValue: 'hello'
    };
  }

  onChangeTextValue = (value) => {
    this.setState({ inputValue: value });
  }

  render() {
    return (
      <RX.TextInput
        placeholder="Placeholder"
        value={ this.state.inputValue }
        onChangeText={this.onChangeTextValue}
        // defaultValue={ this.state.inputValue }
        />
    );
  }
}
```
I tested as is (with `value`) alone and also tested with `defaultValue` alone, but no matter what, I keep getting that error. I would rather not use the `defaultValue`  and instead have the `value` prop pull in the default text from state. That way I can reuse the same field for inputting text as well as editing it, with the `value` being assigned its state from `inputValue`.

Any idea how to fix this to get rid of the pesky error?

deregtd:
What version of react are you using?

erictraut:
I suspect that the RN.TextInput's interface changed in a breaking way. We'll need to update RX's TextInput implementation accordingly. Which version of React Native are you using?

erictraut:
Hmm, interesting. The RX implementation is providing an onChangeText handler but not an onChange handler. The documentation doesn't make any distinction between the two other than that onChangeText is passed the updated text. I suspect that the warning is a recently-introduced bug in RN. Once I know which version you're using, I'll investigate further.

MovingGifts:
Thanks for the response guys, here is the relevant info:
```
"react": "^16.0.0-rc.3",
"react-dom": "^16.0.0-rc.3",
"react-native": "0.46.0"
```

MovingGifts:
I think this issue also occurred with the previous 0.42.0 version of react native as well.
Any update on the fix eta?

MovingGifts:
@erictraut Any update on the investigation of the error and a possible solution?

MovingGifts:
[...]

PR #305, opened 2017-10-10
Added defaultValue to fix React Text Input error

MovingGifts:
Fixes https://github.com/Microsoft/reactxp/issues/289


MovingGifts:
This fix seems to create another error about controlled vs uncontrolled inputs if there is an onChangeText mixed with a defaultValue.

Sorry, I thought I solved it but it seems that I really don't know how to fix this issue.

Issue #216, opened 2017-08-21
ReactNative >= 0.44?


vinils:
Hey guys, I´ve seen that there is some break changes on react native >= 0.44 but i couldn't find any feed back on when that will be done.
I´m waiting for >= 0.45 and just kindle would like to know how that is going on? Any news or an idea when that will happen? 

Tks

ms-deregtd:
We're internally testing bumping our own software up to 0.46, but haven't yet found anything we've needed to change in ReactXP.  Is there something you're running into that's not working on a newer RN?

vinils:
Don´t wanna make this issue very complex but essentially I´m trying to integrate with expo for many good reasons and my problem is to fix this reference's puzzle.

I´ve seen some issues about break changes on 0.44 opened here (example: @erictraut mentioned on https://github.com/Microsoft/reactxp/issues/83).

Just would like to know if and when would be possible to run reactxp with RN 0.45 and react "16.0.0-alpha.12"?

My project has those reference below and expo with typescript requires 0.45 because of ["react-native-typescript-transformer"](https://github.com/ds300/react-native-typescript-transformer).

  "devDependencies": {
    "@types/node": "^7.0.12",
    "@types/webpack": "^2.2.14",
    "awesome-typescript-loader": "3.1.2",
    "source-map-loader": "^0.1.6",
    "ts-node": "^3.0.2",
    "typescript": "2.1.6",
    "webpack": "2.2.1"
  },
  "dependencies": {
    "react": "^15.4.1",
    "react-dom": "^15.4.1",
    "react-native": "^0.42.0",
    "react-native-windows": "^0.33.0",
    "reactxp": "0.42.0-rc.9",
    "reactxp-imagesvg": "^0.2.1",
    "reactxp-video": "^0.1.3",
    "reactxp-virtuallistview": "^0.1.3",
    "simplerestclients": "0.0.18"
  }

Expo references
  "dependencies": {
    "expo": "15.0.0",
    "react": "15.0.0",
    "react-native": "https://github.com/expo/react-native/archive/sdk-15.0.0.tar.gz"
  },
  "devDependencies": {
    "@types/react": "15.0.0",
    "@types/react-native": "0.42.0",
    "react-native-typescript-transformer": "1.0.13",
    "typescript": "2.2.2"
  }
}

Any suggestion are very welcome :)

ms-deregtd:
[...]

PR #306, opened 2017-10-11
 Fixing undefined is not an object error

kgroat:
When using `reactxp` 0.42.x with `react-native` 0.49+, the following error occurs: `undefined is not an object (evaluating 'Navigation.StateUtils')`.  This is because of the fact that `NavigationExperimental` is no longer a part of `react-native`.

This appears to be caused by [Line 30 of `NavigatorExperimentalDelegate`](https://github.com/Microsoft/reactxp/blob/0.42.0/src/native-common/NavigatorExperimentalDelegate.tsx#L30).  This was even occurring if the Navigation portions of `reactxp` were unused.

This is the initial issue talked about in #216.

This PR aims to fix the error by deferring the evaluation of `Navigation.StateUtils` until the delegate is actually used.  This way, it doesn't cause any backwards compatibility issues, but also fixes the error.

### Screenshot:
undefined is not an object

### DIFF
--- a/src/native-common/NavigatorExperimentalDelegate.tsx
+++ b/src/native-common/NavigatorExperimentalDelegate.tsx
@@ -27,8 +27,6 @@ type NavigationState = Navigation.NavigationState;
 type NavigationRoute = Navigation.NavigationRoute;
 type NavigationTransitionProps = Navigation.NavigationTransitionProps;
 
-const StateUtils = Navigation.StateUtils;
-
 interface NavigationRouteState extends NavigationRoute {
     route: Types.NavigatorRoute;
 }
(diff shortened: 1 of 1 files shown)


erictraut:
Have you tried with ReactXP 0.46.x? I wouldn't expect 0.42.x to work with RN 0.49+. There have been several breaking changes made to RN, and we've addressed these in the master branch. I don't want to go back and patch 0.42.x at this point.

kgroat:
Thanks for the quick response.
I hadn't tried 0.46.x.  I had figured since NPM showed 0.42.7 as latest, that it was the highest version available, but now see that 0.46.3 is, and it works like a charm. Closing.

PR #308, opened 2017-10-11
UWP: Added Accessibility.highContrastChangedEvent

antonkh:
For now it only tells if high contrast is enabled or disabled in system settings.
Later it can tell what are the chosen high contrast colors: Windows 10 allows
to create custom high contrast themes with user chosen colors.

### DIFF
--- a/src/common/Accessibility.ts
+++ b/src/common/Accessibility.ts
@@ -14,6 +14,10 @@ import RX = require('../common/Interfaces');
 export abstract class Accessibility extends RX.Accessibility {
     abstract isScreenReaderEnabled(): boolean;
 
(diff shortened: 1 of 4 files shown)


antonkh:
The corresponding change in react-native-windows: https://github.com/Microsoft/react-native-windows/pull/1443

erictraut:
@antonkh You added this to the 0.42.0 branch. You should always submit reactxp PRs on master first, and we can decide if/when to backport them once they're merged into master.

erictraut:
In this case, I've cherry-picked the change in the opposite direction (back to master), so we're good.

antonkh:
Thanks

PR #315, opened 2017-10-17
Fix accessibility props merging to _internalProps

isnotgood:
In case when RX accessibility prop name collides with RN.View prop,
and type differs, then RX value was passed untranslated to RN.View
which caused propTypes validation to fail.

### DIFF
--- a/src/native-common/View.tsx
+++ b/src/native-common/View.tsx
@@ -42,24 +42,20 @@ export class View extends ViewBase<Types.ViewProps, {}> {
         this._internalProps.style = this._getStyles(props);
         this._internalProps.ref = this._setNativeView;
 
-        const importantForAccessibility = AccessibilityUtil.importantForAccessibilityToString(props.importantForAccessibility);
-        const accessibilityLabel = props.accessibilityLabel || props.title;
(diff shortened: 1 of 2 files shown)


src/native-common/View.tsx
```diff
@@ -42,24 +42,20 @@ export class View extends ViewBase<Types.ViewProps, {}> {
         this._internalProps.style = this._getStyles(props);
         this._internalProps.ref = this._setNativeView;
 
-        const importantForAccessibility = AccessibilityUtil.importantForAccessibilityToString(props.importantForAccessibility);
-        const accessibilityLabel = props.accessibilityLabel || props.title;
-        // Set accessibility props on Native only if we have valid importantForAccessibility value or accessibility label or
-        // if this view is not a button.
-        // For a button, we let the Button component compute its own accessibility props.
-        const shouldSetAccessibilityProps = this._internalProps && !this._isButton(props) &&
-            !!(importantForAccessibility || accessibilityLabel);
```
isnotgood: This part may introduce regression. But I believe it doesn't make sense as accessibilityTraits are copied by `this._internalProps = _.clone(props) as any;` anyway.

isnotgood:
Note: we need to back port this to 0.42.

PR #325, opened 2017-10-24
UWP: Added properties related to AcrylicBrush (Fluent)

antonkh:
The new properties can be used once they are added to RNW:
https://github.com/Microsoft/react-native-windows/pull/1467

### DIFF
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -194,9 +194,9 @@ export interface ViewStyle extends ViewAndImageCommonStyle {
     elevation?: number; // Android only
     // Windows 10 RS3 supports acrylic brushes. In earlier versions these properties are ignored.
     // The tint opacity can be set either with acrylicOpacity or with acrylicTintColor, e.g. #f002.
-    acrylicOpacity?: number; // UWP only; default = 1
-    acrylicSource?: 'host' | 'app'; // UWP only; default = "host"
-    acrylicTintColor?: string; // UWP only; default = backgroundColor
+    acrylicOpacityUWP?: number; // UWP only; default = 1
+    acrylicSourceUWP?: 'host' | 'app'; // UWP only; default = "host"
+    acrylicTintColorUWP?: string; // UWP only; default = backgroundColor
 }
 
 export type ViewStyleRuleSet = StyleRuleSet<ViewStyle>;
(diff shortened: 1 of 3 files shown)


src/common/Types.ts
```diff
@@ -192,6 +192,11 @@ export interface ViewStyle extends ViewAndImageCommonStyle {
     // http://facebook.github.io/react-native/releases/0.30/docs/shadow-props.html (iOS only)
     // http://facebook.github.io/react-native/releases/0.30/docs/view.html#style (see elevation property)
     elevation?: number; // Android only
+    // Windows 10 RS3 supports acrylic brushes. In earlier versions these properties are ignored.
+    // The tint opacity can be set either with acrylicOpacity or with acrylicTintColor, e.g. #f002.
+    acrylicOpacity?: number; // UWP only; default = 1
```
erictraut: New style attributes should be added to the documentation. See Styles.md.

Also, consider renaming these to end in "UWP" since they're UWP specific. We have tried to do this where possible (e.g. "shouldRasterizeIOS"). React Native is somewhat inconsistent on this convention though, so ReactXP reflects this inconsistency.

antonkh:
Should `translateX` anf `translateY` properties be `number`s in the md file? They are `string`s now.

antonkh:
There is a possible problem with the new styles: RN for Windows crashes the app if unknown properties are set, so probably do RN for iOS and Android. It would be nice if (1) RX could map the new acrylic styles to props recognized by RN windows (it could remove the UWP suffix) and (2) RX would remove the acrylic props if it's not on Windows. Otherwise apps using RX would have to write `if (isWindows) styles = ...; else styles = ...;`

erictraut:
@antonkh, good catch on translateX/Y. Yes, they should be documented as numbers to match the code. I'll fix.

erictraut:
@antonkh, We already have code to omit style attributes that don't apply to a particular platform. See the forbiddenProps method in native-common/Styles.ts. Currently, all versions of RN accept style attributes for all other version of RN, so we're currently stripping off only the attributes that are web-specific (e.g. wordBreak and appRegion). We may need to start doing platform-specific style attribute filtering.

antonkh:
What do you think about mapping properties to platform specific names, e.g. `acrylicOpacityUWP` -> `acrylicOpacity`?

erictraut:
I'd prefer not to add a mapping step. That adds unnecessary runtime overhead. Plus, it would deviate from the standard that we've used so far where we always adopt the RN name as-is even if it isn't entirely consistent in its naming convention.

antonkh:
We'll need this in `0.42.0` eventually. Do I need to send another PR to cherry-pick the commit?

erictraut:
Skype is going to switch to 0.46.x soon. I’d prefer not to back-port this to 0.42.x. Interface changes generally should be added only to newer versions. We could make an exception here since it's not a breaking change.

antonkh:
We aren't planning to use the acrylic styles anytime soon in S4L, so if this commit appears in 0.46, we don't need to do anything right now.

PR #339, opened 2017-11-01
Fixed crash due to throttled onscroll callback called after unmount.

ms-markda:


ms-markda:
Fixed crash due to throttled onscroll callback called after unmount.

PR #340, opened 2017-11-01
Fixed crash due to throttled onscroll callback called after unmount.

ms-markda:


erictraut:
@ms-markda, looks like you fixed this only in the 0.42.0 branch. You should submit a fix for master first, and we can back-port it to 0.42.0 once that's merged.

PR #341, opened 2017-11-02
Fix double "Code" decoration on code regions.

zholobov:
Fix double "Code" decoration on code regions hiding first line of code region.
image
image


erictraut:
Thanks for fixing. This must have been due to a recent change in github's code.

PR #351, opened 2017-11-10
Adding prettier configuration

mfrachet:
Adding prettier configuration & pre-commit hook to provide the same code formatting for every contributors

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,13 +1,17 @@
 {
   "name": "reactxp",
   "version": "0.46.5",
-  "description": "Cross-platform abstraction layer for writing React-based applications a single time that work identically across web, React Native, and Electron distribution",
+  "description":
+    "Cross-platform abstraction layer for writing React-based applications a single time that work identically across web, React Native, and Electron distribution",
   "author": "ReactXP Team <reactxp@microsoft.com>",
   "license": "MIT",
   "scripts": {
     "build": "npm run tslint && tsc",
-    "tslint": "tslint --project tsconfig.json -r tslint.json -r ./node_modules/tslint-microsoft-contrib --fix || true"
+    "tslint":
+      "tslint --project tsconfig.json -r tslint.json -r ./node_modules/tslint-microsoft-contrib --fix || true",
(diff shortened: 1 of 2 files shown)


erictraut:
@mfrachet, we already use tslint to enforce a consistent code style. I'm not familiar with "prettier". What does it do that tslint doesn't? Why do you think we need both?

mfrachet:
I suggest you reading this doc from prettier : 

https://github.com/prettier/prettier/blob/master/docs/why-prettier.md

The main idea is that prettier will auto-format the code using a determinist approach on each file. 

This way, it's guaranteed that everybody will use the same code formatting, not only the code style, but the spaces between each function, same indent everywhere (and so much more 😄 ) 

It guarantees no unnecessary diffs between PRs and ensure a bit more code formatting.

From this PR, you can check the result of a prettier 'shot' by running :

```
$ npm run format
```

You'll be able to see a first prettier transformation to give you a better (and concrete) idea of what's happening.

Important thing : `prettier` is able to take `tslint` configuration in account, so that automatic code formatting respects the rules.

It's quite opinionated, I find prettier to be a good solution to really enforce code formatting for open sources (with many contributors like ReactXP).

#### Important

This PR is not totally finished. I have to adjust some stuff to make `prettier` take care of `tslint` rules if this insterest you.

erictraut:
I'm potentially interested. As you pointed out, in its current form, it doesn't honor the code style rules we currently use in ReactXP, so it's difficult to tell what (if any) desired formatting changes it's making. All of the changes I reviewed are undesired (e.g. changing from four-space tabs to two-space). If you're able to make it honor the current code style, I'd be interested.

mfrachet:
I'll make this work with your tslint rules :)

mfrachet:
I've studied the subject this week end. It seems that the tslint configuration of this project has too many conflicts with prettier rules.

I'll study a bit more this in the next few weeks to check what it's possible to do.

For now, I m closing this PR as it's not relevant anymore...

Issue #336, opened 2017-11-01
Embedded Buttons aren't accessible


Fitzpasd:
We've built the below component: 

![Imgur](https://i.imgur.com/TyHpruC.png)

It's composed of a root button (highlighted in the picture) and two smaller buttons; Accept and 'X'.  We have different `onPress` handlers for each, and they all work as expected.

**Issue**: The root button steals the accessibility focused event.  There's no way to highlight the inner buttons if pressed.

Our code for this is basically:

```
<RX.Button>
  <RX.View>
    ...
    <RX.Button />
  </RX.View>
  <RX.View>
    ...
    <RX.Button />
  </RX.View>
</RX.Button>
```

Fitzpasd:
Note:  You can overcome this by not embedding buttons and using styles to overlay the components, but I think this is something the platform should support, given the current API design.

erictraut:
You can't nest buttons like that. On the web, buttons are translated into <button/> HTML nodes, and HTML doesn't permit nesting buttons. You can nest a Button within a View that accepts focus and allows for clicks/taps. I think that's how we handle cases like this in the Skype app. Let me consult with one of my team members who is more familiar with this area, and one of us will get back to you.

msftsvachon:
As Eric mentioned, we've been using a parent View in such cases since Buttons can't be nested and we'd set the proper handlers/traits on it. We have seen issues in some cases with screen readers with this configuration though and have been trying to avoid this nesting whenever possible. In this specific case, do you need the parent "button" to receive accessibility focus or could you let it flow to its content? If so, one option would be to set importantForAccessibility = { RX.Types.ImportantForAccessibility.No } on the root view so accessibility focus moves to the inner elements.

Fitzpasd:
Makes sense, should the platform throw a warning or exception to help developers avoid this scenario? 

For us, I do need all 3 buttons to be focusable and accessible.  Worked around it by refactoring to: 

```
<RX.View>
  <RX.Button/>
  <RX.Button/>
  <RX.Button/>
</RX.View>
```

With styling to give the embedded appearance.

erictraut:
The browser (at least Chrome) provides emits a console warning if it encounters nested buttons, but providing a warning within the platform is an interesting idea to consider.

Fitzpasd:
I'll write an implementation for that when I have some time.

PR #362, opened 2017-11-16
Add a development warning when using <Button> with more than one chil…

Fitzpasd:
…d, or with another <Button> in the subtree.

This addresses #336.


erictraut:
I think this detection mechanism has a hole in it. A child component could introduce an RX.Button in subsequent renders without modifying the props of the containing RX.Button. Such cases would go unnoticed. 

Did you instead consider using "context" to pass a value down the hierarchy? I used this approach in the RX.Text implementation (search for "isRxParentAText"). I think this would close the hole, and it wouldn't involve recursively walking the tree.

Fitzpasd:
Good idea.  Using PropTypes and the context API cleans this up a lot.

PR #363, opened 2017-11-16
Fixing tsc and tslint errors in samples

Fitzpasd:


erictraut:
Good catches. Thanks for the fixes.

PR #366, opened 2017-11-17
Jofuria/scroll to click position

jofuria:
Handle mouse down events on CustomScrollbar.
When clicking the scrollbar track the thumb will scroll to that position.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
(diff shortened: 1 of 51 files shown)


erictraut:
@jofuria, looks like you were working with a really old version of the sources. You'll need to rebase and resolve the merge conflicts.

PR #367, opened 2017-11-18
Handle mouse down event on scrollbar

jofuria:
Handle mouse down event.
The scroll bar will scroll to the position the user has clicked.


src/web/CustomScrollbar.ts
```diff
@@ -329,7 +329,24 @@ export class Scrollbar {
     }
 
     private _handleMouseDown(e: React.MouseEvent<any>) {
-        this._prevent(e);
+        const target = e.currentTarget as any;
+
+        if (this._dragging === true || e.target === this._verticalBar.slider || !target) {
```
erictraut: No need to compare against "true" because "_dragging" is already a boolean.

src/web/CustomScrollbar.ts
```diff
@@ -329,7 +329,24 @@ export class Scrollbar {
...
+            this._prevent(e);
+            return;
+        }
+
+        if (this._hasVertical) {
+            const eventOffsetY = e.pageY - target.getBoundingClientRect().top;
+            const halfHeight = this._verticalBar.slider.offsetHeight / 2; 
+            const offsetY = (eventOffsetY - this._verticalBar.slider.offsetTop - halfHeight) * this._verticalBar.slider2Scroll;
+            this._viewport.scrollTop = offsetY + this._viewport.scrollTop;
+        } else if (this._hasHorizontal) {
```
erictraut: What if it has both a horizontal and vertical? I think the "else" should be removed.

src/web/CustomScrollbar.ts
```diff
@@ -329,7 +329,24 @@ export class Scrollbar {
...
+            const eventOffsetX = e.pageX - target.getBoundingClientRect().left;
+            const halfWidth = this._horizontalBar.slider.offsetWidth / 2; 
+            const offsetX = (eventOffsetX - this._horizontalBar.slider.offsetLeft - halfWidth) * this._verticalBar.slider2Scroll;
```
erictraut: This should use this._horizontalBar, not this._verticalBar.

jofuria: argh old version

PR #368, opened 2017-11-18
Handle on mouse click on scroll bar

jofuria:


src/web/CustomScrollbar.ts
```diff
@@ -329,7 +329,24 @@ export class Scrollbar {
     }
 
     private _handleMouseDown(e: React.MouseEvent<any>) {
-        this._prevent(e);
+        const target = e.currentTarget as any;
+
+        if (this._dragging === true || !target) {
```
erictraut: It's redundant to compare a boolean value to the value "true", and it makes the code less readable.

src/web/CustomScrollbar.ts
```diff
@@ -329,7 +329,24 @@ export class Scrollbar {
...
+            this._prevent(e);
+            return;
+        }
+
+        if (this._hasVertical) {
+            const eventOffsetY = e.pageY - target.getBoundingClientRect().top;
+            const halfHeight = this._verticalBar.slider!!!.offsetHeight / 2; 
+            const offsetY = (eventOffsetY - this._verticalBar.slider!!!.offsetTop - halfHeight) * this._verticalBar.slider2Scroll!!!;
+            this._viewport.scrollTop = offsetY + this._viewport.scrollTop;
+        } else if (this._hasHorizontal) {
```
erictraut: This won't work if the view has both a vertical and a horizontal scroll bar. The "else" should be removed.

src/web/CustomScrollbar.ts
```diff
@@ -329,7 +329,24 @@ export class Scrollbar {
...
+            const eventOffsetX = e.pageX - target.getBoundingClientRect().left;
+            const halfWidth = this._horizontalBar.slider!!!.offsetWidth / 2; 
+            const offsetX = (eventOffsetX - this._horizontalBar.slider!!!.offsetLeft - halfWidth) * this._verticalBar.slider2Scroll!!!;
```
erictraut: Should use "this._horizontalBar" here, not "this._verticalBar".

PR #377, opened 2017-12-08
Fixed drag & drop handlers for RN windows

antonkh:
With this change, if the app sets either of the four drag and drop handlers on `RX.View`,
the user will be able to drag files into the view. The event will have the `dataTransfer`
property:

```tsx
<RX.View onDragEnter={ e => console.log(e.dataTransfer.files.length) }>
  ...
</RX.View>
```

This change depends on https://github.com/Microsoft/react-native-windows/pull/1549
as it sets a new `allowDrop` property on `RN.View`. If an older RNW version is used, the
`allowDrop` property will be ignored and drag & drop won't work, but the app won't crash.

### DIFF
--- a/src/native-common/View.tsx
+++ b/src/native-common/View.tsx
@@ -80,7 +80,7 @@ export class View extends ViewBase<Types.ViewProps, {}> {
             const handler = this._internalProps[name];
 
             if (handler) {
-                this._internalProps.allowDrop = true; // TODO: let RN figure this out?
+                this._internalProps.allowDrop = true;
 
                 this._internalProps[name] = (e: React.SyntheticEvent) => {
                     handler({
--- a/src/native-common/View.tsx
+++ b/src/native-common/View.tsx
@@ -18,7 +18,6 @@ import AnimateListEdits from './listAnimations/AnimateListEdits';
 import Button from './Button';
 import Types = require('../common/Types');
 import ViewBase from './ViewBase';
-import { SyntheticEvent } from 'react';
 
 export class View extends ViewBase<Types.ViewProps, {}> {
     private _internalProps: any = {};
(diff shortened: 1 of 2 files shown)


src/native-common/View.tsx
```diff
@@ -69,29 +69,61 @@ export class View extends ViewBase<Types.ViewProps, {}> {
...
+    private _isButton(viewProps: Types.ViewProps): boolean {
         return !!(viewProps.onPress || viewProps.onLongPress);
     }
 
     render() {
         if (this.props.animateChildEnter || this.props.animateChildMove || this.props.animateChildLeave) {
             return (
                 <AnimateListEdits { ...this._internalProps }>
-                    { this.props.children }
+                    {this.props.children}
```
erictraut: Please add back the spaces. The coding convention we use throughout ReactXP includes spaces within curly bracket statements.

src/native-common/View.tsx
```diff
@@ -69,29 +69,61 @@ export class View extends ViewBase<Types.ViewProps, {}> {
...
                 </AnimateListEdits>
             );
         } else if (this._isButton(this.props)) {
             return (
                 <Button { ...this._internalProps }>
-                    { this.props.children }
+                    {this.props.children}
```
erictraut: Same here

src/native-common/View.tsx
```diff
@@ -69,29 +69,61 @@ export class View extends ViewBase<Types.ViewProps, {}> {
...
                 </Button>
             );
         } else {
             return (
                 <RN.View { ...this._internalProps }>
-                    { this.props.children }
+                    {this.props.children}
```
erictraut: Same here.

src/native-common/View.tsx
```diff
@@ -69,29 +69,61 @@ export class View extends ViewBase<Types.ViewProps, {}> {
...
+            this._processDragAndDropProps();
+        }
+    }
+
+    private _processDragAndDropProps() {
+        for (const name of ['onDragEnter', 'onDragOver', 'onDrop', 'onDragLeave']) {
+            const handler = this._internalProps[name];
+
+            if (handler) {
+                this._internalProps.allowDrop = true; // TODO: let RN figure this out?
```
erictraut: If this is just a reminder that we may want to revisit it in the future, don't make it a "TODO". If it's something that needs to be fixed, leave it as a TODO but add a reference to a bug report.

PR #378, opened 2017-12-08
Change TextPropsShared to extend CommonAccessibilityProps

Fitzpasd:
I noticed the `accessibilityLabel` property no longer exists on `Text`.  This PR fixes the issue by matching the pattern used by other component type defs, e.g. `ButtonProps`, by extending `CommonAccessibilityProps`.


Fitzpasd:
Nvm, this was a rushed fix that doesn't do what I thought it would.

PR #380, opened 2017-12-11
Build fix, react-native-windows version matched to react-native

dryganets:
…react-native


src/native-common/View.tsx
```diff
@@ -82,7 +82,7 @@ export class View extends ViewBase<Types.ViewProps, {}> {
             if (handler) {
                 this._internalProps.allowDrop = true;
 
-                this._internalProps[name] = (e: React.SyntheticEvent) => {
+                this._internalProps[name] = (e: React.SyntheticEvent<View>) => {
```
dryganets: SyntheticEvent is typed in public typings.
It doesn't compile without this change.

PR #384, opened 2017-12-14
Nested buttons are allowed in react-native.

dryganets:
View implementation for react-native generates a Button in case the View has the onPress property defined.  RR reverting this part of the change for now as it is hard to control.


erictraut:
I don't think we should disable this check — especially on the master branch. If you need to disable it temporarily for the Skype source base, you can do so on a private fork. Otherwise, we can wait for the bug fix that addresses the unfortunate conditional use of RX.Button within the RX.View implementation.

PR #386, opened 2017-12-15
Fix Text props and docs

Fitzpasd:
1. Removing the Android-only `elevation` prop from Text.  This isn't a valid `RN.Text` prop.
2. Fixing Text documentation to accurately reflect the props.

### DIFF
--- a/docs/docs/components/text.md
+++ b/docs/docs/components/text.md
@@ -16,25 +16,23 @@ Another difference between Text and other components is that Text children are n
 ## Props
 
 ``` javascript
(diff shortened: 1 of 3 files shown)


erictraut:
Thanks @Fitzpasd - all good changes.

reseul:
@erictraut @Fitzpasd  it's the styles (for RN.View and whatever uses them, including RN.Text) that seem to define that elevation.
Perhaps if someone cries we'll have to add support that way.

Issue #375, opened 2017-12-01
Native-common implementation of RX.View should't use RX.Button implementation


erictraut:
The current native-common implementation of RX.View (native-common/View.tsx) uses the native implementation of Button if the caller specifies an onPress or onLongPress handler. This approach is problematic for several reasons. First, it means that some of the props are not composable (e.g. you can't specify animateChildEnter=true along with an onPress handler). Second, it triggers the recently-added nested-button assertion. This assertion was added for good reasons because accessibility APIs don't work with nested buttons (and are also not permitted in HTML). 

We should probably eliminate the use of RX.Button in the RX.View implementation altogether.

erictraut:
Fixed

PR #391, opened 2017-12-18
Fix for #375: avoid using RX.Button when rendering an RX.View with onPress/onLongPress

reseul:
Also removed AnimateListEdits sine it didn't contain enough to warrant one more layer of component instances.


src/native-common/View.tsx
```diff
@@ -9,29 +9,202 @@
...
+                delay: 75,
+                duration: 150,
+                type: LayoutAnimation.Types.linear,
+                property: LayoutAnimation.Properties.opacity
+            };
+            var configDictionary: RN.ILayoutAnimationConfig = {
+                duration: 300,
+            };
+
+            if (this.props.animateChildMove) {
```
reseul: Hmm, something is wrong here, even this should be "nextprops", I think I need to update. Thinking.

erictraut:
@reseul, the changes look good to me. Do you have any additional testing or changes to do before I merge into master?

reseul:
@erictraut, let me do some due diligence and try the layout animation, that part is not used by our tests at all. I'll ping when done.
Also, what's the cherry-picking process to 0.46.x? I open a new PR from my github mirror, or it's done on msft one directly?

---
In reply to: [352487425](https://github.com/Microsoft/reactxp/pull/391#issuecomment-352487425) [](ancestors = 352487425)

erictraut:
This is a pretty big change to back-port to 0.46.x. I prefer to do changes of this size around major updates. If there's a strong need to back-port, let's talk offline.

The process for back-porting is to first merge into master, then submit a second PR on the branch.

reseul:
Ok, master it is.

---
In reply to: [352527001](https://github.com/Microsoft/reactxp/pull/391#issuecomment-352527001) [](ancestors = 352527001)

reseul:
@Erictrout, I think we're good to go. Thanks.

---
In reply to: [352526207](https://github.com/Microsoft/reactxp/pull/391#issuecomment-352526207) [](ancestors = 352526207,352487425)

PR #392, opened 2017-12-20
Removed onPanResponderTerminationRequest override

dryganets:
Removed onPanResponderTerminationRequest override, allowing it to fall back to undefined and use default behavior


erictraut:
Please rebase so you don't pull in all of the merged changes. This makes it difficult to review.

PR #395, opened 2017-12-20
Removed onPanResponderTerminationRequest override, allowing it to fal…

dryganets:
…l back to undefined and use default behavior


erictraut:
This fix isn't in master. All changes in other (older) branches should first be made in master, then back-ported where appropriate.

dryganets:
Now the fix is in the master.

PR #398, opened 2017-12-22
Adding option to check if any modal is being displayed.

dryganets:


Fitzpasd:
This should include some updates to the docs.

src/common/Interfaces.ts
```diff
@@ -103,6 +103,7 @@ export abstract class Modal {
     abstract show(modal: React.ReactElement<Types.ViewProps>, modalId: string, options?: Types.ModalOptions): void;
     abstract dismiss(modalId: string): void;
     abstract dismissAll(): void;
+    abstract anyDisplayed(): boolean;
```
erictraut: Keeping with the naming conventions in ReactXP, this should be called "isAnyDisplayed". I'm not sure this functionality requires a new method. We could allow users to specify an empty string to "isDisplayed" if they want to query whether any modal is displayed.

erictraut: Also, any new exposed functionality should update the docs and the RXPTest sample.

dryganets:
Updated the doc and API to what Eric proposed.
Now if modalId is undefined we check if we return true if any dialog is displayed.

erictraut:
This change needs to be documented in Modal.md and added to the RXPTest app. I can do it later this week when I'm back from vacation or you can do it in the meantime.

erictraut:
Never mind, I see that you added this to the documentation already. I'll add the RXPTest later.

PR #404, opened 2018-01-01
Keyboard focusing/navigation support for the UWP platform.

reseul:
Existing React Native for UWP uses Border/Canvas based implementation for RCTView, with no support for keyboard input of any kind.
An updated version (covered by a concurrent PR) contains a helper FocusableWindows component that maps to a UserControl/Canvas based native view, component that can be used to add keyboard input/focusing capability to other components.

Main Changes:
1. "windows" flavors of View, Button, Link, TextInput, and ScrollViewer. We keep "Text" as is, with a "focus" method used strictly for narrator purposes.
2. The existing Web FocusManager is now refactored into:
	- Common implementation (in common/utils)
	- Web and Desktop flavors through derived classes.
   Methods marked as "xxxx /* static */" are logically static, yet Typescript is not nice with us on that regard..
3. A native-desktop folder containing what would be common between uwp and the macos platforms: App, RootView, UserInterface, etc.
Some more refined refactoring may happen when macos platform is merged in.

### DIFF
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -63,7 +63,6 @@ export abstract class FocusManager {
     protected abstract /* static */ removeFocusListenerFromComponent(component: React.Component<any, any>, onFocus: () => void): void;
     protected abstract /* static */ focusComponent(component: React.Component<any, any>): boolean;
 
-    // protected abstract /* static */ focusFirst(last?: boolean): void;
     protected abstract /* static */ resetFocus() : void;
     protected abstract /* static */ _updateComponentFocusRestriction(storedComponent: StoredFocusableComponent): void;
 
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -14,10 +14,16 @@ import AppConfig from '../../common/AppConfig';
 
 let _lastComponentId: number = 0;
 
+export interface FocusableInternal {
+    focusableComponentId?: string;
+}
+
+export type FocusableComponentInternal = React.Component<any, any> & FocusableInternal;
+
 export interface StoredFocusableComponent {
     id: string;
     numericId: number;
-    component: React.Component<any, any>;
+    component: FocusableComponentInternal;
     onFocus: () => void;
     restricted: boolean;
     limitedCount: number;
@@ -63,9 +69,9 @@ export abstract class FocusManager {
         }
     }
 
-    protected abstract /* static */ addFocusListenerOnComponent(component: React.Component<any, any>, onFocus: () => void): void;
-    protected abstract /* static */ removeFocusListenerFromComponent(component: React.Component<any, any>, onFocus: () => void): void;
-    protected abstract /* static */ focusComponent(component: React.Component<any, any>): boolean;
+    protected abstract /* static */ addFocusListenerOnComponent(component: FocusableComponentInternal, onFocus: () => void): void;
+    protected abstract /* static */ removeFocusListenerFromComponent(component: FocusableComponentInternal, onFocus: () => void): void;
+    protected abstract /* static */ focusComponent(component: FocusableComponentInternal): boolean;
 
     protected abstract /* static */ resetFocus() : void;
     protected abstract /* static */ _updateComponentFocusRestriction(storedComponent: StoredFocusableComponent): void;
@@ -73,8 +79,8 @@ export abstract class FocusManager {
     // Whenever the focusable element is mounted, we let the application
     // know so that FocusManager could account for this element during the
     // focus restriction.
-    addFocusableComponent(component: React.Component<any, any>) {
-        if ((component as any)._focusableComponentId) {
+    addFocusableComponent(component: FocusableComponentInternal) {
+        if (component.focusableComponentId) {
             return;
         }
 
@@ -92,7 +98,7 @@ export abstract class FocusManager {
             }
         };
 
-        (component as any)._focusableComponentId = componentId;
+        component.focusableComponentId = componentId;
         FocusManager._allFocusableComponents[componentId] = storedComponent;
 
         let withinRestrictionOwner: boolean = false;
@@ -118,8 +124,11 @@ export abstract class FocusManager {
         this.addFocusListenerOnComponent(component, storedComponent.onFocus);
     }
 
-    removeFocusableComponent(component: React.Component<any, any>) {
-        const componentId: string = (component as any)._focusableComponentId;
+    removeFocusableComponent(component: FocusableComponentInternal) {
+        if (!component.focusableComponentId) {
+            return;
+        }
+        const componentId: string = component.focusableComponentId;
 
         if (componentId) {
             const storedComponent: StoredFocusableComponent = FocusManager._allFocusableComponents[componentId];
@@ -139,7 +148,7 @@ export abstract class FocusManager {
             }
 
             delete FocusManager._allFocusableComponents[componentId];
-            delete (component as any)._focusableComponentId;
+            delete component.focusableComponentId;
         }
     }
 
@@ -267,7 +276,7 @@ export abstract class FocusManager {
         this.removeFocusLimitation();
     }
 
-    subscribe(component: React.Component<any, any>, callback: FocusableComponentStateCallback) {
+    subscribe(component: FocusableComponentInternal, callback: FocusableComponentStateCallback) {
         const storedComponent = FocusManager._getStoredComponent(component);
 
         if (storedComponent) {
@@ -279,7 +288,7 @@ export abstract class FocusManager {
         }
     }
 
-    unsubscribe(component: React.Component<any, any>, callback: FocusableComponentStateCallback) {
+    unsubscribe(component: FocusableComponentInternal, callback: FocusableComponentStateCallback) {
         const storedComponent = FocusManager._getStoredComponent(component);
 
         if (storedComponent && storedComponent.callbacks) {
@@ -289,7 +298,7 @@ export abstract class FocusManager {
         }
     }
 
-    isComponentFocusRestrictedOrLimited(component: React.Component<any, any>): boolean {
+    isComponentFocusRestrictedOrLimited(component: FocusableComponentInternal): boolean {
         const storedComponent = FocusManager._getStoredComponent(component);
         return !!storedComponent && (storedComponent.restricted || storedComponent.limitedCount > 0);
     }
@@ -298,8 +307,8 @@ export abstract class FocusManager {
         return FocusManager._currentFocusedComponent ? FocusManager._currentFocusedComponent.id : undefined;
     }
 
-    private static _getStoredComponent(component: React.Component<any, any>): StoredFocusableComponent|undefined {
-        const componentId: string = (component as any)._focusableComponentId;
+    private static _getStoredComponent(component: FocusableComponentInternal): StoredFocusableComponent|undefined {
+        const componentId: string | undefined = component.focusableComponentId;
 
         if (componentId) {
             return FocusManager._allFocusableComponents[componentId];
@@ -346,23 +355,23 @@ export function applyFocusableComponentMixin(Component: any, isConditionallyFocu
     contextTypes.focusManager = PropTypes.object;
     Component.contextTypes = contextTypes;
 
-    inheritMethod('componentDidMount', function (this: React.Component<any, any>, focusManager: FocusManager) {
+    inheritMethod('componentDidMount', function (this: FocusableComponentInternal, focusManager: FocusManager) {
         if (!isConditionallyFocusable || isConditionallyFocusable.call(this)) {
             focusManager.addFocusableComponent(this);
         }
     });
 
-    inheritMethod('componentWillUnmount', function (this: React.Component<any, any>, focusManager: FocusManager) {
+    inheritMethod('componentWillUnmount', function (this: FocusableComponentInternal, focusManager: FocusManager) {
         focusManager.removeFocusableComponent(this);
     });
 
-    inheritMethod('componentWillUpdate', function (this: React.Component<any, any>, focusManager: FocusManager, origArguments: IArguments) {
+    inheritMethod('componentWillUpdate', function (this: FocusableComponentInternal, focusManager: FocusManager, origArgs: IArguments) {
         if (isConditionallyFocusable) {
-            let isFocusable = isConditionallyFocusable.apply(this, origArguments);
+            let isFocusable = isConditionallyFocusable.apply(this, origArgs);
 
-            if (isFocusable && !(this as any)._focusableComponentId) {
+            if (isFocusable && !this.focusableComponentId) {
                 focusManager.addFocusableComponent(this);
-            } else if (!isFocusable && (this as any)._focusableComponentId) {
+            } else if (!isFocusable && this.focusableComponentId) {
                 focusManager.removeFocusableComponent(this);
             }
         }
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -18,6 +18,7 @@ export interface StoredFocusableComponent {
     onFocus: () => void;
     restricted: boolean;
     limitedCount: number;
+    latestFocused?: boolean;
     origTabIndex?: number;
     origAriaHidden?: string;
     removed?: boolean;
@@ -81,7 +82,16 @@ export abstract class FocusManager {
             restricted: false,
             limitedCount: 0,
             onFocus: () => {
-                FocusManager._currentFocusedComponent = storedComponent;
+                if (FocusManager._currentFocusedComponent !== storedComponent) {
+                    if (FocusManager._currentFocusedComponent && FocusManager._currentFocusedComponent.latestFocused) {
+                        delete FocusManager._currentFocusedComponent.latestFocused;
+                        this._updateComponentFocusRestriction(FocusManager._currentFocusedComponent);
+                    }
+
+                    FocusManager._currentFocusedComponent = storedComponent;
+                    FocusManager._currentFocusedComponent.latestFocused = true;
+                    this._updateComponentFocusRestriction(FocusManager._currentFocusedComponent);
+                }
             }
         };
 
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -1,0 +1,379 @@
+/**
+* FocusManager.ts
+*
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT license.
+*
+* Manages focusable elements for better keyboard navigation.
+*/
+
+import React = require('react');
+import PropTypes = require('prop-types');
+
+let _lastComponentId: number = 0;
+
+export interface StoredFocusableComponent {
+    id: string;
+    component: React.Component<any, any>;
+    onFocus: () => void;
+    restricted: boolean;
+    limitedCount: number;
+    origTabIndex?: number;
+    origAriaHidden?: string;
+    removed?: boolean;
+    callbacks?: FocusableComponentStateCallback[];
+}
+
+export interface OriginalAttributeValues {
+    tabIndex: number|undefined;
+    ariaHidden: string|undefined;
+}
+
+export type FocusableComponentStateCallback = (restrictedOrLimited: boolean) => void;
+
+export abstract class FocusManager {
+    private static _rootFocusManager: FocusManager;
+
+    private static _restrictionStack: FocusManager[] = [];
+    private static _currentRestrictionOwner: FocusManager|undefined;
+    private static _restoreRestrictionTimer: number|undefined;
+    private static _pendingPrevFocusedComponent: StoredFocusableComponent|undefined;
+    private static _currentFocusedComponent: StoredFocusableComponent|undefined;
+    protected static _allFocusableComponents: { [id: string]: StoredFocusableComponent } = {};
+    protected static _skipFocusCheck = false;
+
+    private _parent: FocusManager|undefined;
+    private _isFocusLimited: boolean;
+    private _prevFocusedComponent: StoredFocusableComponent|undefined;
+    private _myFocusableComponentIds: { [id: string]: boolean } = {};
+
+    constructor(parent: FocusManager|undefined) {
+        if (parent) {
+            this._parent = parent;
+        } else if (FocusManager._rootFocusManager) {
+            console.error('FocusManager: root is already set');
+        } else {
+            FocusManager._rootFocusManager = this;
+        }
+    }
+
+    protected abstract /* static */ addFocusListenerOnComponent(component: React.Component<any, any>, onFocus: () => void): void;
+    protected abstract /* static */ removeFocusListenerFromComponent(component: React.Component<any, any>, onFocus: () => void): void;
+    protected abstract /* static */ focusComponent(component: React.Component<any, any>): boolean;
+
+    // protected abstract /* static */ focusFirst(last?: boolean): void;
+    protected abstract /* static */ resetFocus() : void;
+    protected abstract /* static */ _updateComponentFocusRestriction(storedComponent: StoredFocusableComponent): void;
+
+    // Whenever the focusable element is mounted, we let the application
+    // know so that FocusManager could account for this element during the
+    // focus restriction.
+    addFocusableComponent(component: React.Component<any, any>) {
+        if ((component as any)._focusableComponentId) {
+            return;
+        }
+
+        const componentId: string = 'fc-' + ++_lastComponentId;
+
+        let storedComponent: StoredFocusableComponent = {
+            id: componentId,
+            component: component,
+            restricted: false,
+            limitedCount: 0,
+            onFocus: () => {
+                FocusManager._currentFocusedComponent = storedComponent;
+            }
+        };
+
+        (component as any)._focusableComponentId = componentId;
+        FocusManager._allFocusableComponents[componentId] = storedComponent;
+
+        let withinRestrictionOwner: boolean = false;
+
+        for (let parent: FocusManager|undefined = this; parent; parent = parent._parent) {
+            parent._myFocusableComponentIds[componentId] = true;
+
+            if (FocusManager._currentRestrictionOwner === parent) {
+                withinRestrictionOwner = true;
+            }
+
+            if (parent._isFocusLimited) {
+                storedComponent.limitedCount++;
+            }
+        }
+
+        if (!withinRestrictionOwner && FocusManager._currentRestrictionOwner) {
+            storedComponent.restricted = true;
+        }
+
+        this._updateComponentFocusRestriction(storedComponent);
+
+        this.addFocusListenerOnComponent(component, storedComponent.onFocus);
+    }
+
+    removeFocusableComponent(component: React.Component<any, any>) {
+        const componentId: string = (component as any)._focusableComponentId;
+
+        if (componentId) {
+            const storedComponent: StoredFocusableComponent = FocusManager._allFocusableComponents[componentId];
+
+            this.removeFocusListenerFromComponent(component, storedComponent.onFocus);
+
+            storedComponent.removed = true;
+            storedComponent.restricted = false;
+            storedComponent.limitedCount = 0;
+
+            this._updateComponentFocusRestriction(storedComponent);
+
+            delete storedComponent.callbacks;
+
+            for (let parent: FocusManager|undefined = this; parent; parent = parent._parent) {
+                delete parent._myFocusableComponentIds[componentId];
+            }
+
+            delete FocusManager._allFocusableComponents[componentId];
+            delete (component as any)._focusableComponentId;
+        }
+    }
+
+    restrictFocusWithin(noFocusReset?: boolean) {
+        // Limit the focus received by the keyboard navigation to all
+        // the descendant focusable elements by setting tabIndex of all
+        // other elements to -1.
+        if (FocusManager._currentRestrictionOwner === this) {
+            return;
+        }
+
+        if (FocusManager._currentRestrictionOwner) {
+            this._removeFocusRestriction();
+        }
+
+        if (!this._prevFocusedComponent) {
+            this._prevFocusedComponent = FocusManager._pendingPrevFocusedComponent || FocusManager._currentFocusedComponent;
+        }
+
+        FocusManager._clearRestoreRestrictionTimeout();
+
+        FocusManager._restrictionStack.push(this);
+        FocusManager._currentRestrictionOwner = this;
+
+        Object.keys(FocusManager._allFocusableComponents).forEach(componentId => {
+            if (!(componentId in this._myFocusableComponentIds)) {
+                const storedComponent = FocusManager._allFocusableComponents[componentId];
+                storedComponent.restricted = true;
+                this._updateComponentFocusRestriction(storedComponent);
+            }
+        });
+
+        if (!noFocusReset) {
+            this.resetFocus();
+        }
+    }
+
+    removeFocusRestriction() {
+        // Restore the focus to the previous view with restrictFocusWithin or
+        // remove the restriction if there is no such view.
+        FocusManager._restrictionStack = FocusManager._restrictionStack.filter(focusManager => focusManager !== this);
+
+        if (FocusManager._currentRestrictionOwner === this) {
+            // We'll take care of setting the proper focus below,
+            // no need to do a regular check for focusout.
+            FocusManager._skipFocusCheck = true;
+
+            let prevFocusedComponent = this._prevFocusedComponent;
+            this._prevFocusedComponent = undefined;
+
+            this._removeFocusRestriction();
+            FocusManager._currentRestrictionOwner = undefined;
+
+            // Defer the previous restriction restoration to wait for the current view
+            // to be unmounted, or for the next restricted view to be mounted (like
+            // showing a modal after a popup).
+            FocusManager._clearRestoreRestrictionTimeout();
+            FocusManager._pendingPrevFocusedComponent = prevFocusedComponent;
+
+            FocusManager._restoreRestrictionTimer = setTimeout(() => {
+                FocusManager._restoreRestrictionTimer = undefined;
+                FocusManager._pendingPrevFocusedComponent = undefined;
+
+                const prevRestrictionOwner = FocusManager._restrictionStack.pop();
+
+                let needsFocusReset = true;
+
+                const currentFocusedComponent = FocusManager._currentFocusedComponent;
+                if (currentFocusedComponent && !currentFocusedComponent.removed &&
+                        !(currentFocusedComponent.id in this._myFocusableComponentIds)) {
+                    // The focus has been manually moved to something outside of the current
+                    // restriction scope, we should skip focusing the component which was
+                    // focused before the restriction and keep the focus as it is.
+                    prevFocusedComponent = undefined;
+                    needsFocusReset = false;
+                }
+
+                if (prevFocusedComponent && !prevFocusedComponent.removed &&
+                        !prevFocusedComponent.restricted && !prevFocusedComponent.limitedCount) {
+                    // If possible, focus the previously focused component.
+                    needsFocusReset = !this.focusComponent(prevFocusedComponent.component);
+                }
+
+                if (prevRestrictionOwner) {
+                    prevRestrictionOwner.restrictFocusWithin(true);
+                }
+
+                if (needsFocusReset) {
+                    this.resetFocus();
+                }
+            }, 100);
+        }
+    }
+
+    limitFocusWithin() {
+        if (this._isFocusLimited) {
+            return;
+        }
+
+        this._isFocusLimited = true;
+
+        Object.keys(this._myFocusableComponentIds).forEach(componentId => {
+            let storedComponent = FocusManager._allFocusableComponents[componentId];
+            storedComponent.limitedCount++;
+            this._updateComponentFocusRestriction(storedComponent);
+        });
+    }
+
+    removeFocusLimitation() {
+        if (!this._isFocusLimited) {
+            return;
+        }
+
+        Object.keys(this._myFocusableComponentIds).forEach(componentId => {
+            let storedComponent = FocusManager._allFocusableComponents[componentId];
+            storedComponent.limitedCount--;
+            this._updateComponentFocusRestriction(storedComponent);
+        });
+
+        this._isFocusLimited = false;
+    }
+
+    release() {
+        this.removeFocusRestriction();
+        this.removeFocusLimitation();
+    }
+
+    subscribe(component: React.Component<any, any>, callback: FocusableComponentStateCallback) {
+        const storedComponent = FocusManager._getStoredComponent(component);
+
+        if (storedComponent) {
+            if (!storedComponent.callbacks) {
+                storedComponent.callbacks = [];
+            }
+
+            storedComponent.callbacks.push(callback);
+        }
+    }
+
+    unsubscribe(component: React.Component<any, any>, callback: FocusableComponentStateCallback) {
+        const storedComponent = FocusManager._getStoredComponent(component);
+
+        if (storedComponent && storedComponent.callbacks) {
+            storedComponent.callbacks = storedComponent.callbacks.filter(cb => {
+                return cb !== callback;
+            });
+        }
+    }
+
+    isComponentFocusRestrictedOrLimited(component: React.Component<any, any>): boolean {
+        const storedComponent = FocusManager._getStoredComponent(component);
+        return !!storedComponent && (storedComponent.restricted || storedComponent.limitedCount > 0);
+    }
+
+    private static _getStoredComponent(component: React.Component<any, any>): StoredFocusableComponent|undefined {
+        const componentId: string = (component as any)._focusableComponentId;
+
+        if (componentId) {
+            return FocusManager._allFocusableComponents[componentId];
+        }
+
+        return undefined;
+    }
+
+    protected static _callFocusableComponentStateChangeCallbacks(storedComponent: StoredFocusableComponent, restrictedOrLimited: boolean) {
+        if (!storedComponent.callbacks) {
+            return;
+        }
+
+        storedComponent.callbacks.forEach(callback => {
+            callback.call(storedComponent.component, restrictedOrLimited);
+        });
+    }
+
+    private /* static */ _removeFocusRestriction() {
+        Object.keys(FocusManager._allFocusableComponents).forEach(componentId => {
+            let storedComponent = FocusManager._allFocusableComponents[componentId];
+            storedComponent.restricted = false;
+            this._updateComponentFocusRestriction(storedComponent);
+        });
+    }
+
+    private static _clearRestoreRestrictionTimeout() {
+        if (FocusManager._restoreRestrictionTimer) {
+            clearTimeout(FocusManager._restoreRestrictionTimer);
+            FocusManager._restoreRestrictionTimer = undefined;
+            FocusManager._pendingPrevFocusedComponent = undefined;
+        }
+    }
+}
+
+// A mixin for the focusable elements, to tell the views that
+// they exist and should be accounted during the focus restriction.
+//
+// isConditionallyFocusable is an optional callback which will be
+// called for componentDidMount() or for componentWillUpdate() to
+// determine if the component is actually focusable.
+export function applyFocusableComponentMixin(Component: any, isConditionallyFocusable?: Function) {
+    let contextTypes = Component.contextTypes || {};
+    contextTypes.focusManager = PropTypes.object;
+    Component.contextTypes = contextTypes;
+
+    inheritMethod('componentDidMount', function (this: React.Component<any, any>, focusManager: FocusManager) {
+        if (!isConditionallyFocusable || isConditionallyFocusable.call(this)) {
+            focusManager.addFocusableComponent(this);
+        }
+    });
+
+    inheritMethod('componentWillUnmount', function (this: React.Component<any, any>, focusManager: FocusManager) {
+        focusManager.removeFocusableComponent(this);
+    });
+
+    inheritMethod('componentWillUpdate', function (this: React.Component<any, any>, focusManager: FocusManager, origArguments: IArguments) {
+        if (isConditionallyFocusable) {
+            let isFocusable = isConditionallyFocusable.apply(this, origArguments);
+
+            if (isFocusable && !(this as any)._focusableComponentId) {
+                focusManager.addFocusableComponent(this);
+            } else if (!isFocusable && (this as any)._focusableComponentId) {
+                focusManager.removeFocusableComponent(this);
+            }
+        }
+    });
+
+    function inheritMethod(methodName: string, action: Function) {
+        let origCallback = Component.prototype[methodName];
+
+        Component.prototype[methodName] = function () {
+            let focusManager: FocusManager = this._focusManager || (this.context && this.context.focusManager);
+
+            if (focusManager) {
+                action.call(this, focusManager, arguments);
+            } else {
+                console.error('FocusableComponentMixin: context error!');
+            }
+
+            if (origCallback) {
+                origCallback.apply(this, arguments);
+            }
+        };
+    }
+}
+
+export default FocusManager;
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -14,11 +14,11 @@ let _lastComponentId: number = 0;
 
 export interface StoredFocusableComponent {
     id: string;
+    numericId: number;
     component: React.Component<any, any>;
     onFocus: () => void;
     restricted: boolean;
     limitedCount: number;
-    latestFocused?: boolean;
     origTabIndex?: number;
     origAriaHidden?: string;
     removed?: boolean;
@@ -39,9 +39,10 @@ export abstract class FocusManager {
     private static _currentRestrictionOwner: FocusManager|undefined;
     private static _restoreRestrictionTimer: number|undefined;
     private static _pendingPrevFocusedComponent: StoredFocusableComponent|undefined;
-    private static _currentFocusedComponent: StoredFocusableComponent|undefined;
+    protected static _currentFocusedComponent: StoredFocusableComponent|undefined;
     protected static _allFocusableComponents: { [id: string]: StoredFocusableComponent } = {};
     protected static _skipFocusCheck = false;
+    protected static _resetFocusTimer: number | undefined;
 
     private _parent: FocusManager|undefined;
     private _isFocusLimited: boolean;
@@ -74,24 +75,17 @@ export abstract class FocusManager {
             return;
         }
 
-        const componentId: string = 'fc-' + ++_lastComponentId;
+        const numericComponentId = ++_lastComponentId;
+        const componentId: string = 'fc-' + numericComponentId;
 
         let storedComponent: StoredFocusableComponent = {
             id: componentId,
+            numericId: numericComponentId,
             component: component,
             restricted: false,
             limitedCount: 0,
             onFocus: () => {
-                if (FocusManager._currentFocusedComponent !== storedComponent) {
-                    if (FocusManager._currentFocusedComponent && FocusManager._currentFocusedComponent.latestFocused) {
-                        delete FocusManager._currentFocusedComponent.latestFocused;
-                        this._updateComponentFocusRestriction(FocusManager._currentFocusedComponent);
-                    }
-
-                    FocusManager._currentFocusedComponent = storedComponent;
-                    FocusManager._currentFocusedComponent.latestFocused = true;
-                    this._updateComponentFocusRestriction(FocusManager._currentFocusedComponent);
-                }
+                FocusManager._currentFocusedComponent = storedComponent;
             }
         };
 
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -10,6 +10,8 @@
 import React = require('react');
 import PropTypes = require('prop-types');
 
+import AppConfig from '../../common/AppConfig';
+
 let _lastComponentId: number = 0;
 
 export interface StoredFocusableComponent {
@@ -53,7 +55,9 @@ export abstract class FocusManager {
         if (parent) {
             this._parent = parent;
         } else if (FocusManager._rootFocusManager) {
-            console.error('FocusManager: root is already set');
+            if (AppConfig.isDevelopmentMode()) {
+                console.error('FocusManager: root is already set');
+            }
         } else {
             FocusManager._rootFocusManager = this;
         }
@@ -373,7 +377,9 @@ export function applyFocusableComponentMixin(Component: any, isConditionallyFocu
             if (focusManager) {
                 action.call(this, focusManager, arguments);
             } else {
-                console.error('FocusableComponentMixin: context error!');
+                if (AppConfig.isDevelopmentMode()) {
+                    console.error('FocusableComponentMixin: context error!');
+                }
             }
 
             if (origCallback) {
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -297,6 +297,10 @@ export abstract class FocusManager {
         return !!storedComponent && (storedComponent.restricted || storedComponent.limitedCount > 0);
     }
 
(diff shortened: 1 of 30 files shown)


src/common/utils/FocusManager.ts
```diff
@@ -0,0 +1,387 @@
...
+        } else {
+            FocusManager._rootFocusManager = this;
+        }
+    }
+
+    protected abstract /* static */ addFocusListenerOnComponent(component: React.Component<any, any>, onFocus: () => void): void;
+    protected abstract /* static */ removeFocusListenerFromComponent(component: React.Component<any, any>, onFocus: () => void): void;
+    protected abstract /* static */ focusComponent(component: React.Component<any, any>): boolean;
+
+    // protected abstract /* static */ focusFirst(last?: boolean): void;
```
mshoho: Why is static commented and why is focusFirst commented?
If there is a reason, I guess we need to remove the commented code completely.

src/native-common/TextInput.tsx
```diff
@@ -48,51 +48,60 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
...
             />
         );
     }
 
+    render() {
+        const editable = (this.props.editable !== undefined ? this.props.editable : true);
+        const blurOnSubmit = this.props.blurOnSubmit || !this.props.multiline;
+
+        const internalProps: RN.TextInputProps = {
+            ref: 'nativeTextInput',
```
mshoho: Refs by string are deprecated. And it makes sense to upgrade the previously existing refs by string to callbacks in some files above too, since you're touching the files anyway.

src/common/utils/FocusManager.ts
reseul: Yeah, right, a leftover. Each flavor has its own focusFirst called from the corresponding resetFocus, etc.

---
In reply to: [159244518](https://github.com/Microsoft/reactxp/pull/404#discussion_r159244518) [](ancestors = 159244518)

src/native-common/TextInput.tsx
reseul: Ok. I knew this will come..

---
In reply to: [159245310](https://github.com/Microsoft/reactxp/pull/404#discussion_r159245310) [](ancestors = 159245310)

src/native-common/Button.tsx
```diff
@@ -85,7 +85,7 @@ export class Button extends React.Component<Types.ButtonProps, {}> {
 
     private _isMounted = false;
     private _hideTimeout: number|undefined;
-    private _buttonElement: RN.Animated.View|undefined;
+    private _buttonElement: RN.Animated.View|null = null;
```
rigdern: >null [](start = 45, length = 4)

Why do we switch this from `undefined` to `null`?

src/native-common/utils/EventHelpers.ts
```diff
@@ -0,0 +1,144 @@
...
+                        break;
+
+                    case 'Down':
+                    case 'ArrowDown':
+                        keyCode = 20;
+                        break;
+
+                    case 'Delete':
+                        keyCode = 8;
+                        break;
```
rigdern: Should we move this to be with "Backspace" above which has the same key code?

src/native-common/Button.tsx
```diff
@@ -117,27 +128,25 @@ export class Button extends React.Component<Types.ButtonProps, {}> {
...
-                shouldRasterizeIOS={ this.props.shouldRasterizeIOS }
-                onAccessibilityTapIOS={ this.props.onAccessibilityTapIOS }
-            >
-                { this.props.children }
-            </RN.Animated.View>
-        );
+        let internalProps: RN.ViewProps = {
+            ref: this._onButtonRef,
+            style: Styles.combine([_styles.defaultButton, this.props.style, opacityStyle,
+                            this.props.disabled && _styles.disabled]),
```
rigdern: >this.props.disabled && _styles.disabled]), [](start = 28, length = 42)

Is this line indented too much? Should be only be indented 4 spaces relative to the previous line?

src/windows/Button.tsx
```diff
@@ -0,0 +1,211 @@
...
+const UP_KEYCODES = [KEY_CODE_SPACE];
+
+let _isNavigatingWithKeyboard = false;
+
+UserInterface.keyboardNavigationEvent.subscribe(isNavigatingWithKeyboard => {
+   _isNavigatingWithKeyboard = isNavigatingWithKeyboard;
+});
+
+// Simple check for the presence of the updated React Native for Windows
+const IsUpdatedReactNativeForWindows = (RNW.FocusableWindows !== undefined);
```
rigdern: >IsUpdatedReactNativeForWindows [](start = 6, length = 30)

Consider calling this "HasFocusableWindows" or something similar. I think it's better to think in terms of the APIs the platform supports rather than thinking in terms of versions.

src/windows/Button.tsx
```diff
@@ -0,0 +1,211 @@
...
+        if (this._focusableElement) {
+            let tabIndex: number | undefined = this.getTabIndex();
+            let windowsTabFocusable: boolean = !this.props.disabled && tabIndex !== undefined && tabIndex >= 0;
+
+            this._focusableElement.setNativeProps({
+                tabIndex: tabIndex,
+                isTabStop: windowsTabFocusable
+            });
+        }
+    }
```
rigdern: Should we have a comment noting that all of these things are for implementing a particular interface?

src/windows/Button.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+    getTabIndex(): number | undefined {
+        // Button defaults to a tabIndex of 0
+        // Focus Manager may override this
+        return this.props.tabIndex || 0;
+    }
+
+    updateNativeTabIndex(): void {
+        if (this._focusableElement) {
+            let tabIndex: number | undefined = this.getTabIndex();
```
rigdern: >: number | undefined [](start = 24, length = 20)

Is it preferred to allow TypeScript to infer the type in cases like this?

src/windows/Button.tsx
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        // RNW.FocusableProps tabIndex: default is 0.
+        // -1 has no special semantic similar to DOM.
+        let tabIndex: number | undefined = this.getTabIndex();
+        // RNW.FocusableProps windowsTabFocusable:
+        // - true: keyboard focusable through any mean, receives keyboard input
+        // - false: not focusable at all, doesn't receive keyboard input
+        // The intermediate "focusable, but not in the tab order" case is not supported.
+        let windowsTabFocusable: boolean = !this.props.disabled && tabIndex !== undefined && tabIndex >= 0;
```
rigdern: >!this.props.disabled && tabIndex !== undefined && tabIndex >= 0 [](start = 43, length = 63)

Should we put this in a helper? I see we have this logic a couple of times.

src/windows/Button.tsx
```diff
@@ -0,0 +1,211 @@
...
+            if (this.props.onKeyPress) {
+                this.props.onKeyPress(keyEvent);
+            }
+
+            if (this.props.onPress) {
+                let key = keyEvent.keyCode;
+                // ENTER triggers press on key down
+                if (key === KEY_CODE_ENTER) {
+                    this.props.onPress(keyEvent);
+                    return;
```
rigdern: >return; [](start = 20, length = 7)

Why do we return here?

src/windows/Button.tsx
```diff
@@ -0,0 +1,211 @@
...
+            let keyEvent = EventHelpers.toKeyboardEvent(e);
+            if (this.props.onKeyPress) {
+                this.props.onKeyPress(keyEvent);
+            }
+
+            if (this.props.onPress) {
+                let key = keyEvent.keyCode;
+                // ENTER triggers press on key down
+                if (key === KEY_CODE_ENTER) {
+                    this.props.onPress(keyEvent);
```
rigdern: >keyEvent [](start = 39, length = 8)

Is this the type of event this handler expects?

src/windows/Button.tsx
```diff
@@ -0,0 +1,211 @@
...
+            }
+        }
+    }
+
+    private _onKeyUp = (e: React.SyntheticEvent<any>): void => {
+
+        let keyEvent = EventHelpers.toKeyboardEvent(e);
+        if (keyEvent.keyCode === KEY_CODE_SPACE) {
+            if (!this.props.disabled && this.props.onPress) {
+                this.props.onPress(keyEvent);
```
rigdern: >keyEvent [](start = 35, length = 8)

Is this the type of event this handler expects?

src/native-common/utils/EventHelpers.ts
reseul: Ok, will do

---
In reply to: [159335721](https://github.com/Microsoft/reactxp/pull/404#discussion_r159335721) [](ancestors = 159335721)

src/native-common/Button.tsx
reseul: ok. Was following the "keep close to the first param position" school of thought.

---
In reply to: [159335965](https://github.com/Microsoft/reactxp/pull/404#discussion_r159335965) [](ancestors = 159335965)

src/windows/Button.tsx
reseul: I don't know, I've seen either way. Can be argued that it's clearer (to the viewer) this way.

---
In reply to: [159337210](https://github.com/Microsoft/reactxp/pull/404#discussion_r159337210) [](ancestors = 159337210)

src/windows/Button.tsx
reseul: may become unclear. There are subtle differences... sometimes we have a "disabled", sometime not, etc.

---
In reply to: [159337425](https://github.com/Microsoft/reactxp/pull/404#discussion_r159337425) [](ancestors = 159337425)

src/windows/Button.tsx
reseul: ok

---
In reply to: [159336672](https://github.com/Microsoft/reactxp/pull/404#discussion_r159336672) [](ancestors = 159336672)

src/windows/Button.tsx
reseul: This handler expects less (SyntheticEvent), but sine we already have the beefed up event, we pass that one.

---
In reply to: [159338363](https://github.com/Microsoft/reactxp/pull/404#discussion_r159338363) [](ancestors = 159338363)

src/windows/Button.tsx
reseul: leftover from times we had more code. Removing.

---
In reply to: [159337858](https://github.com/Microsoft/reactxp/pull/404#discussion_r159337858) [](ancestors = 159337858)

src/windows/Button.tsx
reseul: IT expects less (see similar comment elseswhere)

---
In reply to: [159338483](https://github.com/Microsoft/reactxp/pull/404#discussion_r159338483) [](ancestors = 159338483)

src/windows/Button.tsx
reseul: Sure. What would be the customary comment decoration?

---
In reply to: [159337071](https://github.com/Microsoft/reactxp/pull/404#discussion_r159337071) [](ancestors = 159337071)

rigdern: I don't know if we have an established pattern. Something like `// From SomeThing interface`?

---
In reply to: [159342936](https://github.com/Microsoft/reactxp/pull/404#discussion_r159342936) [](ancestors = 159342936,159337071)

src/native-common/FrontLayerViewManager.tsx
```diff
@@ -134,6 +134,14 @@ export class FrontLayerViewManager {
         return null;
     }
 
+    public getActivePopupId() : string | null {
```
rigdern: >getActivePopupId [](start = 11, length = 16)

Why wasn't this needed in the web implementation?

src/native-common/Image.tsx
```diff
@@ -96,8 +106,21 @@ export class Image extends React.Component<Types.ImageProps, {}> {
...
+
     public setNativeProps(nativeProps: RN.ImageProperties) {
-        (this.refs['nativeImage'] as RN.Image).setNativeProps(nativeProps);
+        if (this._nativeRef) {
+            this._nativeRef.setNativeProps(nativeProps);
+        }
+    }
+
+    getChildContext() {
+        // Let descendant Types components know that their nearest Types ancestor is not an Types.Text.
```
rigdern: >Types [](start = 26, length = 5)

I think all occurrences of `Types` here should be `RX`

src/native-common/Text.tsx
```diff
@@ -47,6 +60,17 @@ export class Text extends React.Component<Types.TextProps, {}> {
         );
     }
 
+    private _onNativeTextRef = (ref: RN.Text|null) => {
+        this._nativeText = ref;
+    }
+
+    getChildContext() {
+        // Let descendant Types components know that their nearest Types ancestor is an Types.Text.
```
rigdern: >Types [](start = 26, length = 5)

I think all occurrences of `Types` here should be `RX`

src/native-common/FrontLayerViewManager.tsx
reseul: Web is structured differently. FrontLayerViewManager calls RenderDOM.render() by passing a RootView element with an onDismissPopup prop.
In native-common it's RootView that calls into FrontLayerViewManager 

---
In reply to: [159348484](https://github.com/Microsoft/reactxp/pull/404#discussion_r159348484) [](ancestors = 159348484)

src/native-common/Image.tsx
reseul: ok

---
In reply to: [159348619](https://github.com/Microsoft/reactxp/pull/404#discussion_r159348619) [](ancestors = 159348619)

src/native-common/Text.tsx
reseul: ok

---
In reply to: [159348943](https://github.com/Microsoft/reactxp/pull/404#discussion_r159348943) [](ancestors = 159348943)

src/native-common/TextInput.tsx
```diff
@@ -117,7 +132,7 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
         }
     }
 
-    private _onSelectionChange = (selEvent: React.SyntheticEvent<TextInput>) => {
+    private _onSelectionChange = (selEvent: React.SyntheticEvent<any>) => {
```
rigdern: >any [](start = 65, length = 3)

Why is this being changed?

src/native-desktop/RootView.tsx
```diff
@@ -0,0 +1,165 @@
...
+            // Provide the context with root FocusManager to all descendants.
+            return {
+                focusManager: this._focusManager
+            };
+        }
+
+        render() {
+            let content = super.render();
+
+            // Using "any" since onKeyDown/onKeyDown are not defined at RN.View property level
```
rigdern: > onKeyDown/onKeyDown [](start = 32, length = 20)

This says `onKeyDown` twice

src/native-common/TextInput.tsx
reseul: For consistency with other places, and because it hurt my eyes... This is a handler called by an RN.TextInput, so the correct type would have to be React.SyntheticEvent<RN.TextInput>, I think.

---
In reply to: [159349506](https://github.com/Microsoft/reactxp/pull/404#discussion_r159349506) [](ancestors = 159349506)

src/native-desktop/RootView.tsx
reseul: lol

---
In reply to: [159357269](https://github.com/Microsoft/reactxp/pull/404#discussion_r159357269) [](ancestors = 159357269)

src/native-desktop/RootView.tsx
```diff
@@ -0,0 +1,165 @@
...
+        _isNavigatingWithKeyboardUpateTimer: number | undefined;
+
+        constructor(...args: any[]) {
+            super(...args);
+            // Initialize the root FocusManager which is aware of all
+            // focusable elements.
+            this._focusManager = new FocusManager(undefined);
+        }
+
+        _onTouchStartCapture = (e: SyntheticEvent) => {
```
rigdern: >_onTouchStartCapture [](start = 8, length = 20)

Does this get called for mouse input? If not, who calls `_updateKeyboardNavigationState` for mouse input?

src/native-desktop/RootView.tsx
```diff
@@ -0,0 +1,165 @@
...
+            if (this._isNavigatingWithKeyboardUpateTimer) {
+                window.clearTimeout(this._isNavigatingWithKeyboardUpateTimer);
+                this._isNavigatingWithKeyboardUpateTimer = undefined;
+            }
+
+            if (this._isNavigatingWithKeyboard !== isNavigatingWithKeyboard) {
+                this._isNavigatingWithKeyboard = isNavigatingWithKeyboard;
+
+                UserInterface.keyboardNavigationEvent.fire(isNavigatingWithKeyboard);
+            }
```
rigdern: >} [](start = 12, length = 1)

How does the focus outline get rendered? It looks like the web version supports different focus outlines depending on whether the navigation was done with keyboard.

src/typings/react-native-windows.d.ts
```diff
@@ -0,0 +1,36 @@
...
+    import React = require('react');
+    import RN = require('react-native');
+
+    //
+    // Focusable view related declarations
+    // ----------------------------------------------------------------------
+    interface FocusableProps extends RN.ViewProps {
+        isTabStop?                      : boolean;
+        tabIndex?                       : number;
+        tabNavigation?                  : string; // enum( 'local', 'cycle', 'once' );
```
rigdern: >string; // enum( 'local', 'cycle', 'once' ); [](start = 42, length = 44)

I think you can use string types here:

`tabNavigation?: 'local' | 'cycle' | 'once'

src/typings/react-native-windows.d.ts
```diff
@@ -0,0 +1,36 @@
...
+    // ----------------------------------------------------------------------
+
+    import React = require('react');
+    import RN = require('react-native');
+
+    //
+    // Focusable view related declarations
+    // ----------------------------------------------------------------------
+    interface FocusableProps extends RN.ViewProps {
+        isTabStop?                      : boolean;
```
rigdern: >isTabStop?                      : boolean; [](start = 8, length = 42)

I think we prefer to avoid the column-like alignment because column-like alignment forces us to update the spacing of all the lines when we happen to add a new prop that has a name which is too long.

src/typings/react-native.d.ts
```diff
@@ -310,6 +317,11 @@ declare module 'react-native' {
         overScrollMode?: string; //enum( 'always', 'always-if-content-scrolls', 'never' )
         // iOS
         scrollIndicatorInsets?: {top: number, left: number, bottom: number, right: number };
+        // Windows only
+        onKeyDown? : Function;
+        onKeyUp? : Function;
+        tabNavigation? : string; // enum( 'local', 'cycle', 'once' );
```
rigdern: >string [](start = 25, length = 6)

I think you can use string types here:

tabNavigation?: 'local' | 'cycle' | 'once'

src/windows/Animated.tsx
```diff
@@ -0,0 +1,101 @@
...
+                { ...this.props }
+            >
+                { this.props.children }
+            </ReactAnimatedView>
+        );
+    }
+
+    focus() {
+        if (this._animatedComponent && this._animatedComponent._component) {
+            this._animatedComponent._component.focus();
```
rigdern: >_component [](start = 36, length = 10)

What is `_component`? Why don't we need this indirection when implementing `setNativeProps`?

src/windows/Link.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    private _onKeyDown = (e: React.SyntheticEvent<any>): void => {
+        if (this.props.onPress) {
+            let keyEvent = EventHelpers.toKeyboardEvent(e);
+            let key = keyEvent.keyCode;
+            // ENTER triggers press on key down
+            if (key === KEY_CODE_ENTER) {
+                // Defer to base class
+                this._onPress(keyEvent);
+                return;
```
rigdern: >return [](start = 16, length = 6)

is this return needed?

src/windows/ReactXP.ts
```diff
@@ -66,6 +66,7 @@ AccessibilityUtil.setAccessibilityPlatformUtil(AccessibilityPlatformUtil);
 // -- STRANGE THINGS GOING ON HERE --
 // See web/ReactXP.tsx for more details.
 
+// TODO change
```
rigdern: What is this TODO for?

src/windows/Link.tsx
```diff
@@ -0,0 +1,150 @@
...
+    render() {
+
+        // Fallback to native-common fast if the keyboard enabled component is not available
+        if (!HasFocusableWindows) {
+            return super.render();
+        }
+
+        let content: JSX.Element = super.render();
+
+        // The "in text parent" case requires a special nyi control.
```
rigdern: >// The "in text parent" case requires a special nyi control. [](start = 8, length = 60)

Is this something we plan on doing later? If so, I think we should create a VSTS task for it and prefix this comment with `TODO` and reference the VSTS task number.

Also, I think it would be clearer if you created the branch for this case and then put the comment in that otherwise empty branch.

src/windows/ScrollView.tsx
```diff
@@ -0,0 +1,51 @@
...
+                undefined;
+
+        // TODO: #737970 Remove special case for UWP when this bug is fixed. The bug
+        //   causes you to have to click twice instead of once on some pieces of UI in
+        //   order for the UI to acknowledge your interaction.
+        const keyboardShouldPersistTaps = 'always';
+
+        return (
+
+            <RN.ScrollView
```
rigdern: extra new line

src/windows/ScrollView.tsx
```diff
@@ -0,0 +1,51 @@
...
+import EventHelpers from '../native-common/utils/EventHelpers';
+
+export class ScrollView extends ScrollViewBase {
+
+    protected _render(props: RN.ScrollViewProps): JSX.Element {
+
+        var onKeyDownCallback = this.props.onKeyPress ?
+                // We have a callback function, call the wrapper
+                this._onKeyDown :
+                undefined;
```
rigdern: This looks indented a little too much. I think it should be indented 4 spaces relative to the previous line.

src/windows/TextInput.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    protected _render(props: RN.TextInputProps): JSX.Element {
+
+        return (
+            <RN.TextInput
+            {...props}
+            tabIndex = {this.getTabIndex()}
+            onFocus = { (e: React.FocusEvent<any>) => this._onFocusEx(e, props.onFocus)}
+        />)
+            ;
```
rigdern: I think the formatting is a little messed up. This stuff should be indented about 4 spaces more and the semicolon shouldn't be on a line by itselff.

src/windows/TextInput.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {TextInput as TextInputBase} from '../native-common/TextInput';
+
+export class TextInput extends TextInputBase implements FocusManagerFocusableComponent {
+
+    protected _render(props: RN.TextInputProps): JSX.Element {
+
+        return (
+            <RN.TextInput
+            {...props}
+            tabIndex = {this.getTabIndex()}
```
rigdern: > =  [](start = 20, length = 3)

no spaces around the `=`

src/windows/TextInput.tsx
```diff
@@ -0,0 +1,62 @@
...
+            onFocus = { (e: React.FocusEvent<any>) => this._onFocusEx(e, props.onFocus)}
```
rigdern: >onFocus [](start = 12, length = 7)

Are you stomping on the super class's `onFocus` handler?

src/windows/View.tsx
```diff
@@ -0,0 +1,371 @@
...
+import RNW = require('react-native-windows');
+import Types = require('../common/Types');
+import PropTypes = require('prop-types');
+
+import {View as ViewCommon} from '../native-common/View';
+import EventHelpers from '../native-common/utils/EventHelpers';
+import { applyFocusableComponentMixin, FocusManagerFocusableComponent, FocusManager } from '../native-desktop/utils/FocusManager';
+
+const KEY_CODE_ENTER = 13;
+const KEY_CODE_SPACE = 32;
```
rigdern: Should we define these somewhere common rather than repeating them across multiple files?

src/windows/View.tsx
```diff
@@ -0,0 +1,371 @@
...
+            }
+        }
+    }
+
+    componentWillReceiveProps(nextProps: Types.ViewProps) {
+        super.componentWillReceiveProps(nextProps);
+
+        if (!!this.props.restrictFocusWithin !== !!nextProps.restrictFocusWithin) {
+            console.error('View: restrictFocusWithin is readonly and changing it during the component life cycle has no effect');
+        } else if (!!this.props.limitFocusWithin !== !!nextProps.limitFocusWithin) {
```
rigdern: >limitFocusWithin [](start = 32, length = 16)

Are `restrictFocusWithin` and `limitFocusWithin` mutually exclusive? If not, we should use a second `if` rather than an `else if`

src/windows/View.tsx
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+
+    componentWillUnmount() {
+        super.componentWillUnmount();
+        if (this._focusManager) {
+            this._focusManager.release();
+        }
+        // Mixins may have added this
+        if (super.componentWillUnmount !== undefined) {
```
rigdern: >if (super.componentWillUnmount !== undefined) { [](start = 8, length = 47)

Can this be simplified to?:

`if (super.componentWillUnmount) {`

src/windows/View.tsx
```diff
@@ -0,0 +1,371 @@
...
+    protected _buildInternalProps(props: Types.ViewProps) {
+
+        // Base class does the bulk of _internalprops creation
+        super._buildInternalProps(props);
+
+        if (props.onKeyPress) {
+
+            // Define the handler for "onKeyDown" on first use, it's the safest way when functions
+            // called from super constructors are involved.
+            if (this._onKeyDown === undefined) {
```
rigdern: >if (this._onKeyDown === undefined) { [](start = 12, length = 36)

Can this be simplified to?:

if (this._oknKeyDown) {

src/windows/View.tsx
```diff
@@ -0,0 +1,371 @@
...
+        super.blur();
+        // Only forward to Button.
+        // The other cases are RN.View based elements with no meaningful focus support
+        if (this._focusableElement) {
+            this._focusableElement.blur();
+        }
+    }
+
+    getChildContext() {
+        // Let descendant Types components know that their nearest Types ancestor is not an Types.Text.
```
rigdern: >Types [](start = 26, length = 5)

`Types` should be `RX`

src/windows/View.tsx
```diff
@@ -0,0 +1,371 @@
...
+        if (this.props.onKeyPress) {
+            this.props.onKeyPress(keyEvent);
+        }
+
+        if (this.props.onPress) {
+            let key = keyEvent.keyCode;
+            // ENTER triggers press on key down
+            if (key === KEY_CODE_ENTER) {
+                this.props.onPress(keyEvent);
+                return;
```
rigdern: >return [](start = 16, length = 6)

extra return

src/native-desktop/RootView.tsx
reseul: Yes. It's a "mouse down", basically.
One thing that I know RNW doesn't have it fully is the direct "on mouse move" event. It does have the "touch move", but that's something different (more like "mouse dragging"). I'll open an issue when I have a chance (and fix).

---
In reply to: [159357430](https://github.com/Microsoft/reactxp/pull/404#discussion_r159357430) [](ancestors = 159357430)

src/native-desktop/RootView.tsx
reseul: We just defer to the UWP system focus rectangles (black rectangles) for now. Its displaying basically mimics the flag we have for "keyboard navigation on"", with some exceptions (the flag gets reset when pressing ESC, whereas I haven't found a way to convey this to UWP as well).
It's true web implementation offers 2 outlines. Also S4L disables that completely and uses an internal "focus outline" extension that shows a red rectangle. I chose not to cover that for now until it's clear why we have all these random initiatives and what we actually want.

---
In reply to: [159357484](https://github.com/Microsoft/reactxp/pull/404#discussion_r159357484) [](ancestors = 159357484)

src/windows/TextInput.tsx
reseul: I think I was trying to be fancy by not having assumptions on what the base class passed in the props. The arrow function though is created one each render now, kind of bad. I'll try to reshuffle.

---
In reply to: [159359055](https://github.com/Microsoft/reactxp/pull/404#discussion_r159359055) [](ancestors = 159359055)

src/windows/View.tsx
reseul: Yeah, maybe. I kind of hate the "if (x.method)" syntax, but if that's the customary pattern...

---
In reply to: [159359262](https://github.com/Microsoft/reactxp/pull/404#discussion_r159359262) [](ancestors = 159359262)

src/windows/View.tsx
reseul: Maybe, but not sure where.

---
In reply to: [159359132](https://github.com/Microsoft/reactxp/pull/404#discussion_r159359132) [](ancestors = 159359132)

src/typings/react-native-windows.d.ts
reseul: react-native.d.ts is wildly inconsistent from your p.o.v. 
I'll follow your advice, hope I won't be getting the opposite as well...

---
In reply to: [159357801](https://github.com/Microsoft/reactxp/pull/404#discussion_r159357801) [](ancestors = 159357801)

src/typings/react-native-windows.d.ts
reseul: Will do. This trick doesn't seem to be used elsewhere, unfortunately.

---
In reply to: [159357725](https://github.com/Microsoft/reactxp/pull/404#discussion_r159357725) [](ancestors = 159357725)

src/typings/react-native.d.ts
reseul: ok

---
In reply to: [159357972](https://github.com/Microsoft/reactxp/pull/404#discussion_r159357972) [](ancestors = 159357972)

src/windows/Animated.tsx
reseul: _animatedComponent is a helper component used by the Animated package (you can check in react-native/Libraries/Animated/src/AnimatedImplementation.js, createAnimatedComponent).
It renders whatever you passed in (an RN.View in this case), with a ref function setting _component. So _component is the RN.View instance in this case.
Now, animatedComponent has a straight forwarder of setNativeProps to _component.setNativeProps, so no special indirection is needed there.
The "focus" method is outside this story.


---
In reply to: [159358021](https://github.com/Microsoft/reactxp/pull/404#discussion_r159358021) [](ancestors = 159358021)

src/windows/Link.tsx
reseul: Yes, it's the Xaml Link we talked about.
I think the task exists, yet adding the # here would be not very relevant to outside Msft, that's why I avoided it.  

I don't get the branch comment.

---
In reply to: [159358659](https://github.com/Microsoft/reactxp/pull/404#discussion_r159358659) [](ancestors = 159358659)

src/windows/Link.tsx
reseul: nope

---
In reply to: [159358380](https://github.com/Microsoft/reactxp/pull/404#discussion_r159358380) [](ancestors = 159358380)

src/windows/ReactXP.ts
reseul: To... remove itself!

---
In reply to: [159358539](https://github.com/Microsoft/reactxp/pull/404#discussion_r159358539) [](ancestors = 159358539)

src/windows/ScrollView.tsx
reseul: Ok (wth are the linters good for??)

---
In reply to: [159358840](https://github.com/Microsoft/reactxp/pull/404#discussion_r159358840) [](ancestors = 159358840)

src/windows/ScrollView.tsx
reseul: ok

---
In reply to: [159358750](https://github.com/Microsoft/reactxp/pull/404#discussion_r159358750) [](ancestors = 159358750)

src/windows/TextInput.tsx
reseul: ok

---
In reply to: [159358977](https://github.com/Microsoft/reactxp/pull/404#discussion_r159358977) [](ancestors = 159358977)

src/windows/TextInput.tsx
reseul: ok

---
In reply to: [159358874](https://github.com/Microsoft/reactxp/pull/404#discussion_r159358874) [](ancestors = 159358874)

src/windows/View.tsx
reseul: True, copied from elsewhere. I'll change both places.

---
In reply to: [159359222](https://github.com/Microsoft/reactxp/pull/404#discussion_r159359222) [](ancestors = 159359222)

src/windows/View.tsx
reseul: ok

---
In reply to: [159359502](https://github.com/Microsoft/reactxp/pull/404#discussion_r159359502) [](ancestors = 159359502)

src/windows/View.tsx
reseul: ok

---
In reply to: [159359635](https://github.com/Microsoft/reactxp/pull/404#discussion_r159359635) [](ancestors = 159359635)

src/windows/Link.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    // From FocusManagerFocusableComponent interface
+    //
+    onFocus() {
+        // Focus Manager hook
+    }
+
+    getTabIndex(): number | undefined {
+        // Focus Manager may override this
+        return 0;
```
rigdern: Why don't we reference `this.props.tabIndex`? It looks like it's a part of `Types.LinkProps`.

src/windows/View.tsx
```diff
@@ -0,0 +1,367 @@
...
+
+            this._focusableElement.setNativeProps({
+                tabIndex: tabIndex,
+                isTabStop: windowsTabFocusable
+            });
+        }
+    }
+}
+
+// A value for tabIndex (even <0) marks a View as being potentially keyboard focusable
```
rigdern: Why is it focusable with a `tabIndex` of `-1`? That seems different than the web implementation.

What does "keyboard focusable" mean? That you can hit the tab key to focus the View?

src/native-desktop/UserInterface.tsx
```diff
@@ -0,0 +1,29 @@
...
+import {UserInterface as UserInterfaceCommon} from '../native-common/UserInterface';
+
+export class UserInterface extends UserInterfaceCommon {
+    private _isNavigatingWithKeyboard: boolean = false;
+    constructor() {
+        super();
+        this.keyboardNavigationEvent.subscribe(this._keyboardNavigationStateChanged);
+    }
+
+    isNavigatingWithKeyboard(): boolean {
```
rigdern: You put this under `native-desktop` but I wonder if this would be useful in non-desktop scenarios (e.g. an iPad with a physical keyboard). I believe Howard has been working on some iPad stuff. You might want to loop him in.

src/windows/Link.tsx
rigdern: What I meant is to write something like this:

```
if (this.context && !this.context.isRxParentAText) {
  /* a bunch of implementation details that you have */
} else if (this.context && this.context.isRxParentAText) {
  // TODO: The "in text parent" case requires a React Native view that maps to XAML Hyperlink but this RN view isn't implemented yet.
}
```

---
In reply to: [159486509](https://github.com/Microsoft/reactxp/pull/404#discussion_r159486509) [](ancestors = 159486509,159358659)

reseul:
Have to make a fix to align with PR feedback changes in RN for W, then I'll go through the latest comments.

src/native-desktop/utils/FocusManager.ts
```diff
@@ -0,0 +1,247 @@
...
+            // Override available, use this one
+            return this._tabIndexOverride;
+        } else {
+            // Override not available, defer to original handler to return the prop
+            return origCallback.call(this);
+        }
+    });
+
+    // Implement 'setTabIndexOverride'
+    Component.prototype['setTabIndexOverride'] = function (this: FocusManagerFocusableComponentInternal, tabIndex: number) {
```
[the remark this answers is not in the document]
erictraut: I agree with Adam's comment here.

src/typings/react-native.d.ts
```diff
@@ -267,6 +267,13 @@ declare module 'react-native' {
         // iOS
         onAccessibilityTapIOS?: Function;
         shouldRasterizeIOS? : boolean;
+
+        // Windows
+        onKeyDown?: Function;
```
erictraut: Yikes, I don't like the idea of adding these props to all components. These should be added selectively only to the components where they make sense. Some of these are even duplicates of existing props, for example the onHoverStart/End for RX.Button.

Any place that we're adding props for keyboard or mouse we should be asking ourselves why we don't also add support for these same methods on the web (and eventually MacOS). They should be equally applicable.

Also, any callbacks should be strongly typed, not generic "Function" types.

src/typings/react-native.d.ts
```diff
@@ -310,6 +317,11 @@ declare module 'react-native' {
         overScrollMode?: string; //enum( 'always', 'always-if-content-scrolls', 'never' )
         // iOS
         scrollIndicatorInsets?: {top: number, left: number, bottom: number, right: number };
+        // Windows only
+        onKeyDown? : Function;
+        onKeyUp? : Function;
+        tabNavigation? : 'local' | 'cycle' | 'once';
+        disableKeyboardBasedScrolling?: boolean;
```
erictraut: disableKeyboardBasedScrolling needs to be documented. Is it windows-specific? If so, it should also have a  comment.

src/typings/react-native.d.ts
```diff
@@ -310,6 +317,11 @@ declare module 'react-native' {
         overScrollMode?: string; //enum( 'always', 'always-if-content-scrolls', 'never' )
         // iOS
         scrollIndicatorInsets?: {top: number, left: number, bottom: number, right: number };
+        // Windows only
+        onKeyDown? : Function;
```
erictraut: Why do we need keydown/up methods in a scroll view? In the unlikely event that someone needs this, they can install the callbacks on a sub-view.

src/typings/react-native.d.ts
```diff
@@ -397,6 +409,8 @@ declare module 'react-native' {
         textBreakStrategy?: 'highQuality' | 'simple' | 'balanced';
         // macOS only property for submitting the text on enter
         submitTextOnEnter?: boolean;
+        // Windows only
+        tabIndex?: number;
```
erictraut: Is tabIndex really needed for ScrollView? If so, why is it not applicable to web as well?

src/windows/Link.tsx
```diff
@@ -0,0 +1,149 @@
...
+*
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT license.
+*
+* RN Desktop-specific implementation of the cross-platform Link abstraction.
+*/
+
+import React = require('react');
+import RNW = require('react-native-windows');
+import {applyFocusableComponentMixin, FocusManagerFocusableComponent} from '../native-desktop/utils/FocusManager';
```
erictraut: Curly braces should have spaces.

src/windows/Link.tsx
```diff
@@ -0,0 +1,149 @@
...
+import PropTypes = require('prop-types');
+
+import EventHelpers from '../native-common/utils/EventHelpers';
+import {Link as LinkCommon } from '../native-common/Link';
```
erictraut: See above.

src/windows/Link.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+const KEY_CODE_ENTER = 13;
+const KEY_CODE_SPACE = 32;
+
+const DOWN_KEYCODES = [KEY_CODE_SPACE, KEY_CODE_ENTER];
+const UP_KEYCODES = [KEY_CODE_SPACE];
+
+export interface LinkContext {
+    isRxParentAText?: boolean;
```
erictraut: See question above about isRxParentAText. I don't understand why this is needed in native.

src/windows/Link.tsx
```diff
@@ -0,0 +1,149 @@
...
+            let focusableViewProps: RNW.FocusableProps = {
+                ref: this._onFocusableRef,
+                isTabStop: windowsTabFocusable,
+                tabIndex: tabIndex,
+                disableSystemFocusVisuals: false,
+                handledKeyDownKeys: DOWN_KEYCODES,
+                handledKeyUpKeys: UP_KEYCODES,
+                onKeyDown: this._onKeyDown,
+                onKeyUp: this._onKeyUp,
+                onFocus: this._onFocus,
```
erictraut: Remove extraneous comma

src/windows/Link.tsx
```diff
@@ -0,0 +1,149 @@
...
+            };
+
+            content = (
+                <RNW.FocusableWindows
+                    {...focusableViewProps}
```
erictraut: Curly braces should have spaces.

src/windows/Link.tsx
```diff
@@ -0,0 +1,149 @@
...
+            );
+        }
+
+        return content;
+    }
+
+    focus() {
+        if (this._focusableElement &&
+            this._focusableElement.focus) {
+                this._focusableElement.focus();
```
erictraut: Please fix indenting for clarity.

src/windows/Link.tsx
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    blur() {
+        if (this._focusableElement &&
+            this._focusableElement.blur) {
+                this._focusableElement.blur();
```
erictraut: Please fix indenting for clarity.

src/windows/ScrollView.tsx
```diff
@@ -0,0 +1,50 @@
...
+            undefined;
+
+        // TODO: #737970 Remove special case for UWP when this bug is fixed. The bug
+        //   causes you to have to click twice instead of once on some pieces of UI in
+        //   order for the UI to acknowledge your interaction.
+        const keyboardShouldPersistTaps = 'always';
+
+        return (
+            <RN.ScrollView
+                {...props}
```
erictraut: Curly braces should have spaces.

src/windows/ScrollView.tsx
```diff
@@ -0,0 +1,50 @@
...
+                onKeyDown={ onKeyDownCallback }
+                keyboardShouldPersistTaps={ keyboardShouldPersistTaps }
+                tabNavigation={ this.props.tabNavigation }
+                disableKeyboardBasedScrolling={true}
```
erictraut: See above.

src/windows/TextInput.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+import {TextInput as TextInputBase} from '../native-common/TextInput';
+
+export class TextInput extends TextInputBase implements FocusManagerFocusableComponent {
+
+    protected _render(props: RN.TextInputProps): JSX.Element {
+
+        return (
+            <RN.TextInput
+            {...props}
```
erictraut: Please fix indentation and add spaces in curly braces.

src/windows/TextInput.tsx
```diff
@@ -0,0 +1,62 @@
...
+        // Focus Manager may override this
+        return this.props.tabIndex;
+    }
+
+    updateNativeTabIndex(): void {
+        if (this._textInputRef) {
+            let tabIndex: number | undefined = this.getTabIndex();
+            this._textInputRef.setNativeProps({
+                tabIndex: tabIndex,
+                value: this.state.inputValue, // mandatory for some reason
```
erictraut: Remove extraneous comma.

src/windows/View.tsx
```diff
@@ -0,0 +1,367 @@
...
+*
+* Windows-specific implementation of View.
+*/
+
+import React = require('react');
+import RNW = require('react-native-windows');
+import Types = require('../common/Types');
+import PropTypes = require('prop-types');
+
+import {View as ViewCommon} from '../native-common/View';
```
erictraut: Curly braces should have spaces.

src/windows/View.tsx
```diff
@@ -0,0 +1,367 @@
...
+                handledKeyUpKeys: UP_KEYCODES,
+                onKeyDown: this._onFocusableKeyDown,
+                onKeyUp: this._onFocusableKeyUp,
+                onFocus: this._onFocus,
+                onBlur: this._onBlur
+            };
+
+            content = (
+                <RNW.FocusableWindows
+                    {...focusableViewProps}
```
erictraut: Curly braces should have spaces.

src/native-desktop/utils/FocusManager.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    function inheritMethod(methodName: string, action: Function) {
+        let origCallback = Component.prototype[methodName];
+
+        if (origCallback) {
+            Component.prototype[methodName] = function () {
+                return action.call(this, origCallback, arguments);
+            };
+        } else {
+            console.error('FocusableComponentMixin: ' + methodName + ' error!');
```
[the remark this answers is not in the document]
reseul: The methods have to exist on components, they're mandatory. I can make the error message friendlier.

---
In reply to: [159582412](https://github.com/Microsoft/reactxp/pull/404#discussion_r159582412) [](ancestors = 159582412)

src/typings/react-native.d.ts
reseul: This one is part of the RN.TextInputProps, and yes, it's the Windows flavor of RN that supports it.
RX.TextInput maps to <textarea> or <input> in the web implementation. RX.TextInputProps does contain tabIndex, which is not conveyed to the HTML elements (it should!), yet the web flavor of FocusManager does set the tabIndex DOM property on those elements liberally.
We needed something similar in RN for W. The resultant of RX.TextInputProps.tabIndex & FocusManager overrides is pushed to RN.TextInput through this property.

---
In reply to: [159786215](https://github.com/Microsoft/reactxp/pull/404#discussion_r159786215) [](ancestors = 159786215)

src/typings/react-native.d.ts
reseul: RX.ScrollViewProps contains onKeyPress, (inherited from RX.ViewPropsShared, along with other props that don't really work in ScrollView...). onKeyPress is used at least in the VLV to implement app-side arrow based focus navigation.
The Web implementation passes onKeyPress to the backing <div> (just when using custom scrollbars, not sure why).
For Windows I decided to handle the onKeyDown bubbled React events from whatever "keyboard focusable" children of the scroll view.  

---
In reply to: [159786141](https://github.com/Microsoft/reactxp/pull/404#discussion_r159786141) [](ancestors = 159786141)

src/typings/react-native.d.ts
reseul: This is RN for Windows only, but we don't expose it over RX API.  I meant for the ""windows only" comment to apply to the whole paragraph, including disableKeyboardBasedScrolling. Where should I document more? Note this is react-native.d.ts, not Typings.ts


---
In reply to: [159786086](https://github.com/Microsoft/reactxp/pull/404#discussion_r159786086) [](ancestors = 159786086)

src/native-desktop/RootView.tsx
```diff
@@ -0,0 +1,165 @@
...
+// 3. A place for the root FocusManager
+type Constructor<T extends React.Component> = new (...args: any[]) => T;
+
+function applyDesktopBehaviorMixin<TRootViewBase extends Constructor<React.Component>>(RootViewBase: TRootViewBase) {
+    return class RootView extends RootViewBase implements React.ChildContextProvider<any> {
+        static childContextTypes: React.ValidationMap<any> = {
+            focusManager: PropTypes.object
+        };
+
+        _focusManager: FocusManager;
```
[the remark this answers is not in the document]
reseul: I really wanted to mark these accordingly, but any attempt to do that triggers an error later in the file (where we generate the new decorated classes) that looks like this "Property '_focusManager' of exported class expression may not be private or protected.".

The 0.42 flavor of this was simpler, I had just one class extensions. The current RXP versions brought multiple RootView flavors (with store/with props), so I tried the mixin approach to "extend" both base RootView classes with one set of code.


---
In reply to: [159783697](https://github.com/Microsoft/reactxp/pull/404#discussion_r159783697) [](ancestors = 159783697)

src/typings/react-native.d.ts
reseul: You mentioned RX.Button, so you may think this change is in Typings.ts. It's actually in react-native.d.ts.
onKeyDown handles bubbled events raised by children.
onMouse* are supported by a range of RN for W views. They map to RX.ViewPropShared.onMouse* properties, similar to what the web counterpart does with <div>.onMouse* events.

---
In reply to: [159786010](https://github.com/Microsoft/reactxp/pull/404#discussion_r159786010) [](ancestors = 159786010)

src/native-desktop/utils/FocusManager.ts
```diff
@@ -0,0 +1,247 @@
...
+    _onFocusSink?: () => void;
+}
+
+export class FocusManager extends FocusManagerBase {
+
+    constructor(parent: FocusManager | undefined) {
+        super(parent);
+    }
+
+    protected /* static */ addFocusListenerOnComponent(component: React.Component<any, any>, onFocus: () => void): void {
```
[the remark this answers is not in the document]
reseul: The signature is shared with the Web flavor of FocusManager, whereas FocusManagerFocusableComponentInternal  is specific to the "desktop" implementations only. 
Not sure how to proceed to make it look better and less nasty.

---
In reply to: [159784667](https://github.com/Microsoft/reactxp/pull/404#discussion_r159784667) [](ancestors = 159784667)

src/native-desktop/utils/FocusManager.ts
```diff
@@ -0,0 +1,247 @@
...
+
+        if (_isNavigatingWithKeyboard) {
+            // When we're in the keyboard navigation mode, we want to have the
+            // first focusable component to be focused straight away, without the
+            // necessity to press Tab.
+
+            // Defer the focusing to let the view finish its initialization and to allow for manual focus setting (if any)
+            // to be processed (the asynchronous nature of focus->onFocus path requires a delay)
+            FocusManager._resetFocusTimer = setTimeout(() => {
+                FocusManager._resetFocusTimer = undefined;
```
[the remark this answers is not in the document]
reseul: The handler touches just the FocusManager._currentFocusedComponent, which could be stale. The check below should also have a "|| storedComponent.removed". I'll fix here and wherever I see the pattern.
There won't be a crash without the fix, but the code may decide there's no need to call "focusFirst" when it actually is.

---
In reply to: [159784893](https://github.com/Microsoft/reactxp/pull/404#discussion_r159784893) [](ancestors = 159784893)

src/native-desktop/utils/FocusManager.ts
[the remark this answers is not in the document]
reseul: Yeah, I can do that.


---
In reply to: [159583189](https://github.com/Microsoft/reactxp/pull/404#discussion_r159583189) [](ancestors = 159583189)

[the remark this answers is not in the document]
reseul: Same answer.

---
In reply to: [159785346](https://github.com/Microsoft/reactxp/pull/404#discussion_r159785346) [](ancestors = 159785346)

src/windows/Button.tsx
```diff
@@ -0,0 +1,212 @@
...
+            onBlur: this._onBlur
+        };
+
+        return (
+            <RNW.FocusableWindows
+                {...focusableViewProps}
+            >
+                <RN.Animated.View
+                    {...internalProps}
+                    onMouseEnter={this._onMouseEnter}
```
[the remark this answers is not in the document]
reseul: RX.ButtonProps has onHoverStart/etc., there's no conflict. I kept RNW.FocusableWindows solely for the keyboard related stuff.


---
In reply to: [159584124](https://github.com/Microsoft/reactxp/pull/404#discussion_r159584124) [](ancestors = 159584124)

src/windows/Link.tsx
reseul: Aha, The 0.42 RXP didn't have it. 
The current one does. Weirdly enough this tabIndex was added to LinkProps instead of making the LinkProps derive from that CommonAccessibilityProps or so..

---
In reply to: [159584978](https://github.com/Microsoft/reactxp/pull/404#discussion_r159584978) [](ancestors = 159584978)

src/windows/View.tsx
reseul: "keyboard focusable" is meant to be used when differentiating from "mouse focusable".
Otherwise, we use "tab focusable" for being able to keyboard focus by tabbing.

Now, I decided to keep the few views with tabIndex==-1 under the FocusManager supervision in order to be able to implement the "focus the tabIndex==-1 component" trick. I'll take a new look, with all the changes you guys are asking for.

---
In reply to: [159586940](https://github.com/Microsoft/reactxp/pull/404#discussion_r159586940) [](ancestors = 159586940)

src/windows/Link.tsx
reseul: Basically it covers a case that *could* be covered by React Native for Windows (either by mitigating, or by implementing the right approach), but it currently doesn't.

Our RX.Link is using a plain RN.Text for iOs/Android, whereas for Windows we encapsulate one of those in a RNW.FocusableWindows control to give it keyboard focusing capability.
Problem is that currently inlining such a construct in a parent RN.Text (when embedding a link in a longer text) crashes RNW completely.

I chose to mitigate this for now by avoiding the RNW.FocusableWindows encapsulation when the parent is an RN.Text generating RX.Text, with the hope that a NYI XamlLink based view would come to help this scenario, or maybe all of them (there are some tradeoffs between accessibility and select-ability, so we're not sure of the solution yet).

But if this tracking is not allowed on RXP side, what I'm left with is removing RX.Link from the story for now, and we'll see what future reserves us.




---
In reply to: [159786422](https://github.com/Microsoft/reactxp/pull/404#discussion_r159786422) [](ancestors = 159786422)

src/native-desktop/RootView.tsx
```diff
@@ -0,0 +1,165 @@
...
+                // If Esc is pressed and the focused element stays the same after some time,
+                // switch the keyboard navigation off to dismiss the outline.
+                const activeComponent = FocusManager.getCurrentFocusedComponent();
+
+                if (this._isNavigatingWithKeyboardUpateTimer) {
+                    clearTimeout(this._isNavigatingWithKeyboardUpateTimer);
+                }
+
+                this._isNavigatingWithKeyboardUpateTimer = setTimeout(() => {
+                    this._isNavigatingWithKeyboardUpateTimer = undefined;
```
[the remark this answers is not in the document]
reseul: Not in this case, we don't dereference/call any component

---
In reply to: [159784381](https://github.com/Microsoft/reactxp/pull/404#discussion_r159784381) [](ancestors = 159784381)

src/typings/react-native.d.ts
erictraut: Ah, my bad. Yes, for some reason, I thought this was Typings.ts. Never mind.

erictraut:
@reseul, what's the current status of this PR? Do you have more changes you'd like to make before I merge it, or is it good to go?

reseul:
@erictraut, not ready yet. I have to fix that ugly casting + a little of the demixinification. 
I'll send a ping when done.

reseul:
@erictraut , ok, now this is ready in my opinion.
On full disclosure, I added these since fist PR was sent:
- changed the way we instantiate FocusableWindow: RNW has  "factory" createFocusableComponent(component) now.
- made the "calling focus on components with tabIndex -1" a little more accurate.
- I kept the IsRxTextParent hack in for now.

Now, you and @rigdern had a bunch of comments, and I *think* I address all of them, at least with some answers if not code changes, but I guess you should confirm.
As far as I'm concerned, this should go in! :)

PR #406, opened 2018-01-02
isDisplayed() method for Popup API

Fitzpasd:
Adding an `isDisplayed()` method to the Popup API, mimicking the same method from the Modal API.


erictraut:
Looks good. Thanks for providing documentation and test code as well!

Issue #409, opened 2018-01-04
RX.TextInput on mobile web keyboardType not applied


MovingGifts:
It seems that on web (checked on mobile safari), the `keyboardType` prop on `RX.TextInput` does not work correctly. For web, the correct HTML5 values for input type should carry over for a truly hybrid (ios/android/web) virtual keyboard approach.

These are the keyboards that need to be fixed for web to display the correct input type:
`keyboardType: 'default' | 'numeric' | 'email-address' | 'number-pad';`

default -> `<input type="text">`
numeric -> `<input type="tel">`
email-address -> `<input type="email">`
number-pad -> `<input type="tel">`

Above, I listed the correct HTML5 equivalent type to the right of each `keyboardType`. The numeric/number-pad should have tel as the type I think, check the [mobile input types](http://mobileinputtypes.com/) for why it's superior to the `<input type="number">`

Currently, all `keyboardTypes` render as `<input type="text"` in web/mobile web, which should be corrected for a better UX on mobile web.

Any thoughts on this?

erictraut:
This is similar to this issue: https://github.com/Microsoft/reactxp/issues/381. As I mentioned, we have put little or no energy into making mobile browsers work well because we assume that mobile users will want a native experience. If this is important to you, feel free to contribute a PR.

PR #410, opened 2018-01-05
Updated TextInput with _keyboardType() for HTML5

MovingGifts:
This fixes issue #409

### DIFF
--- a/src/web/TextInput.tsx
+++ b/src/web/TextInput.tsx
@@ -122,10 +122,27 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
                     onMouseUp={ this._checkSelectionChanged }
                     onPaste={ this._onPaste }
                     aria-label={ this.props.accessibilityLabel }
-                    type={ this.props.secureTextEntry ? 'password' : 'text' }
+                    type={ this.props.secureTextEntry ? 'password' : this._keyboardType() }
                 />
             );
         }
+    }
+
+    private _keyboardType = () => {
+        // Show the correct virtual keyboardType in HTML 5
+        var keyboardType = this.props.keyboardType || 'text';
+
+        if (keyboardType === 'default') {
+          keyboardType = 'text';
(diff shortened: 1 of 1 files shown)


src/web/TextInput.tsx
```diff
@@ -122,12 +122,29 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
...
                     onPaste={ this._onPaste }
                     aria-label={ this.props.accessibilityLabel }
-                    type={ this.props.secureTextEntry ? 'password' : 'text' }
+                    type={ this.props.secureTextEntry ? 'password' : this._keyboardType() }
                 />
             );
         }
     }
 
+    private _keyboardType = () => {
```
erictraut: Should be called _getKeyboardType. (All methods should generally begin with verb.)

src/web/TextInput.tsx
```diff
@@ -122,12 +122,29 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
...
+        // Show the correct virtual keyboardType in HTML 5
+        var keyboardType = this.props.keyboardType || 'text';
```
erictraut: Should use "let" rather than "var".

src/web/TextInput.tsx
```diff
@@ -122,12 +122,29 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
...
+
+        if (keyboardType === 'default') {
+          keyboardType = 'text';
```
erictraut: Should use four-space tabs for consistency.

src/web/TextInput.tsx
erictraut: I recommend not initializing keyboardType with this.props.keyboardType. It's safer to always initialize it to the default value 'text'. Then override the default if this.props.keyboardType is one of the known overrides. With the current implementation, if a new keyboardType is added to RN, it will cause this code to break.

PR #414, opened 2018-01-08
Update URLs tutorial

kant:
plus minor fixes (proposal)


erictraut:
Thanks, these are good changes.

Issue #12, opened 2017-04-08
Please but please consider Fable integration.


OnurGumus:
[Fable](http://fable.io/) is an excellent tool to convert F# to js. It also has a focus on react integration. Please consider adding fable bindings.

erictraut:
We don't currently have any plans to develop F# bindings, but we welcome contributions if you'd like to develop this.

PR #416, opened 2018-01-08
Typo

kant:
on #12


erictraut:
I don't see any change here. What is the typo you're fixing?

berickson1:
Mis-spelling of "example". GitHub inline diff highlighting shows it, while side-by-side does not

>  --- for exmaple,

PR #417, opened 2018-01-08
Updated _getKeyboardType() for Search on mobileweb

MovingGifts:
Wasn't sure how to use rebase correctly, here is a second attempt with the latest code base pulled 👍

### DIFF
--- a/src/web/TextInput.tsx
+++ b/src/web/TextInput.tsx
@@ -101,7 +101,9 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
                 />
             );
         } else {
-            return (
+            let { keyboardTypeValue, wrapInForm } = this._getKeyboardType();
+            
+            let input = (
                 <input
                     style={ combinedStyles as any }
                     value={ this.props.defaultValue ? undefined : this.state.inputValue }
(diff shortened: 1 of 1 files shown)


MovingGifts:
@erictraut I know this was merged but when I do an `npm i -S reaxtxp` it installs `reactxp@0.46.6` which does not seem to have it. Shouldn't `npm i -S reaxtxp` install the latest version with the merged commit? Not sure what I am doing wrong, any idea how to get the correct npm version installed with the reflected changes?

erictraut:
This change was merged into the master branch, not the 0.46.x branch. We generally don't add new functionality like this to a production branch (like 0.46.x) because it introduces too much risk.

We've since published several versions of the 0.51.x branch which contains this change. You can change your package.json to pick up reactxp ^0.51.0. This will install the latest minor version of the stable 0.51.x branch.

MovingGifts:
Got it now, thank you very much!

MovingGifts:
So it seems a few things need to be fixed for the search form to work correctly:

- Style added to expand to full width `style={{ display: 'flex', flex: 1 }}`, not sure if you prefer this to be a defined style in the very top instead of inline.
- Once you hit "Search" on the virtual keyboard, the keyboard remains open. The correct way to hide it is to cause the input field to blur onSubmit of the form. I put `alert('hide');` in the correct location where the enclosed input needs to be blurred, but I am not sure how to target it to blur it. Thoughts?

```jsx
<form action=''
      onSubmit={ (ev) => { /* prevent form submission/page reload */ ev.preventDefault(); alert('hide'); } }
      style={{ display: 'flex', flex: 1 }}>
      { input }
</form>
```

erictraut:
Hmm, how did this work before? Did you test your previous PR? Please test all code thoroughly before submitting it.

It should be a predefined style on top because it will prevent unnecessary renders.

There is already a blur method on the TextInput component. You can just call this.blur() within the onSubmit handler.

MovingGifts:
@erictraut Sorry, I updated the comment with an addition as well regarding numbers. I will implement your suggestions, sorry about the style flex part not sure how that slipped.

MovingGifts:
Suggestions implemented locally and `this.blur()` works great :) Awaiting your guidance on the `pattern="\d*"` slash `\` stripping.

erictraut:
_getKeyboardType should return a pattern rather than having that logic in multiple places. You can add another field to the object returned by _getKeyboardType. Call it "pattern".

The reason the backslash is getting stripped is that it's an escape character in javascript strings. You need to add a double backslash to represent a single backslash within a string literal - i.e. '\\d*'.

MovingGifts:
Thanks for the guidance, implemented changes in [this pull request](https://github.com/Microsoft/reactxp/pull/472). Tested locally works really well.

PR #422, opened 2018-01-11
Fix View animation regression

reseul:
Copied more animation/underlay color code from RX.Button to RX.View.


reseul:
@erictraut Are we waiting on more reviews?

PR #428, opened 2018-01-16
Fix some linter reds, no intended code changes.

reseul:


erictraut:
Thanks Radu! I'm not sure why tslint didn't generate errors when I wrote this code. Must have it configured incorrectly.

reseul:
There's two more issue types left from what I see:
- Empty blocks in the handlers. Tried with a space, didn't silence it. The one that works is {/* */}, but yuck.
- Lines too long.
So those are left as is for now.

erictraut:
I just fixed the remaining issues.

PR #454, opened 2018-01-18
Exposing aria-roledescription to workaround some VoiceOver issues.

mshoho:


src/common/Types.ts
```diff
@@ -590,6 +591,7 @@ export interface ViewPropsShared extends CommonProps, CommonAccessibilityProps {
     importantForLayout?: boolean; // Web-only, additional invisible DOM elements will be added to track the size changes faster
     id?: string; // Web-only. Needed for accessibility.
     ariaLabelledBy?: string; // Web-only. Needed for accessibility.
+    ariaRoleDescription?: string; // Web-only. Needed for accessibility.
```
erictraut: Since you're adding a new prop, this needs to be added to the documentation (view.md). 

It also needs to be added to a test case (e.g. ViewBasicTest.tsx in RXPTest).

erictraut: Is there no way we can derive this prop from other (cross-platform) accessibility props like accessibilityTraits? I hate to add more web-specific props to ReactXP.

mshoho: I hate this as well and the plan is to have a hopefully better accessibility API for the whole RX in the upcoming weeks.

mshoho: Done.

PR #456, opened 2018-01-19
[A11Y] Fix announcement issue on Android

george-cz:
On Android, the voice over fails to announce text if it's the same as last one. The way it's constructed, that's actually expected, because the props of RN.View don't change. Work around this by keeping the information about last announcement and append space if they are the same.


erictraut:
Change looks good. I'll merge it into master. Please submit a second PR merging it into the 0.51.x branch, which is what Skype is currently using.

PR #457, opened 2018-01-19
Prevent web page reload when hitting enter in RX.TextInput

isnotgood:
In RX.TextInput pressing enter in `<input>` submits enclosing `<form>` element which leads in most cases to page reload. 

It happens when component is in uncontrolled mode (no handlers are installed on input to prevent default behavior).

Introduced by #417


erictraut:
Change looks good. I'll merge it into the master branch. You'll probably want to submit a second PR to integrate it into the 0.51.x branch, which is what Skype is currently using.

isnotgood:
Thank you Eric. Pushed backport for 0.51.x.

PR #466, opened 2018-01-24
MacOS implementation of Button and Animated

alregner:
+ compilation fix.


src/macos/Button.tsx
```diff
@@ -0,0 +1,72 @@
+/**
+* Button.tsx
+*
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT license.
+*
+* Mac-specific implementation of the cross-platform Button abstraction.
+*/
+import {Button as ButtonBase, ButtonContext} from '../native-common/Button';
```
erictraut: Please add spaces before/after the curly braces for consistency.

src/macos/Button.tsx
```diff
@@ -0,0 +1,72 @@
...
+import Types = require('../common/Types');
+import React = require('react');
+import RN = require('react-native');
+
+export class Button extends ButtonBase {
```
erictraut: Can't we just put the mouse support in ButtonBase? It should be ignored on systems that don't support mice.

src/web/AccessibilityUtil.ts
```diff
@@ -39,27 +39,27 @@ const roleMap: { [key: string]: string } = {
     [Types.AccessibilityTrait.Status]: 'status',
     [Types.AccessibilityTrait.Dialog]: 'dialog',
     [Types.AccessibilityTrait.Switch]: 'switch'
-}; 
+};
 
 // Map of accesssibility live region to an aria-live property.
-const liveRegionMap: { [key: string]: string } = {
+const liveRegionMap: { [key: string]: 'off' | 'assertive' | 'polite'; } = {
```
erictraut: Nit: unnecessary semicolon.

erictraut: Since you use this union type in multiple places, it would be good to define it once and use the type name rather than another (duplicate) inline definition.

src/macos/Animated.tsx
```diff
@@ -0,0 +1,81 @@
+/**
+* Animated.tsx
+*
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT license.
+*
+* RN-specific implementation of the cross-platform Animation abstraction.
+*/
+import React = require('react');
```
erictraut: I don't like the fact that we're duplicating this code for every platform. We should look for a better way to share this code. If you have an idea for how to do this, feel free. Otherwise, it can wait until some future change.

src/macos/Button.tsx
alregner: I will move this part to native-common/Button, revert Mac implementation and adjust windows implementation.

src/macos/Animated.tsx
alregner: Agree, I think constant ReactNativeAnimatedClasses is the only part that needs to be modified for platform specific needs, Animated wrappers look the same to me (I will need to check windows implementation first).

erictraut:
Nice! I like this approach. This is a pretty big change, so we should make sure to do some thorough testing. Have you had a chance to run the RXPTest app on iOS and Android to make sure nothing is broken? If not, I can do so next week. (I'm on vacation this week and don't have a Mac with me.)

alregner:
I ran RXPTest app on iOS only - nothing seems to be broken. I will verify that Android works tomorrow or the day after (I just came down with flu).

alregner:
I tested on Android today and it works. BTW. There is an issue when using Run All on android (Nesting of <View> within <Text>), that is present on current master as well, i.e. this change has nothing to do with it. Interestingly, it does not happen when you click each test and run one by one. I will try to find the cause tomorrow and create a new PR for that.

PR #476, opened 2018-01-30
Prevent crash when animated component is not mounted in DOM

isnotgood:


isnotgood:
@msfterictraut Could you pls take a look at this? We have crash in VLV when some of the VirtualListCells are not mounted. Not sure if this is exactly the right fix so feel free to trash it.

erictraut:
Hmm. I don't understand how we could get into this situation. A component's setValue method is called only if it has subscribed as a listener to a particular animated value. This subscription is registered only when the component is mounted, and it is unregistered when the component is unmounted. So I don't see a way for it to get called if the component is unmounted. Are you able to repro the problem, or are you guessing that it's the problem based on crash logs?

erictraut:
@isnotgood, any further thoughts on this one?

isnotgood:
@erictraut Sorry Eric for late response, I was OOF. As you said - I'm really guessing from crash logs. I can't repro issue locally but I can still see it the wild (I'll ping you from ticket so you can check out).

erictraut:
@isnotgood Is it possible that the crashes were occurring in a version that used ReactXP 0.46.x? I rewrote the web animation module in 0.51.x. Just want to make sure we're not chasing an old bug that no longer exists.

isnotgood:
@erictraut checked the revision and package.json contains "reactxp": "0.51.1"

erictraut:
I think I've found the underlying bug. I've fixed it on master and the 0.51.x branch.

isnotgood:
@erictraut Thank you Eric.

PR #478, opened 2018-01-31
Fix master to work with latest react typings (accessibility html attributes added)

asdanilenk:


src/web/AccessibilityUtil.ts
```diff
@@ -42,15 +42,15 @@ const roleMap: { [key: string]: string } = {
 }; 
 
 // Map of accesssibility live region to an aria-live property.
-const liveRegionMap: { [key: string]: string } = {
+const liveRegionMap: { [key: string]: 'off' | 'assertive' | 'polite' } = {
```
erictraut: Did you consider defining a named type for this so we don't have to duplicate the literal type definition multiple times? I worry that they'll get out of sync.

asdanilenk: Out of sync should not happen as aria-live is a standardized attribute with only these 3 values. I'm surprised that @types did not introduce a type for this.

Done

PR #482, opened 2018-02-01
 Added valuenow attribute for slider role support.

msftagoldbe:


erictraut:
@msftagoldbe, looks like there's a merge conflict that needs to be resolved.

PR #483, opened 2018-02-01
 Added valuenow attribute for slider role support.

msftagoldbe:


src/common/Types.ts
```diff
@@ -368,6 +368,7 @@ export interface CommonAccessibilityProps {
 
     // Desktop only.
     tabIndex?: number;
+    valueNow?: number;
```
erictraut: New props should be documented. In this case, please add it to the 'accessibility.md' documentation. (I alphabetized the props in the docs, BTW.) There's no need to modify the documentation on the 0.51.x branch because only the master branch docs are published.

erictraut: New props should also be exercised (or at least used) in the RXPTest app (in the samples directory). This helps prevent future regressions.

src/typings/react-native.d.ts
```diff
@@ -240,6 +240,9 @@ declare module 'react-native' {
         accessible?                      : boolean;
         onAcccessibilityTap?             : Function;
 
+        // desktop
+        valueNow?                        : number;
```
mshoho: Is it somehow used in Native?
I see a property being added for View for Web, but no action for Native.
If it's Web-only let's have it ariaValueNow for the consistency with other attributes. And I'll be easier to clean up later, when the API improvements will come.

src/common/Types.ts
msftagoldbe: Updated doc with valueNow prop info.

msftagoldbe: Added test to AccessibilityTests to exercise valueNow prop.

src/typings/react-native.d.ts
msftagoldbe: Updated to ariaValueNow.

src/typings/react-native.d.ts
```diff
@@ -240,6 +240,9 @@ declare module 'react-native' {
         accessible?                      : boolean;
         onAcccessibilityTap?             : Function;
 
+        // desktop
+        ariaValueNow?                        : number;
```
mshoho: Actually, this one is not needed in this file at all, please have a look at ariaRoleDescription or ariaLabelledBy. They only present in ViewPropsShared in Types.ts and in the web implementation of the View.

msftagoldbe: Removed

PR #486, opened 2018-02-02
Jofuria/fix params

jofuria:


berickson1:
Approved for master in #471 - backporting

PR #488, opened 2018-02-07
Fix random Android crashes when Talkback is enabled

isnotgood:


isnotgood:
Btw performance was taken in account before I choose to go with try-catch block (as some engines are supposedly not able to optimize such functions when they get hot). This function is called sparingly  anyway so I don't expect disaster :) I didn't put whole thing into try-catch as I don't want to swallow exceptions from RN.NativeModules.UIManager.sendAccessibilityEvent().

PR #492, opened 2018-02-08
Fix measureInWindow() callback signature

ladipro:
Unlike measure(), measureInWindow() is supposed to return only x, y,
width, height.

https://facebook.github.io/react-native/docs/direct-manipulation.html#measureinwindowcallback


ladipro:
This was found when playing with react-native-windows:
https://github.com/Microsoft/react-native-windows/pull/1625

PR #496, opened 2018-02-12
Add missing properties to SyntheticEvent type

Fitzpasd:


src/common/Types.ts
```diff
@@ -1121,8 +1123,12 @@ export type SyntheticEvent = {
     readonly cancelable: boolean;
     readonly defaultPrevented: boolean;
     readonly timeStamp: number;
-    readonly currentTarget: EventTarget;
+    readonly currentTarget: EventTarget | NodeHandle;
```
erictraut: I intentionally removed these extra fields from SyntheticEvent because they were not present on all platforms. Do you see a need for them to be present?

Fitzpasd: I thought there were.  They are in the type definitions for [react-native](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L344) and [react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L436), and would need to be here to eventually move to `@types/react-native`.

erictraut: I'd prefer to have a ReactXP-specific definition of SyntheticEvent rather than rely on either (or both) of the react and react-native public definition files. The ReactXP version should be a proper subset of all supported platforms.

Fitzpasd: Makes sense.  This change was motivated by my desire to get a dependency on `@types/react-native` rather than using an internal typing file.  This would then be pulled down by consumers of RXP automatically and they could use it whenever diving into RN specific functionality in their app.

Right now, this project has no good solution for this scenario IMO.  I'd like to get valid RN typings for use in a RXP project automatically either by:
1. RXP having a dependency on `@types/react-native`
2. RXP shipping an internal `react-native.d.ts` typing file that gets picked up automatically by consumers.

Alternatively, let people pull down `@types/react-native` themselves, but I'd hope this is something we could solve for them (like is done currently with both `@types/react` and `@types/react-dom`.

Fitzpasd:
Sounds like we've had similar experiences.  I've added a few definitions to the public file after trying to use it in ReactXP.

Feel free to close this PR since this isn't something we want to align on yet.  For my own understanding, which platforms don't support these event props?

I do think that long term we need a solution for people wanting a set of valid RN typings when using this library (without tracking one down themselves).  I'll continue to send small PRs as I have time to move towards the goal of including `@types/react-native`.  If that eventually proves impossible, we can hopefully explore adding a custom definition in the source that gets picked up by consumers.

PR #498, opened 2018-02-12
Remove parameter from onBlur event

Fitzpasd:
This doesn't exist in the latest @types/react-native definitions.


pronebird:
@Fitzpasd could you please clarify why you removed `FocusEvent`?

PR #500, opened 2018-02-13
Pass 'selection' prop to TextInput only if explicitly set

mrousal:
Original code is causing glitches for UWP because of the loop from onSelectionChange back to render (props update). If it is needed for any other platform I can create separate impl for Windows.

### DIFF
--- a/src/native-common/TextInput.tsx
+++ b/src/native-common/TextInput.tsx
@@ -27,9 +27,14 @@ export interface TextInputState {
     isFocused: boolean;
 }
 
+interface Selection {
+    start: number;
+    end: number;
+}
+
 export class TextInput extends React.Component<Types.TextInputProps, TextInputState> {
-    private _selectionStart: number = 0;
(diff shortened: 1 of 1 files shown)


src/native-common/TextInput.tsx
```diff
@@ -192,27 +198,20 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
...
     }
 
     getSelectionRange(): { start: number, end: number } {
-        let range = {
-            start: this._selectionStart,
-            end: this._selectionEnd
-        };
-
-        return range;
+        return this._selection;
```
erictraut: I don't see any place where this._selection is updated. It appears that it will always return { start: 0, end: 0 } regardless of the current selection.

mrousal: It is updated in _onSelectionChange, line 161

erictraut: Ah yes, I missed that. Looks good.

PR #504, opened 2018-02-14
Pass selection in render only after explicitly set (#500)

mrousal:
backport from master to 0.51.x

### DIFF
--- a/src/native-common/TextInput.tsx
+++ b/src/native-common/TextInput.tsx
@@ -27,9 +27,14 @@ export interface TextInputState {
     isFocused: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


erictraut:
@mrousal, do we really need this change in 0.51.x? We generally try to avoid risky changes on the "stable branch" that impact multiple platforms. If you can wait for another week or so, we will be publishing the 1.0.x branch, which already contains this change. If this is critical, we can backport it, but I will need you to do extensive testing on other platforms to make sure that it doesn't introduce a regression.

mrousal:
In that case it is not needed I think.

PR #514, opened 2018-02-21
Use white-space:pre for the aria-live region

ladipro:
Announcing the same text repeatedly with announceForAccessibility() is done
through triggering DOM updates by appending (and removing) a space character
to the text. This does not work reliably if the region is not marked as
preformatted.

For example Chrome trims leading and trailing whitespace from the text
at this callstack if not marked as preformatted:

blink_core!blink::TextIteratorTextNodeHandler::HandleTextNodeInRange+0x5f7
blink_core!blink::TextIteratorAlgorithm<blink::EditingAlgorithm<blink::NodeTraversal> >::HandleTextNode+0x270
blink_core!blink::TextIteratorAlgorithm<blink::EditingAlgorithm<blink::NodeTraversal> >::Advance+0x865
blink_core!blink::TextIteratorAlgorithm<blink::EditingAlgorithm<blink::NodeTraversal> >::TextIteratorAlgorithm+0x6e3
blink_core!blink::CreatePlainText<blink::EditingAlgorithm<blink::NodeTraversal> >+0xf5
blink_core!blink::PlainText+0x25
blink_core!blink::LayoutText::PlainText+0x96
blink_modules!blink::AXLayoutObject::TextAlternative+0x1c9
blink_modules!blink::AXObject::GetName+0x119
blink_modules!blink::WebAXObject::GetName+0xc7
content!content::BlinkAXTreeSource::SerializeNode+0x41b

The DOM update then does not result in a change to the text and no accessibility
event is fired by the browser.

### DIFF
--- a/src/web/RootView.tsx
+++ b/src/web/RootView.tsx
@@ -147,8 +147,8 @@ export class RootView extends React.Component<RootViewProps, RootViewState> {
                 } else {
 
                     // Additionally, alternate between announcement text directly under the aria-live element and
-                    // nested in a div to work around issues with some browsers. Chrome on Windows is known to
-                    // not fire accessibility events reliably without this, for example.
+                    // nested in a div to work around issues with some readers. NVDA on Windows is known to
+                    // not announce aria-live reliably without this, for example.
                     this.setState({
                         announcementText: announcement,
                         announcementTextInNestedDiv: !this.state.announcementTextInNestedDiv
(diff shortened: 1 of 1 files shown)


ladipro:
This fixes announceForAccessibility() in S4L on Mac web (Chrome and Safari). S4L Mac Electron worked without it for some reason; the way the DOM is modified there throws off the text comparison somehow.

Windows, both web (Chrome) and Electron still need the other workaround of injecting a nested div, although we are one step closer to eliminating it. Each announceForAccessibility() invocation now results in NotifyWinEvent() and NVDA speaks the text. The remaining problem is that it speaks too much :)

ladipro:
And the 'speaks too much' problem appears to be specific to NVDA indeed, I don't see it with JAWS.

Issue #515, opened 2018-02-21
If Button is given accessibilityTraits=MenuItem under Windows this leads to not passing Button trait to implementation


zholobov:
If Button is given accessibilityTraits=MenuItem under Windows this leads to not passing Button trait to implementation, and so it it does not render as button from accessibility point of view.

zholobov:
fixed

PR #516, opened 2018-02-22
fix for #515 "If Button is given accessibilityTraits=MenuItem under Windows this leads to not passing Button trait to implementation"

zholobov:


zholobov:
@erictraut @reseul @rigdern @gmangov can you please take a look?

erictraut:
@zholobov, looks good to me.

PR #517, opened 2018-02-22
Ability to set limitFocusWithin without automatically setting aria-hidden=true.

mshoho:
Needed for the proper VoiceOver navigation in the lists.

### DIFF
--- a/docs/docs/components/view.md
+++ b/docs/docs/components/view.md
@@ -62,7 +62,7 @@ ignorePointerEvents: boolean = false; // web only
 // setFocusRestricted() below.
 // WARNING: For the sake of performance, this property is readonly and
 // changing it during the View life cycle will produce an error.
-restrictFocusWithin: boolean = false; // web only
+restrictFocusWithin?: boolean = false; // web only
 
 // When the keyboard navigation is happening, do not focus on this view
 // and on all focusable elements inside this view. See also the companion
@@ -71,9 +71,13 @@ restrictFocusWithin: boolean = false; // web only
 // one list item (and item's internal focusable elements) after another
 // using the Tab key and implement the switching between the items using
 // the arrow keys (or using some other behaviour).
+// When limitFocusWithin=LimitFocusType.Limited, the View and the focusable
+// components inside the View get both tabIndex=-1 and aria-hidden=true.
+// When limitFocusWithin=LimitFocusType.Accessible, the View and the focusable
+// components inside the View get only tabIndex=-1.
 // WARNING: For the sake of performance, this property is readonly and
 // changing it during the View life cycle will produce an error.
-limitFocusWithin: boolean = false; // web only
+limitFocusWithin?: LimitFocusType = LimitFocusType.Unlimited; // web only
 
 // Additional invisible DOM elements will be added inside the view
 // to track the size changes that are performed behind our back by
--- a/docs/docs/components/view.md
+++ b/docs/docs/components/view.md
@@ -62,7 +62,7 @@ ignorePointerEvents: boolean = false; // web only
 // setFocusRestricted() below.
 // WARNING: For the sake of performance, this property is readonly and
 // changing it during the View life cycle will produce an error.
-restrictFocusWithin?: boolean = false; // web only
+restrictFocusWithin: boolean = false;
 
 // When the keyboard navigation is happening, do not focus on this view
 // and on all focusable elements inside this view. See also the companion
@@ -77,7 +77,7 @@ restrictFocusWithin?: boolean = false; // web only
 // components inside the View get only tabIndex=-1.
 // WARNING: For the sake of performance, this property is readonly and
 // changing it during the View life cycle will produce an error.
-limitFocusWithin?: LimitFocusType = LimitFocusType.Unlimited; // web only
+limitFocusWithin: LimitFocusType = LimitFocusType.Unlimited;
 
 // Additional invisible DOM elements will be added inside the view
 // to track the size changes that are performed behind our back by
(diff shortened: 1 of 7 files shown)


src/common/utils/FocusManager.ts
```diff
@@ -110,8 +110,14 @@ export abstract class FocusManager {
                 withinRestrictionOwner = true;
             }
 
-            if (parent._isFocusLimited) {
-                storedComponent.limitedCount++;
+            const isFocusLimited = parent._isFocusLimited;
```
erictraut: _isFocusLimited is no longer a boolean, so this should be:
 = (parent._isFocusLimited !== Types.LimitFocusType.Unlimited)

src/common/utils/FocusManager.ts
```diff
@@ -226,8 +233,11 @@ export abstract class FocusManager {
                     needsFocusReset = false;
                 }
 
-                if (prevFocusedComponent && !prevFocusedComponent.removed &&
-                        !prevFocusedComponent.restricted && !prevFocusedComponent.limitedCount) {
+                if (prevFocusedComponent &&
+                    !prevFocusedComponent.removed &&
+                    !prevFocusedComponent.restricted &&
+                    !prevFocusedComponent.limitedCount &&
```
erictraut: limitedCount and limitedCountAccessible aren't booleans, so the proper check is "> 0".

src/common/utils/FocusManager.ts
```diff
@@ -264,11 +280,17 @@ export abstract class FocusManager {
 
         Object.keys(this._myFocusableComponentIds).forEach(componentId => {
             let storedComponent = FocusManager._allFocusableComponents[componentId];
-            storedComponent.limitedCount--;
+
+            if (this._isFocusLimited === Types.LimitFocusType.Accessible) {
+                storedComponent.limitedCountAccessible--;
+            } else {
```
erictraut: This code is a bit fragile in that it will break if new elements are added to the LimitFocusType enumeration. I recommend adding an explicit check "else if (this._isFocusLimited === Types.LimitFocusType.Limited)".

docs/docs/components/view.md
```diff
@@ -73,7 +73,7 @@ restrictFocusWithin: boolean = false; // web only
 // the arrow keys (or using some other behaviour).
 // WARNING: For the sake of performance, this property is readonly and
 // changing it during the View life cycle will produce an error.
-limitFocusWithin: boolean = false; // web only
+limitFocusWithin?: LimitFocusType = undefined; // web only
```
erictraut: Please add documentation for the LimitFocusType so it's clear what the three different enumeration states mean.

src/web/utils/FocusManager.ts
```diff
@@ -108,7 +106,11 @@ export class FocusManager extends FocusManagerBase {
     static focusFirst(last?: boolean) {
         const focusable = Object.keys(FocusManager._allFocusableComponents)
             .map(componentId => FocusManager._allFocusableComponents[componentId])
-            .filter(storedComponent => !storedComponent.removed && !storedComponent.restricted && !storedComponent.limitedCount)
+            .filter(storedComponent =>
+                !storedComponent.removed &&
+                !storedComponent.restricted &&
+                !storedComponent.limitedCount &&
```
erictraut: limitedCount and limitedCountAccessible are not booleans, so the proper check here is "> 0" - for readability.

mshoho:
I have tightened the value checks.

docs/docs/components/view.md
```diff
@@ -62,7 +62,7 @@ ignorePointerEvents: boolean = false; // web only
 // setFocusRestricted() below.
 // WARNING: For the sake of performance, this property is readonly and
 // changing it during the View life cycle will produce an error.
-restrictFocusWithin: boolean = false; // web only
+restrictFocusWithin?: boolean = false; // web only
```
erictraut: We don't use "?" within the documentation. The fact that it has an "= false" on the end means that it's optional and that "false" is the default value.

docs/docs/components/view.md
```diff
@@ -71,9 +71,13 @@ restrictFocusWithin: boolean = false; // web only
...
 // using the Tab key and implement the switching between the items using
 // the arrow keys (or using some other behaviour).
+// When limitFocusWithin=LimitFocusType.Limited, the View and the focusable
+// components inside the View get both tabIndex=-1 and aria-hidden=true.
+// When limitFocusWithin=LimitFocusType.Accessible, the View and the focusable
+// components inside the View get only tabIndex=-1.
 // WARNING: For the sake of performance, this property is readonly and
 // changing it during the View life cycle will produce an error.
-limitFocusWithin: boolean = false; // web only
+limitFocusWithin?: LimitFocusType = LimitFocusType.Unlimited; // web only
```
erictraut: See note above.

docs/docs/components/view.md
erictraut: Is this really "web only"? Or does it apply to all keyboard-based systems? If it's the latter, delete the "// web only" comment.

docs/docs/components/view.md
erictraut: Is this really "web only"? Or does it apply to all keyboard-based systems? If it's the latter, delete the "// web only" comment.

samples/RXPTest/src/Tests/ViewBasicTest.tsx
```diff
@@ -201,7 +201,7 @@ class BasicView extends RX.Component<RX.CommonProps, RX.Stateless> {
 // blockPointerEvents: boolean = false; // iOS and Android only
 // ignorePointerEvents: boolean = false; // web only
 // restrictFocusWithin: boolean = false; // web only
-// limitFocusWithin: boolean = false; // web only
+// limitFocusWithin: LimitFocusType = LimitFocusType.Unlimited; // web only
```
erictraut: Is this really "web only"? Or does it apply to all keyboard-based systems? If it's the latter, delete the "// web only" comment.

PR #518, opened 2018-02-22
Ability to set limitFocusWithin without automatically setting aria-hidden=true.

mshoho:
Needed for the proper VoiceOver navigation in the lists.

### DIFF
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -567,6 +567,11 @@ export interface AnimatedTextProps extends TextPropsShared {
 
 export type ViewLayerType = 'none' | 'software' | 'hardware';
 
+// When limitFocusWithin=true, the focusable components inside the View get
+// both tabIndex=-1 and aria-hidden=true, when limitFocusWithin='accessible' the
+// components only get tabIndex=-1
+export type LimitFocusType = false | true | 'accessible';
+
 // View
 export interface ViewPropsShared extends CommonProps, CommonAccessibilityProps {
     title?: string;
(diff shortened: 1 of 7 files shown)


src/common/Types.ts
```diff
@@ -567,6 +567,11 @@ export interface AnimatedTextProps extends TextPropsShared {
 
 export type ViewLayerType = 'none' | 'software' | 'hardware';
 
+// When limitFocusWithin=true, the focusable components inside the View get
+// both tabIndex=-1 and aria-hidden=true, when limitFocusWithin='accessible' the
+// components only get tabIndex=-1
+export type LimitFocusType = false | true | 'accessible';
```
erictraut: We shouldn't define types that are a combination of strings and boolean values. Besides being really gross, this won't work for OverReact.

mshoho: It's used the similar way in React (for example, HTMLAttributes interface definition in React, it uses «boolean | 'some' | 'other' | 'value'», or I see things like «size?: number | "small" | "large"» in React Native).

This type definition intends to keep the backwards compatibility. Why wouldn't it work?

erictraut: Yes, I know that this pattern has crept into React in a few places. I don't want to allow it in reactxp because I think it's a bad practice to combine two distinct types like booleans and strings or numbers and strings. 

It won't work with OR because these types can't be represented in C++.

Let's look at other approaches here.

mshoho: What if I declare it as enum for master:
```
export enum LimitFocusType {
    UNLIMITED = 0,
    LIMITED = 1,
    ACCESSIBLE = 2
}
```

But keep like this for the backport for the sake of backward compatibility?

erictraut: This is a fine approach for master. I don't think it should be back-ported to the stable branch. We're going to be moving Skype to the new branch soon (probably within a week — two at most).

mshoho: Changed it in the master PR: https://github.com/Microsoft/reactxp/pull/517

PR #521, opened 2018-02-23
Avoid triggering sync layout on web

ladipro:
Displaying a popup triggers synchronous layout (aka "recalculate style", "forced reflow") at multiple places. This is a collection of tweaks to avoid some of these and make popups faster to appear.

Time spent in layout went down from ~220 ms to ~150 ms when showing the emoticon picker in S4L Electron. This translated to the popup showing onClick handler taking ~80 ms (or ~20%) less time overall.

The remaining offender, not addressed in this PR, is RootView._recalcPosition() which heavily depends on measuring various elements. I'm open to suggestions on how tackle this one. As noted in the commit messages, my web-fu is weak :)

### DIFF
--- a/src/web/CustomScrollbar.ts
+++ b/src/web/CustomScrollbar.ts
@@ -1,5 +1,5 @@
-﻿/**
-* Scrollbar.ts
+/**
+* CustomScrollbar.ts
 *
 * Copyright (c) Microsoft Corporation. All rights reserved.
 * Licensed under the MIT license.
@@ -14,7 +14,6 @@ import React = require('react');
 var UNIT = 'px';
 var SCROLLER_MIN_SIZE = 15;
 var SCROLLER_NEGATIVE_MARGIN = 30;
-var LTR_OVERRIDE_CLASS = 'ltroverride';
 var NEUTRAL_OVERRIDE_CLASS = 'neutraloverride';
 
 interface IScrollbarInfo {
@@ -173,6 +172,7 @@ export class Scrollbar {
     private _handleWheelCallback = this._handleWheel.bind(this);
     private _handleMouseDownCallback = this._handleMouseDown.bind(this);
     private _updateCallback = this.update.bind(this);
+    private _asyncInitTimer: number|undefined;
 
     static getNativeScrollbarWidth() {
         // Have we cached the value alread?
@@ -253,12 +253,8 @@ export class Scrollbar {
             isNeutral = isLeftBound && isRightBound;
 
         this._container.classList.remove(NEUTRAL_OVERRIDE_CLASS);
-        this._container.classList.remove(LTR_OVERRIDE_CLASS);
-
         if (isNeutral) {
             this._container.classList.add(NEUTRAL_OVERRIDE_CLASS);
-        } else if (isLeftBound) {
-            this._container.classList.add(LTR_OVERRIDE_CLASS);
         }
 
         rtlbox.innerHTML = '';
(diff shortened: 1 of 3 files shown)


src/web/CustomScrollbar.ts
```diff
@@ -511,14 +507,22 @@ export class Scrollbar {
...
         this._addScrollbars();
         this.show();
-        this.update();
         this._container.addEventListener('mouseenter', this._updateCallback);
+
+        // Defer remaining init work to avoid triggering sync layout
+        this._asyncInitTimer = window.setTimeout(() => {
+            this._tryLtrOverride();
+            this.update();
+        }, 0);
```
erictraut: You should set this._asyncInitTimer to undefined within the callback. Otherwise clearInterval will be called on an already-fired timer in dispose.

src/web/ViewBase.tsx
```diff
@@ -113,32 +113,32 @@ export abstract class ViewBase<P extends Types.ViewProps, S> extends RX.ViewBase
             return SyncTasks.Resolved<void>();
         }
 
-        const container = this._getContainer();
-        if (!container) {
-            return SyncTasks.Resolved<void>();
-        }
+        const deferred = SyncTasks.Defer<void>();
+        ViewBase._reportLayoutChange(() => {
```
erictraut: I worry about the perf implications of this change. The previous code tried very hard to avoid creating timers in the frequent case where the view size and position didn't change. The new code always allocates a timer on every update. Have you done sufficient perf measurements on parts of the code that will be impacted by this (e.g. the VMV in Skype)?

erictraut:
I don't see a way to work around _recalcPosition. We need that information before we can display the popup in its intended position. Deferring the work will not help the perceived performance and will likely hurt it.

I'm concerned about the negative perf impact of this change — in particular, the change in ViewBase. This change will cause a timer to be initialized for every update of a view that has an onLayout handler specified. The previous code tried very hard to avoid this if it wasn't necessary (i.e. the view didn't change size or position).

src/web/utils/FocusManager.ts
```diff
@@ -157,10 +159,13 @@ export class FocusManager extends FocusManagerBase {
...
             // focusable, focusing it, removing the focus and making it unfocusable
             // back again.
-            const prevTabIndex = FocusManager._setTabIndex(document.body, 0);
-            document.body.focus();
-            document.body.blur();
-            FocusManager._setTabIndex(document.body, prevTabIndex);
+            // Defer the work to avoid triggering sync layout.
+            FocusManager._resetFocusTimer = setTimeout(() => {
+                const prevTabIndex = FocusManager._setTabIndex(document.body, 0);
+                document.body.focus();
```
erictraut: Please have Marat review this to make sure it's not introducing an undesirable race condition. For example, what if the user presses the tab key to move the focus to the next item? This key event used to arrive after the synchronous code ran, but now it can arrive before the deferred code runs.

ladipro: You're absolutely right, it broke focusing even without an explicit user action. Fixed in 66bfc976b by making the deferred code really idempotent. CC @mshoho

src/web/CustomScrollbar.ts
ladipro: Fixed in 66bfc976b23. Thanks!

src/web/ViewBase.tsx
ladipro: I see. I wonder if there's an easy way to use the new behavior only if we know that it's going to help. Maybe only if a popup is being rendered for the first time. I'll report back shortly, thanks!

ladipro:
I have reverted the first commit and instead made ViewBase defer the entire _checkAndReportLayout() if a popup was just shown. This solves the unwanted sync layout problem with popups without incurring any cost to the common path.

ladipro:
>> I don't see a way to work around _recalcPosition. We need that information before we can display the popup in its intended position. Deferring the work will not help the perceived performance and will likely hurt it.

Got it. One layout should be fine. The problem with the current code is that layout is triggered repeatedly and the browser is doing throw-away work just to satisfy our reads. Thanks!

PR #525, opened 2018-02-25
Upgrade typescript to 2.7.2, enable strictPropertyInitialization checks

berickson1:
Strict initialization checks validate that every value that does not list ```undefined``` will be assigned in the constructor. Fixed a bunch of places where properties could be undefined but their types didn't reflect that


erictraut:
Nice work. Thanks for doing this.

PR #527, opened 2018-02-27
Making Windows client to use the default implementation for Linking 

sivachalla:
Making Windows client to use the default implementation for Linking from native-common folder.
@erictraut  We also need this to be back-ported to 0.51 so that we could consume in S4L. Please let me know if I need to do anything.


erictraut:
@sivachalla, I cherry-picked it into the 0.51.x branch.

PR #533, opened 2018-03-01
Fill pageX/Y & clientX/Y in MouseEvent events

reseul:
Fill pageX/Y & clientX/Y in MouseEvent events passed in various onMouse* handlers and onContextMenu (changed the signature on this a bit).


erictraut:
Looks good!

PR #534, opened 2018-03-02
Implement cacheable popups

ladipro:
Popups that are lightweight, expected to be shown repeatedly, and easy to reset to
their initial state are good candidates for this feature. To elect to be "hideable",
a popup sets hideOnDismiss to true and optionally subscribes to onReshow().

When a hideable popup is dismissed, instead of removing it from the tree, it is still being
rendered, only with visibility: hidden. Then when the same popup is brought up again,
its onReshow() callback is called to give it a chance to reset its state, and visibility
goes back to its default value.

This massively improves the time required to bring up the popup as it saves a bunch of
work done by React when a new node subtree is introduced.

--

Picker popups (reactions, emoticons) in Skype are shown 10x faster when made hideable, with little impact on memory footprint. This is a web-only implementation for now. More platforms will be added if this approach proves useful.

### DIFF
--- a/src/native-common/FrontLayerViewManager.tsx
+++ b/src/native-common/FrontLayerViewManager.tsx
@@ -13,6 +13,7 @@ import React = require('react');
 import RN = require('react-native');
 import SubscribableEvent from 'subscribableevent';
 
+import App from './App';
 import { ModalContainer } from '../native-common/ModalContainer';
 import PopupContainerView from './PopupContainerView';
 import Types = require('../common/Types');
@@ -41,6 +42,12 @@ export class FrontLayerViewManager {
     private _cachedPopups: PopupStackContext[] = [];
 
     event_changed = new SubscribableEvent<() => void>();
+
+    constructor() {
+        App.memoryWarningEvent.subscribe(() => {
+            this._cachedPopups = [];
+        });
+    }
 
     public showModal(modal: React.ReactElement<Types.ViewProps>, modalId: string, options?: Types.ModalOptions): void {
         const index = this._findIndexOfModal(modalId);
@@ -164,28 +171,17 @@ export class FrontLayerViewManager {
     private _renderPopup(context: PopupStackContext, hidden: boolean): JSX.Element {
         const key = (context.popupOptions.cacheable ? 'CP:' : 'P:') + context.popupId;
         return (
-            <ModalContainer
+            <PopupContainerView
                 key={ key }
-                hidden={ hidden }>
-                <RN.TouchableWithoutFeedback
-                    onPressOut={ this._onBackgroundPressed }
-                    importantForAccessibility={ 'no' }
-                >
-                    <RN.View
-                        style={ _styles.fullScreenView }>
-                        <PopupContainerView
-                            popupOptions={ context.popupOptions }
-                            anchorHandle={ hidden ? undefined : context.anchorHandle }
-                            onDismissPopup={ hidden ? undefined : () => this.dismissPopup(context.popupId) }
-                            hidden={ hidden }
-                        />
-                    </RN.View>
-                </RN.TouchableWithoutFeedback>
-            </ModalContainer>
+                popupOptions={ context.popupOptions }
+                anchorHandle={ hidden ? undefined : context.anchorHandle }
+                onDismissPopup={ hidden ? undefined : () => this.dismissPopup(context.popupId) }
+                hidden={ hidden }
+                />
         );
     }
 
-    public getPopupLayerView(rootViewId?: string | null): JSX.Element[] | null {
+    public getPopupLayerView(rootViewId?: string | null): JSX.Element | null {
         if (rootViewId === null) {
             // The Popup layer is only supported on root views that have set an id, or
             // the default root view (which has an undefined id)
@@ -204,7 +200,22 @@ export class FrontLayerViewManager {
         }
         this._cachedPopups.map(context => popupContainerViews.push(this._renderPopup(context, true)));
 
-        return popupContainerViews;
+        if (popupContainerViews.length > 0) {
+            return (
+                <ModalContainer hidden={ !overlayContext }>
+                    <RN.TouchableWithoutFeedback
+                        onPressOut={ this._onBackgroundPressed }
+                        importantForAccessibility={ 'no' }
+                        >
+                        <RN.View
+                            style={ _styles.fullScreenView }>
+                            { popupContainerViews }
+                        </RN.View>
+                    </RN.TouchableWithoutFeedback>
+                </ModalContainer>
+            );
+        }
+        return null;
     }
 
     private _onBackgroundPressed = (e: RN.SyntheticEvent<any>) => {
--- a/src/native-common/FrontLayerViewManager.tsx
+++ b/src/native-common/FrontLayerViewManager.tsx
@@ -36,6 +36,7 @@ const _styles = {
 
 export class FrontLayerViewManager {
     private _overlayStack: (ModalStackContext | PopupStackContext)[] = [];
+    private _cachedPopups: PopupStackContext[] = [];
 
     event_changed = new SubscribableEvent<() => void>();
 
@@ -82,6 +83,11 @@ export class FrontLayerViewManager {
         if (index === -1) {
             const nodeHandle = RN.findNodeHandle(popupOptions.getAnchor());
             if (nodeHandle) {
+                if (popupOptions.cacheable) {
+                    // The popup is transitioning from cached to active.
+                    this._cachedPopups = this._cachedPopups.filter(popup => popup.popupId !== popupId);
+                }
+
                 this._overlayStack.push(new PopupStackContext(popupId, popupOptions, nodeHandle));
                 this.event_changed.fire();
                 return true;    
@@ -96,6 +102,11 @@ export class FrontLayerViewManager {
             const popupContext = this._overlayStack[index] as PopupStackContext;
             if (popupContext.popupOptions.onDismiss) {
                 popupContext.popupOptions.onDismiss();
+            }
+
+            if (popupContext.popupOptions.cacheable) {
+                // The popup is transitioning from active to cached.
+                this._cachedPopups.push(popupContext);
             }
 
             this._overlayStack.splice(index, 1);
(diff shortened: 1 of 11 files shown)


src/web/PopupContainer.tsx
```diff
@@ -0,0 +1,104 @@
...
+* Common parent of all components rendered into a popup. Calls onShow and onHide
+* callbacks when the popup is hidden (i.e. "closed" but still rendered as hidden)
+* and re-shown.
+*/
+
+import React = require('react');
+import PropTypes = require('prop-types');
+import FocusManager from './utils/FocusManager';
+
+export interface PopupContainerProps {
```
erictraut: This should extend RX.Types.CommonProps to pick up the 'children' and 'style' fields.

src/web/PopupContainer.tsx
```diff
@@ -0,0 +1,104 @@
...
+        return {
+            focusManager: this.context.focusManager,
+            popupContainer: this as PopupContainer
+        };
+    }
+
+    render() {
+        let style = this.props.style;
+        if (this.props.hidden) {
+            style['visibility'] = 'hidden';
```
erictraut: You're modifying the object passed into props by the caller. You should clone it before adding to it. Or, better yet, pass in an array of styles to the div below. That way, you can avoid cloning it.

erictraut:
I'm not convinced we should do this. It's problematic from an API standpoint because it's adding complexity and exposing implementation details. I've never seen any other popup API in any other framework that exposes a "hide on dismiss" feature — and for good reason. It's inelegant.

From an implementation perspective, I have the following concerns. All of these could be addressed relatively easily.
1. There doesn't appear to be a way to deallocate a popup that uses this feature — even when explicitly calling dismiss(). Such a popup will stick around in memory for the lifetime of the app. That's not a good design.
2. While a popup is allocated, a periodic timer is running. This timer checks the location of the active popup to see if it needs to be repositioned or resized. With this feature, this timer will run indefinitely for the lifetime of the application.
3. The ```isDisplayed``` method will return true if a popup is dismissed but hidden.
4. If we were to implement something like this, it should be implemented for all platforms.

Here's a variation of your proposal that I think I could get behind. It exposes less complexity in the interface and is more aligned with React's programming philosophy.
1. We eliminate the ```hideOnDismiss``` and ```onReshow``` but add a ```cacheable: boolean``` field. This is simply a hint that tells the framework that it's OK for the contents of a popup not to be unmounted when dismissed. It doesn't guarantee that it won't be unmounted. The framework can decide to free up resources if and when it wants. In all other respects, a dismissed popup acts the same whether it's cacheable or not.
2. If cacheable is true, we use your technique to hide the popup, but we disable the reposition timer I describe above if there are no visible popups. We allow up to N such cacheable popups to remain mounted at any given time, where N is some small number like 4. We can use a simple MRU technique to manage the cache. We will need to render up to N+1 instances of PopupContainerView — the hidden cacheable ones plus the active one (if it's not cacheable).
3. When a caller shows a popup, its renderPopup method is called and it needs to render it in its entirety. If its popupId is cached _and_ the caller specifies "cached" as true in the second call to ```RX.Popup.show```, we'll render it using the cached PopupContainer. If the contents haven't changed since the last time it was rendered (or have changed relatively little), React's VDOM comparison will eliminate all (or most) of the DOM updates. This won't be quite as fast as your first implementation, which completely eliminates the second render, but it will still be much faster than today.
4. We'll still need to go through a measure pass before unhiding a cached popup because it may need to be resized or put in a new position, since we weren't actively tracking it with a reposition timer.
5. If ```isDisplayed``` is called, it should return true only if there's a visible popup. Cached (but hidden) popups are not considered "displayed".
6. This should be implemented for both web and native-common.

ladipro:
Thank you for the feedback!

> There doesn't appear to be a way to deallocate a popup that uses this feature — even when explicitly calling dismiss(). Such a popup will stick around in memory for the lifetime of the app. That's not a good design.

We never have more than one popup allocated. In showPopup() we always assign to this._activePopupOptions. So effectively the caching strategy is MRU with N == 1. Or at least that was my intention for this first implementation. Is there really something that stick around for the lifetime of the app? Thanks!

> When a caller shows a popup, its renderPopup method is called and it needs to render it in its entirety. If its popupId is cached and the caller specifies "cached" as true in the second call to RX.Popup.show, we'll render it using the cached PopupContainer. If the contents haven't changed since the last time it was rendered (or have changed relatively little), React's VDOM comparison will eliminate all (or most) of the DOM updates. This won't be quite as fast as your first implementation, which completely eliminates the second render, but it will still be much faster than today.

I don't believe that my implementation completely eliminates the second render. We still always (i.e. even when re-showing) end up in _renderRootView() and then call the active popup's renderPopup(). Actually, I believe that it does exactly what you're proposing. And that's the part that convinced me that this is worth doing because it is kind of elegant. We call renderPopup() exactly as we do without this feature, the difference is in the React workload because the old and new tree are expected to be very similar.

src/web/PopupContainer.tsx
ladipro: Done.

src/web/PopupContainer.tsx
ladipro: Fixed by cloning. Div doesn't seem to take style arrays.

```
interface HTMLAttributes<T> extends DOMAttributes<T> {
        ...
        style?: CSSProperties;
        ...
```

src/native-common/FrontLayerViewManager.tsx
```diff
@@ -34,6 +34,8 @@ const _styles = {
     }
 };
 
+const MAX_CACHED_POPUPS = 4;
```
ladipro: Ideally this would be controller by an app policy. Do we have a mechanism for apps to pass global settings down to ReactXP?

erictraut: No, and I don't think this should be controlled by app policy. There are many caches within React and React Native that are opaque to the app. Exposing all of these knobs is unnecessary and would just add complexity.

src/native-common/PopupContainerView.tsx
```diff
@@ -113,9 +111,13 @@ export class PopupContainerView extends React.Component<PopupContainerViewProps,
     }
 
     render() {
-        var popupView = this.props.activePopupOptions.renderPopup(
-            this.state.anchorPosition, this.state.anchorOffset,
-            this.state.constrainedPopupWidth, this.state.constrainedPopupHeight);
+        const popupView = (this.props.hidden ?
+            this.props.popupOptions.renderPopup('top', 0, 0, 0) :
```
ladipro: We could remember the arguments from when the popup was active and keep passing them when it's not. It doesn't seem to make much difference and probably not worth the added complexity.

src/web/FrontLayerViewManager.tsx
```diff
@@ -88,6 +91,16 @@ export class FrontLayerViewManager {
...
         }
 
+        if (options.cacheable) {
+            // New popup is transitioning from cached to active.
+            this._cachedPopups = this._cachedPopups.filter(popup => popup.id !== popupId);
+        }
+        if (this._activePopupOptions && this._activePopupOptions.cacheable) {
+            // Old popup is transitioning from active to cached.
+            this._cachedPopups.push({ options: this._activePopupOptions, id: this._activePopupId!!! });
+            this._cachedPopups = this._cachedPopups.slice(-MAX_CACHED_POPUPS);
```
erictraut: Are you sure slice() does the right thing if you pass a negative value that's greater than the current length of the array? I think it does, but the documentation for the method doesn't specify this case.

src/web/PopupContainer.tsx
```diff
@@ -0,0 +1,106 @@
+/**
```
erictraut: How different is this from the native implementation? If it's very close, we should collapse the two into a common file rather than duplicating the code.

src/web/RootView.tsx
```diff
@@ -23,11 +23,13 @@ import Styles from './Styles';
 import Types = require('../common/Types');
 import FocusManager from './utils/FocusManager';
 import UserInterface from './UserInterface';
+import PopupContainer from './PopupContainer';
 
 export interface RootViewProps {
     mainView?: React.ReactNode;
     modal?: React.ReactElement<Types.ViewProps>;
-    activePopupOptions?: Types.PopupOptions;
+    activePopupOptions?: { options: Types.PopupOptions, id: string };
```
erictraut: See comment above about using a named type here rather than duplicating the type definition.

erictraut: Also, the name "activePopupOptions" should be changed since this is no longer just the options. Perhaps "activePopup" or "activePopupInfo". Same below for "cachedPopupOptions".

src/native-common/FrontLayerViewManager.tsx
```diff
@@ -147,39 +161,50 @@ export class FrontLayerViewManager {
         return !!(options === rootViewId || options && options.rootViewId === rootViewId);
     }
 
-    public getPopupLayerView(rootViewId?: string | null): React.ReactElement<any> | null {
+    private _renderPopup(context: PopupStackContext, hidden: boolean): JSX.Element {
+        const key = (context.popupOptions.cacheable ? 'CP:' : 'P:') + context.popupId;
+        return (
+            <ModalContainer
```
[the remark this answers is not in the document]
ladipro: I did it after noticing that ModalContainer usurps the screen when rendered. Fixed by rendering it as hidden if no popup is active. Thanks!

src/native-common/FrontLayerViewManager.tsx
```diff
@@ -34,8 +34,11 @@ const _styles = {
     }
 };
 
+const MAX_CACHED_POPUPS = 4;
+
 export class FrontLayerViewManager {
     private _overlayStack: (ModalStackContext | PopupStackContext)[] = [];
+    private _cachedPopups: PopupStackContext[] = [];
```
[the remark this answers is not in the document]
ladipro: Done.

src/web/FrontLayerViewManager.tsx
```diff
@@ -88,6 +91,16 @@ export class FrontLayerViewManager {
             }
         }
 
+        if (options.cacheable) {
+            // New popup is transitioning from cached to active.
+            this._cachedPopups = this._cachedPopups.filter(popup => popup.id !== popupId);
```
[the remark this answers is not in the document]
ladipro: Ah, I completely missed that the caller can mutate cacheable. Fixed.

src/web/FrontLayerViewManager.tsx
```diff
@@ -88,6 +91,16 @@ export class FrontLayerViewManager {
...
+        }
+        if (this._activePopupOptions && this._activePopupOptions.cacheable) {
+            // Old popup is transitioning from active to cached.
+            this._cachedPopups.push({ options: this._activePopupOptions, id: this._activePopupId!!! });
```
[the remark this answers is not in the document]
ladipro: D'oh! Fixed.

src/web/FrontLayerViewManager.tsx
ladipro: ECMA 262 22.1.3.23 Array.prototype.slice ( start, end )
...
4. If relativeStart < 0, let k be max((len + relativeStart), 0); else let k be min(relativeStart, len).
...

Based on this I'd say that it is well specified. If this language spec is not relevant for this project I'll fix it.

src/web/PopupContainer.tsx
ladipro: They are very different. Perhaps some of the positioning logic in web/RootView.tsx could be moved to web/PopupContainer.tsx and then a subset of it extracted to a common base but that feels like a separate piece of work.

src/web/RootView.tsx
ladipro: Done.

src/native-common/FrontLayerViewManager.tsx
```diff
@@ -147,38 +168,53 @@ export class FrontLayerViewManager {
...
+    private _renderPopup(context: PopupStackContext, hidden: boolean): JSX.Element {
+        const key = (context.popupOptions.cacheable ? 'CP:' : 'P:') + context.popupId;
+        return (
+            <PopupContainerView
+                key={ key }
+                popupOptions={ context.popupOptions }
+                anchorHandle={ hidden ? undefined : context.anchorHandle }
+                onDismissPopup={ hidden ? undefined : () => this.dismissPopup(context.popupId) }
+                hidden={ hidden }
+                />
```
erictraut: Nit: Indented one tab too much.

PR #538, opened 2018-03-05
Make VirtualListView use a generic type for VirtualListViewItemInfo

Fitzpasd:
The example [in the docs](https://microsoft.github.io/reactxp/docs/extensions/virtuallistview.html) won't compile in some TS configurations due to: 
```
Types of parameters 'item' and 'item' are 
Type 'VirtualListViewItemInfo' is not assignable to type 'FruitListItemInfo'.
Property 'text' is missing in type 'VirtualListViewItemInfo'.
```

This PR fixes that issue by using a generic type for the item info.

### DIFF
--- a/extensions/virtuallistview/src/VirtualListView.tsx
+++ b/extensions/virtuallistview/src/VirtualListView.tsx
@@ -63,12 +63,12 @@ export interface VirtualListViewItemInfo extends VirtualListCellInfo {
     isNavigable?: boolean;
 }
 
-export interface VirtualListViewProps extends RX.CommonStyledProps<RX.Types.ViewStyleRuleSet> {
+export interface VirtualListViewProps<ItemInfo extends VirtualListViewItemInfo> extends RX.CommonStyledProps<RX.Types.ViewStyleRuleSet> {
     // Ordered list of descriptors for items to display in the list.
-    itemList: VirtualListViewItemInfo[];
+    itemList: ItemInfo[];
 
     // Callback for rendering item when it becomes visible within view port.
-    renderItem: (item: VirtualListViewItemInfo, hasFocus?: boolean) => JSX.Element | JSX.Element[];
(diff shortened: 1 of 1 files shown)


hnager:
@mshoho you should probably take a look at this

erictraut:
Which typescript options are you using? I'm not able to repro the errors you're seeing.

erictraut:
Incidentally, we are talking about doing a major update to the VLV in the near future. We've been making a number of performance and accessibility improvements on a private version, and I'd like to get those changes into the public version. We're going to do some additional internal testing on the new code before we do this.

Fitzpasd:
Here's my tsconfig.json that was giving this error:

```
{
  "compilerOptions": {
    "allowUnreachableCode": true,
    "allowSyntheticDefaultImports": true,
    "baseUrl": ".",
    "declaration": false,
    "experimentalDecorators": true,
    "importHelpers": true,
    "jsx": "react",
    "lib": [
      "dom",
      "es2016",
      "es2017.object"
    ],
    "module": "es2015",
    "moduleResolution": "node",
    "noEmit": true,
    "noResolve": false,
    "noImplicitAny": true,
    "noImplicitReturns": true,
    "paths": {
      "*": [ 
        "*",
        "src/*" 
      ]
    },
    "reactNamespace": "RX",
    "sourceMap": true,
    "strict": true,
    "target": "es2015"
  },
  "exclude": [
    "node_modules"
  ]
}
```

Typescript 2.7.2, VSC 1.20.1

Fitzpasd:
Looking forward to seeing the new VLV.  I'd love to see a `horiztonal` feature :).  If not, it's something I might look to add when I get some time (sadly probably not for at least a month).

erictraut:
I'm not in favor of adding a horizontal feature. The VLV is already very complicated in its current form. Adding more flexibility to it will make it even more complicated and difficult to maintain and tune. In my experience, the content within horizontally-scrolling views is typically relatively small or is effectively unbounded in size (think photos in a large library). In either case, a generalized VLV isn't a good solution.

PR #546, opened 2018-03-14
Replaced onInput with onChange event for TextInput in web

RomanLysogor:
Fixing TextInput warning  when running on web:

> Failed form propType: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`."

Since ReactJS is [already](https://github.com/facebook/react/pull/4003/files) superseding onInput vs onChange it is safe to subscribe to onChange event as well.


erictraut:
Behavior of onChange and onInput are different. For example, onInput is triggered when pressing arrow keys or the delete key when the contents are empty. You can't simply replace one for the other without breaking existing software.

PR #547, opened 2018-03-14
Add Image.prefetchAndGetCachedPath

rigdern:
When prefetching an image, enables consumers to find out where on disk the image gets cached to. Useful if you need a local file path for an image rather than a URL and if you're okay with there being no guarantees around how long the local path will exist.

`RX.Image.prefetchAndGetCachedPath` is currently only interesting on Windows. On other platforms, it has the same effect as `RX.Image.prefetch`.

We've only invested in implementing this on Windows so far because, in our app, this API is needed for a Windows-specific scenario which requires calling a native API which accepts an image as a local file path but not as a URL. We can extend this API to other platforms in the future if needed.

React Native Windows added support for `prefetchAndGetCachedPath` in this commit: https://github.com/Microsoft/react-native-windows/commit/e0c0d820d5a932d9063be47537c47688ee6357e9

### DIFF
--- a/docs/docs/components/image.md
+++ b/docs/docs/components/image.md
@@ -61,4 +61,19 @@ getNativeHeight(): number;
 getNativeWidth(): number;
 ```
 
+## Static Methods
+```javascript
+// Prefetches a remote image and stores it in a cache. This can decrease the
+// amount of time it takes when you later want to show the image because the
+// image only has to be fetched from the local cache rather than over the
+// network.
+prefetch(url: string): Promise<boolean>;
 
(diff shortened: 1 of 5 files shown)


erictraut:
I don't think we should provide this. It's not possible to implement it on the web, so it doesn't belong in the ReactXP abstraction. There are also no ReactXP APIs that take a file path as a parameter, so there's nothing a caller can do with a path if it's returned.

rigdern:
Those are good points. We'll expose this at a different layer.

PR #557, opened 2018-03-19
Expose onWheel on RX.View

ladipro:
Useful for custom scrolling, zooming, pagination, etc.


ladipro:
So this is primarily for web and desktop native.

I quickly checked UWP and it doesn't support mouse wheel events inside scroll views. It admittedly makes sense. Do we generally go for the lowest common denominator or is it ok to leave web unrestricted?

erictraut:
Wheel events are already supported in the RX.GestureView component. I don't think we should further complicate RX.View by exposing it there as well.

ladipro:
I see, yes, please ignore. I'll open a separate PR with the mouse eventing test. Thanks!

PR #571, opened 2018-03-23
Implement hidden scroll indicators in web/ScrollView

ladipro:
Sometimes it's OK to leave it up to touch and the mouse wheel to do the scrolling
instead of cluttering the UI with scrollbars.

This commit makes the web ScrollView support the showsHorizontalScrollIndicator
and showsVerticalScrollIndicator props.


erictraut:
Looks good. Thanks for including test code!

PR #575, opened 2018-03-26
Add onCut support for TextInput

pronebird:
This PR adds `onCut` support for web platform. I tested and it works in my app.

### DIFF
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -873,6 +873,7 @@ export interface TextInputPropsShared extends CommonProps, CommonAccessibilityPr
     onFocus?: (e: FocusEvent) => void;
     onBlur?: () => void;
     onPaste?: (e: ClipboardEvent) => void;
+    onCut?: (e: ClipboardEvent) => void;
     onChangeText?: (newValue: string) => void;
     onSelectionChange?: (start: number, end: number) => void;
     onSubmitEditing?: () => void;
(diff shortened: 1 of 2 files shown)


erictraut:
Sorry for the slow reply. I thought I had already responded to this PR. 

I'd like to understand the scenario in which "onCut" is useful — especially given that it's not supported by React Native platforms. There's really no difference from the app's perspective between a "cut" operation and a "delete" operation. Both can be detected using the onChangeText callback.

pronebird:
Hi @erictraut,

I have a managed input that automatically formats the entered text and selection as user types. I do not handle `OnChangeText` and instead listen to key events via `onKeyPress` and `onPaste` and `onCut`.

This all works well except that `onCut` is not implemented in ReactXP so my input simply does nothing on cut, it does not delete text either because it's fully managed.

erictraut:
While it's possible to implement "onCut" on the web, this won't work on RN platforms without submitting changes to the various RN implementations. How do you propose to make your implementation work on native platforms?

I faced a similar problem, and I was able to work around this limitation using onChangeText. It was a bit of extra work because I needed to infer cut/delete operations. This technique also worked for system-level text modifications like auto-correct replacement.

pronebird:
That's right. This is likely has to be solved in the upstream dependencies first.

However, I am not able to find a single mention of `onPaste` on RN repository either. So it seems that RX already makes an exception for paste event. I am not saying there should be more exceptions though.

pronebird:
I think we should close this PR for now. We can continue this discussion later if RN ever introduces the paste, copy and cut events.

PR #582, opened 2018-03-30
Bump RXPTest reactxp dependency

bclermont:
Fix

```
ERROR in [at-loader] ./src/Tests/PopupTest.tsx:291:13 
    TS2345: Argument of type '{ dismissIfShown: true; cacheable: boolean; getAnchor: () => Button; getElementTriggeringPopup: (...' is not assignable to parameter of type 'PopupOptions'.
  Object literal may only specify known properties, and 'cacheable' does not exist in type 'PopupOptions'.
```


bclermont:
close because https://github.com/Microsoft/reactxp/commit/1575093b43fcf4ae2cdabc5c195b74b0985189d9

PR #585, opened 2018-04-02
Ignore tabIndex=-1 in focusFirst().

mshoho:
Elements with tabIndex=-1 should not be focused during the keyboard navigation using Tab and focusFirst() is used in the Tab navigation.

### DIFF
--- a/src/web/utils/FocusManager.ts
+++ b/src/web/utils/FocusManager.ts
@@ -114,7 +114,7 @@ export class FocusManager extends FocusManagerBase {
                 storedComponent.limitedCount === 0 &&
                 storedComponent.limitedCountAccessible === 0)
             .map(storedComponent => ReactDOM.findDOMNode(storedComponent.component) as HTMLElement)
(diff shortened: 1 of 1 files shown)


src/web/utils/FocusManager.ts
```diff
@@ -114,7 +114,7 @@ export class FocusManager extends FocusManagerBase {
                 storedComponent.limitedCount === 0 &&
                 storedComponent.limitedCountAccessible === 0)
             .map(storedComponent => ReactDOM.findDOMNode(storedComponent.component) as HTMLElement)
-            .filter(el => el && el.focus);
+            .filter(el => el && el.focus && (el.tabIndex !== -1));
```
erictraut: If I'm reading the HTML spec correctly, we should be checking for any negative number here, not just -1.

mshoho: Yes, that's correct, we just never use anything lower than -1. Updated to consider any negative tabIndex.

PR #589, opened 2018-04-10
Ramyasen/reactxp disabledopacity override

RamyaSenk:
Currently there is no way to override the disabled opacity value of a button. There are requirements where this is needed. 
In code using react xp, we can have a workaround. We can override opacity and in the on press handler - we simply return. 

This is a cleaner way of doing it.

### DIFF
--- a/docs/docs/components/button.md
+++ b/docs/docs/components/button.md
@@ -34,6 +34,9 @@ delayLongPress: number = 1000;
 // If disabled, touch and mouse input events are ignored
 disabled: boolean = false;
 
+// By default, opacity of a disabled element is 0.5. This value be overriden with this property
+disabledOpacity: number = undefined;
+
 // Called when VoiceOver is on and the user double tapped to
 // activate a control
 onAccessibilityTapIOS: (e: SyntheticEvent) => void; // iOS Only
--- a/docs/docs/components/button.md
+++ b/docs/docs/components/button.md
@@ -34,7 +34,7 @@ delayLongPress: number = 1000;
 // If disabled, touch and mouse input events are ignored
 disabled: boolean = false;
 
-// By default, opacity of a disabled element is 0.5. This value be overriden with this property
+// By default, opacity of a disabled element is 0.5. This value can be overriden with this property
 disabledOpacity: number = undefined;
 
 // Called when VoiceOver is on and the user double tapped to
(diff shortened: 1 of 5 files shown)


docs/docs/components/button.md
```diff
@@ -34,6 +34,9 @@ delayLongPress: number = 1000;
 // If disabled, touch and mouse input events are ignored
 disabled: boolean = false;
 
+// By default, opacity of a disabled element is 0.5. This value can be overriden with this property
+disabledOpacity: number = undefined;
```
erictraut: If you're adding a new prop, it needs to be documented in Button.md and added to the RXPTest app.

src/native-common/Button.tsx
```diff
@@ -128,11 +128,18 @@ export class Button extends React.Component<Types.ButtonProps, {}> {
             _defaultAccessibilityTrait);
 
         const opacityStyle = !this.props.disableTouchOpacityAnimation && this._opacityAnimatedStyle;
+        let disabledStyle = this.props.disabled && _styles.disabled;
+
+        if (this.props.disabled && this.props.disabledOpacity) {
```
erictraut: I think the check should be this.props.disabledOpacity !== undefined. As it's written, the conditional will evaluate to false if someone explicitly sets the disabledOpacity to 0, which seems like a valid (although not recommended) action.

src/web/Button.tsx
```diff
@@ -157,7 +154,7 @@ export class Button extends React.Component<Types.ButtonProps, {}> {
         }
 
         if (this.props.disabled) {
-            buttonStyleMutations.opacity = 0.5;
+            buttonStyleMutations.opacity = this.props.disabledOpacity || 0.5;
```
erictraut: This also doesn't properly handle the case where disabledOpacity is specified as 0.

RamyaSenk: Yes, thanks for catching it.

src/native-common/Button.tsx
RamyaSenk: Yes will do - thanks

docs/docs/components/button.md
RamyaSenk: Ive already added it to button.md -> will add it to the test

RamyaSenk: @erictraut i have added this change to the test - any thing else is required from my side?

PR #592, opened 2018-04-11
update virtual list to support on screen change

lneir:
adds support for callback 'onScreenChanged' that will notify when an item has come onto the screen or left the screen view.

### DIFF
--- a/extensions/virtuallistview/package.json
+++ b/extensions/virtuallistview/package.json
@@ -1,6 +1,6 @@
 {
   "name": "reactxp-virtuallistview",
-  "version": "0.1.9",
+  "version": "0.1.9-sym-1.0",
   "description": "General-purpose virtualized list view built on the cross-platform ReactXP library",
   "author": "ReactXP Team <reactxp@microsoft.com>",
   "license": "MIT",
--- a/extensions/virtuallistview/src/VirtualListCell.tsx
+++ b/extensions/virtuallistview/src/VirtualListCell.tsx
@@ -12,6 +12,7 @@ import RX = require('reactxp');
 
 export interface VirtualListCellInfo {
     key: string;
+    onScreenChanged?: (isOnScreen: boolean) => void;
 }
 
 export interface VirtualListCellProps extends RX.CommonProps {
@@ -91,6 +92,7 @@ export class VirtualListCell extends RX.Component<VirtualListCellProps, null> {
     private static _hiddenTopValue = -32768;
 
     private _isVisible = false;
+    private _isOnScreen = false;
     private _top: number = VirtualListCell._hiddenTopValue;
     private _calculatedHeight = 0;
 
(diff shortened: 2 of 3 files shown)


extensions/virtuallistview/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "reactxp-virtuallistview",
-  "version": "0.1.9",
+  "version": "0.1.9-sym-1.0",
```
erictraut: Please don't change the package version. I'll increment it if/when I publish a new official version.

extensions/virtuallistview/src/VirtualListView.tsx
```diff
@@ -1010,8 +1010,11 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
         // Set the "live" values as well.
         let cell = this.refs[cellInfo.virtualKey] as VirtualListCell;
         if (cell) {
+            // on screen when full message is within layoutHeight area.
+            let isOnScreen = top > this._lastScrollTop && (top + cellInfo.height) < this._lastScrollTop + this._layoutHeight;
```
erictraut: This check requires that the item be on screen in its entirety, as opposed to partly on screen. It's not clear to me that's what all callers will want. It depends on the use case.

extensions/virtuallistview/src/VirtualListCell.tsx
```diff
@@ -12,6 +12,7 @@ import RX = require('reactxp');
 
 export interface VirtualListCellInfo {
     key: string;
+    onScreenChanged?: (isOnScreen: boolean) => void;
```
erictraut: An interface change like this should be accompanied by an update to the documentation (VirtualListView.md).

erictraut: The terminology here isn't ideal. "Screen" usually implies the full physical display surface, whereas this is referring to the view port of the list.

extensions/virtuallistview/src/VirtualListCell.tsx
```diff
@@ -223,6 +225,13 @@ export class VirtualListCell extends RX.Component<VirtualListCellProps, null> {
         return this._isVisible;
     }
 
+    setIsOnScreen(isOnScreen: boolean) {
```
erictraut: I don't like exposing the notion of "visible in view port" at the list cell level. A list cell shouldn't have this knowledge. The visible range is something that's maintained by the list, so if we add a callback, it seems like it should be exposed by the VirtualListView, not the VirtualListCell.

extensions/virtuallistview/package.json
govindvarshney: This was more done for internal purposes. We have our own internal repository.

govindvarshney:
"_We've talked about maybe adding the same interface to VirtualListView, but we haven't had a use case for it so far. Would this solution meet your needs?_" That would be ideal.

PR #593, opened 2018-04-12
Fix popup API reentrancy issues

ladipro:
The onDismiss() callback may attempt to invoke popup API; a common case is
calling Popup.dismiss() on the very same popup from within the callback.
This used to work before #534 as long as the callback implemented its own
reentrancy check. After #534 we crash in FrontLayerViewManager.dismissPopup
accessing the cacheable field on the already-nulled-out _activePopupOptions.

Fixed in this commit by always calling onDismiss() last, after the state has
been mutated.

### DIFF
--- a/docs/docs/apis/popup.md
+++ b/docs/docs/apis/popup.md
@@ -47,7 +47,8 @@ interface PopupOptions {
     // with button and chevron icon.
     getElementTriggeringPopup?: () => React.Component<any, any>;
 
-    // Called when the popup is dismissed.
+    // Called when the popup is dismissed. Popup.isDisplayed() will return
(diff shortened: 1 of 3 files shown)


docs/docs/apis/popup.md
```diff
@@ -47,7 +47,8 @@ interface PopupOptions {
     // with button and chevron icon.
     getElementTriggeringPopup?: () => React.Component<any, any>;
 
-    // Called when the popup is dismissed.
+    // Called when the popup is dismissed. Popup.isDisplayed() will return
```
ladipro: Or maybe it would be safer to say that the effect of calling Popup API from within this callback is undefined.

PR #594, opened 2018-04-12
Accessibility improvements

mshoho:
1. A powerful autoFocus logic to simplify the application logic and solve some long lasting accessibility issues (the motivation is described in the docs).
2. Adjusting TextInput aria-live workaround to for VoiceOver on Mac to enable it only for the input which is being edited.
3. Fixing the bug in FocusManager when the focus wan't restoring properly (in the keyboard navigation mode) to a component which was focused before the focus restriction.


docs/docs/components/textinput.md
```diff
@@ -24,8 +24,9 @@ autoCapitalize: 'none' | 'sentences' | 'words' | 'characters';
 // Should auto-correction be applied to contents?
 autoCorrect: boolean = true;
 
-// Should focus be applied to text input on componentDidMount?
-autoFocus: boolean = false;
+// Should be focused when the component is mounted,
+// see https://microsoft.github.io/reactxp/docs/apis/focusutils.html
+autoFocus: AutoFocus | AutoFocus[] = AutoFocus.No;
```
erictraut: See note above. Let's make this an array, not a complex type that's either an array or a scalar.

docs/docs/apis/focusutils.md
```diff
@@ -0,0 +1,78 @@
...
+
+//     // We can target the keyboard navigation mode (if none of
+//     // WhenNavigatingWithKeyboard and WhenNavigatingWithoutKeyboard are specified,
+//     // both are assumed to be enabled).
+//     WhenNavigatingWithKeyboard,
+//     WhenNavigatingWithoutKeyboard,
+
+//     // We can target the platform (if none of the platforms are specified, all
+//     // platforms are assumed to be enabled).
+//     Android,
```
erictraut: Exposing specific platforms within an enumeration is not an approach we should be using. It's not scalable, and it goes against the principles of ReactXP. These should be removed. If the caller wants to make it conditional based on platform, they can include the check within their app code.

docs/docs/apis/focusutils.md
```diff
@@ -0,0 +1,78 @@
...
+//     // win if present, otherwise the close button will be focused).
+//     // Default priority is PriorityLow.
+//     PriorityLow,
+//     PriorityHigh,
+//     PriorityHighest, // Highest priority is used internally and you shouldn't use
+//                      // it unless you are really sure what you're up to.
+
+//     // Sometimes it might be needed to delay the autofocusing a little (if no
+//     // delay is specified, the component will be focused without the delay).
+//     Delay100, // 100 ms
```
erictraut: We shouldn't expose discrete time delays as enumerations. This should be a separate field, and it should be encoded as a millisecond value (i.e. a number).

Also, it's not clear to me why someone would ever need to defer auto focus. The comment says "sometimes it might be needed", but it's not clear why.

docs/docs/apis/focusutils.md
```diff
@@ -0,0 +1,78 @@
...
+// Another thing is that it is often required to have different autofocusing logic
+// on different platforms. For example it is often ok to automatically focus the
+// input field on desktop when something shows, but on mobile it will bring up the
+// keyboard which is not always the desired behaviour.
+
+// To tackle above we have this concept of smart automatic focus.
+// See the Types.AutoFocus enum definition:
+
+// export enum AutoFocus {
+//     No = 0,
```
erictraut: This interface is a mess. You're trying to map a complex structure to an enumeration. It should be an object with fields that map to the individual concepts.

docs/docs/apis/focusutils.md
```diff
@@ -0,0 +1,78 @@
...
+//     // Sometimes a common high level component has default autofocusable (for
+//     // example, close button in a dialog), but the subcomponents want to be
+//     // autofocused too (for example, some input inside a particular kind of dialog).
+//     // We can specify the priority (Low for the close button, High for the
+//     // input) without forking the logic on the application level (the input will
+//     // win if present, otherwise the close button will be focused).
+//     // Default priority is PriorityLow.
+//     PriorityLow,
+//     PriorityHigh,
+//     PriorityHighest, // Highest priority is used internally and you shouldn't use
```
erictraut: We shouldn't be exposing internal-only values in a public interface.

docs/docs/apis/focusutils.md
```diff
@@ -0,0 +1,78 @@
...
+//     // delay is specified, the component will be focused without the delay).
+//     Delay100, // 100 ms
+//     Delay500, // 500 ms
+//     Delay1000 // 1000 ms
+// }
+
+// In general, using autoFocus property of the components which support it, should
+// be enough, however sometimes we want to schedule something for automatic focus
+// when an already mounted component is being updated.
+autoFocus(value: Types.AutoFocus|Types.AutoFocus[], focus: () => void, isAvailable: () => boolean): boolean;
```
erictraut: I don't understand why we'd expose a method called "autoFocus". The "auto" part implies that it's automatic, not something that needs to be programmatically invoked. It makes sense for "auto focus" to be a prop, but not a method.

src/common/Types.ts
```diff
@@ -835,7 +881,7 @@ export interface LinkProps extends CommonStyledProps<LinkStyleRuleSet> {
 export interface TextInputPropsShared extends CommonProps, CommonAccessibilityProps {
     autoCapitalize?: 'none' | 'sentences' | 'words' | 'characters';
     autoCorrect?: boolean;
-    autoFocus?: boolean;
+    autoFocus?: AutoFocus | AutoFocus[]; // Should autofocus depending on the set of the specified conditions
```
erictraut: Changing the type of this prop will result in a breaking change to the existing interface. There needs to be a _really_ compelling reason to do so, and I don't think there is in this case. We should continue to support the old prop but also include support for a new prop (with a different name to avoid conflict).

docs/docs/apis/focusutils.md
```diff
@@ -0,0 +1,78 @@
...
+//     Mac,
+
+//     // Sometimes a common high level component has default autofocusable (for
+//     // example, close button in a dialog), but the subcomponents want to be
+//     // autofocused too (for example, some input inside a particular kind of dialog).
+//     // We can specify the priority (Low for the close button, High for the
+//     // input) without forking the logic on the application level (the input will
+//     // win if present, otherwise the close button will be focused).
+//     // Default priority is PriorityLow.
+//     PriorityLow,
```
erictraut: I don't like this way of exposing the concept. Given that you have only three discrete priority values, it's not clear how they compose. Also, components need to know about the values passed by other components, which is not a good design.

I assume that these conflicts don't occur very often. When they do, can't we apply a simple heuristic — like use the one that's "deeper" in the view hierarchy? If that's not possible, it would probably be better to use a numeric value rather than two or three discrete priorities. This is how, for example, zIndex and taxIndex work.

src/common/utils/AutoFocusHelper.ts
```diff
@@ -0,0 +1,198 @@
...
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT license.
+*
+* Provides the logic to decide if the component needs to be autofocused
+* on mount, depending on a combination of AutoFocus enum values.
+*/
+
+import Types = require('../Types');
+
+let _isAndroid = false;
```
erictraut: This module shouldn't have any knowledge of individual platforms.

docs/docs/components/view.md
```diff
@@ -79,6 +79,10 @@ restrictFocusWithin: boolean = false;
 // changing it during the View life cycle will produce an error.
 limitFocusWithin: LimitFocusType = LimitFocusType.Unlimited;
 
+// Should be focused when the component is mounted,
+// see https://microsoft.github.io/reactxp/docs/apis/focusutils.html
+autoFocus: AutoFocusValue = undefined;
```
zholobov: >autoFocus [](start = 0, length = 9)

looks like this should trigger focusable rendering for UWP. I'll add Radu who owns that.

zholobov:
@radu is added to the review. #Closed

zholobov:
@reseul can you take a look at this? Looks like autoFocus on View should trigger focusable creation for RNW?

zholobov:
@reseul is added to the review. #Closed

docs/docs/components/button.md
```diff
@@ -37,6 +37,16 @@ disabled: boolean = false;
 // By default, opacity of a disabled element is 0.5. This value can be overriden with this property
 disabledOpacity: number = undefined;
 
+// Should be focused when the component is mounted, see also View's arbitrateFocus
+// property.
+// WARNING: autoFocus=true means that this Button's focus() method will be called,
+// however calling focus() might have no effect (for example the button is disabled),
+// the application has to handle this either while setting this property or in the
+// FocusArbitrator callback.
+// NOTE: The focus() call will be performed on the next tick after the current
```
erictraut: Is it necessary to document that it "will be performed on the next tick..."? This seems like an implementation detail that is unimportant to document. If we do document it, we'll be constrained to implement it this way on all platforms in the future.

src/common/Interfaces.ts
```diff
@@ -37,10 +37,12 @@ export abstract class AnimatedText extends AnimatedComponent<Types.AnimatedTextP
 export abstract class AnimatedTextInput extends AnimatedComponent<Types.AnimatedTextInputProps, {}> {
 }
 
-export abstract class AnimatedView extends AnimatedComponent<Types.AnimatedViewProps, {}> {
-    abstract focus(): void;
+export abstract class AnimatedView extends AnimatedComponent<Types.AnimatedViewProps, {}> implements FocusableComponent {
     abstract setFocusRestricted(restricted: boolean): void;
     abstract setFocusLimited(limited: boolean): void;
+    abstract focus(): void;
+    abstract realFocus(): void;
```
erictraut: I don't think "realFocus" is a good name. All methods should start with a verb. And "real" is ambiguous in this context. Perhaps we should call this method "focus", and the other method (the one currently called "focus") should be called "requestFocus". This will make it clear that "focus" takes place immediately, the same as "blur". And "requestFocus" is something that is deferred and arbitrated.

src/common/Types.ts
```diff
@@ -449,6 +450,26 @@ export enum AccessibilityTrait {
...
+export interface FocusCandidate {
+    // An instance of the component which wants to be focused.
+    component: RX.FocusableComponent;
+
+    // Returns component.props.accessibilityId (if specified).
+    accessibilityId?: string;
+
+    // If the candidate is inside the View with arbitrateFocus property specified,
+    // returns accessibilityId of that View.
+    parentAccessibilityId?: string;
```
erictraut: Why would we include this field? This should be the same for all candidates passed to the focus arbitrator. And the caller should already know what the parent accessibility ID is because the focus arbitrator was installed on that node. I recommend removing this.

src/common/Types.ts
```diff
@@ -449,6 +450,26 @@ export enum AccessibilityTrait {
...
+// To set this callback use View's arbitrateFocus property.
+export type FocusArbitrator = (candidates: FocusCandidate[]) => FocusCandidate | undefined;
+
+// FocusArbitrator callback will be called with an array of FocusCandidate.
+// See View's arbitrateFocus property.
+export interface FocusCandidate {
+    // An instance of the component which wants to be focused.
+    component: RX.FocusableComponent;
+
+    // Returns component.props.accessibilityId (if specified).
```
erictraut: Nit: "Returns" should be "Value of" (since it doesn't really "return" the value).

src/common/Types.ts
```diff
@@ -585,9 +621,15 @@ export interface ViewPropsShared extends CommonProps, CommonAccessibilityProps {
     shouldRasterizeIOS?: boolean; // iOS-only prop, if view should be rendered as a bitmap before compositing
     viewLayerTypeAndroid?: ViewLayerType; // Android only property
 
-    restrictFocusWithin?: boolean; // Web-only, during the keyboard navigation, the focus will not go outside this view
+    restrictFocusWithin?: RestrictFocusType; // Web-only, during the keyboard navigation, the focus will not go outside this view
```
erictraut: This appears to be a  breaking change. We'll need to leave this as a boolean value and find another way of exposing this new functionality if it's needed.

docs/docs/components/button.md
```diff
@@ -101,5 +109,20 @@ underlayColor: string = undefined; // iOS and Android only
...
 focus(): void;
+
+// The preferable way to focus the component. When requestFocus() is called,
+// the actual focus() will be deferred, and if requestFocus() has been
+// called for several components, only one of those components will actually
+// get a focus() call. By default, last component for which requestFocus() is
+// called will get a focus() call, but you can specify arbitrateFocus property
+// of a parent View and provide the callback to decide which one of that View's
+// descendants should be focused. This is useful for the accessibility: when
+// cosecutive focus() calls happen one after another, the next one interrupts
```
erictraut: Minor: cosecutive => consecutive

docs/docs/components/link.md
```diff
@@ -65,5 +65,22 @@ style: LinkStyleRuleSet | LinkStyleRuleSet[] = [];
...
+focus(): void;
+
+// The preferable way to focus the component. When requestFocus() is called,
+// the actual focus() will be deferred, and if requestFocus() has been
+// called for several components, only one of those components will actually
+// get a focus() call. By default, last component for which requestFocus() is
+// called will get a focus() call, but you can specify arbitrateFocus property
+// of a parent View and provide the callback to decide which one of that View's
+// descendants should be focused. This is useful for the accessibility: when
+// cosecutive focus() calls happen one after another, the next one interrupts
```
erictraut: Same here.

docs/docs/components/text.md
```diff
@@ -70,6 +78,21 @@ style: TextStyleRuleSet | TextStyleRuleSet[] = [];
...
 focus(): void;
+
+// The preferable way to focus the component. When requestFocus() is called,
+// the actual focus() will be deferred, and if requestFocus() has been
+// called for several components, only one of those components will actually
+// get a focus() call. By default, last component for which requestFocus() is
+// called will get a focus() call, but you can specify arbitrateFocus property
+// of a parent View and provide the callback to decide which one of that View's
+// descendants should be focused. This is useful for the accessibility: when
+// cosecutive focus() calls happen one after another, the next one interrupts
```
erictraut: Same here.

docs/docs/components/button.md
```diff
@@ -37,6 +37,14 @@ disabled: boolean = false;
 // By default, opacity of a disabled element is 0.5. This value can be overriden with this property
 disabledOpacity: number = undefined;
 
+// Should be focused when the component is mounted, see also View's arbitrateFocus
+// property.
+// WARNING: autoFocus=true means that this Button's focus() method will be called,
```
erictraut: Should this now say "requestFocus() method" rather than "focus() method"?

src/common/Types.ts
```diff
@@ -362,6 +362,7 @@ export interface Stateless {}
 export interface CommonAccessibilityProps {
     // iOS, Android, and Desktop
     importantForAccessibility?: ImportantForAccessibility;
+    accessibilityId?: string;
```
erictraut: This new prop should be added to the documentation for all applicable components.

src/common/Types.ts
```diff
@@ -550,6 +568,9 @@ export interface TextPropsShared extends CommonProps {
     textBreakStrategy?: 'highQuality' | 'simple' | 'balanced';
 
     importantForAccessibility?: ImportantForAccessibility;
+    accessibilityId?: string;
+
+    autoFocus?: boolean; // The component is a candidate for being autofocused.
```
erictraut: This prop should be added to the documentation for Button, Text, etc.

PR #595, opened 2018-04-12
Making clipboard copy text to handle line breaks

RamyaSenk:
Currently when copying text from reactxp method for web - if text contains new lines (\r) and we paste it outside of a web environment - the new lines are not preserved and each newlins gets replaced by a space instead.

Fix: Replacing span element with a text area and using \r\n instead of only \r -> fixes the issue. The solution has been inspired by other sources online. 

The copy implementation works fine on mobile environments.

### DIFF
--- a/src/web/Clipboard.ts
+++ b/src/web/Clipboard.ts
@@ -13,7 +13,10 @@ import SyncTasks = require('synctasks');
 export class Clipboard extends RX.Clipboard {
     public setText(text: string) {
         let node = Clipboard._createInvisibleNode();
-        node.textContent = text;
+        // Replace carraige return /r with /r/n, so that pasting outside browser environment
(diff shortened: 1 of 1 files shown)


src/web/Clipboard.ts
```diff
@@ -13,7 +13,10 @@ import SyncTasks = require('synctasks');
 export class Clipboard extends RX.Clipboard {
     public setText(text: string) {
         let node = Clipboard._createInvisibleNode();
-        node.textContent = text;
+        // Replace carriage return /r with /r/n, so that pasting outside browser environment
+        // (eg in a native app) preserves this new line
+        text = text.replace(/(?:\\[r])+/g, '\r\n');
+        node.innerHTML = text;
```
plisy: Is replacing multiple \r-s with single \r\n intended?

PR #596, opened 2018-04-13
[Web] Preventing the aria-live container used for announcements from receiving screen reader focus

tbombach:
- Adding a short timer that clears the text in the `aria-live` container after adding it (triggered by an `announceForAccessibility` event). This change prevents the container from being focusable via screen reader navigation after the announcement finishes. The timer gives the browser enough time to recognize the text addition and queue up the announcement.
- Moving the `newAnnouncementReady` subscription and `aria-live` container to a separate component called `AccessibilityAnnouncer` to move announcements out of the `RootView` React lifecycle.

### DIFF
--- a/src/web/AccessibilityAnnouncer.tsx
+++ b/src/web/AccessibilityAnnouncer.tsx
@@ -1,0 +1,139 @@
+ /**
+* AccessibilityAnnouncer.tsx
+*
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT license.
+*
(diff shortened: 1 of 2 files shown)


tbombach:
We've tested this change in Skype - including cases when an announcement is queued while the screen reader is reading a long label. We see the intended behavior - the screen reader makes the announcement after reading the current label.

Additionally, we've tested this change with a really long announcement. We see the intended behavior - the screen reader does not interrupt the announcement, it finishes reading the entire string.

PR #599, opened 2018-04-17
[Windows] Add onContextMenu callback to GestureView

thezero:
When using RX.GestureView as an interaction handler, it's impossible to detect right mouse button clicks right now in UWP. This info is already propagated from native code in the event received by RX, but it's hidden by GestureView. This PR passes the information further so that user can detect right clicks.

### DIFF
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -703,7 +703,7 @@ export interface TapGestureState extends GestureState {
     clientY: number;
     pageX: number;
     pageY: number;
-    isRightButton?: boolean; // UWP only, for desktop context menu
+    button: MouseButton;
 }
 
 export enum GestureMouseCursor {
@@ -1149,9 +1149,15 @@ export interface ClipboardEvent extends SyntheticEvent {
 
 export type FocusEvent = SyntheticEvent;
 
+export enum MouseButton {
+    Primary,
+    Auxiliary,
+    Secondary
+}
+
 export interface MouseEvent extends SyntheticEvent {
     altKey: boolean;
-    button: number;
+    button: MouseButton;
     clientX: number;
     clientY: number;
     ctrlKey: boolean;
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -1149,15 +1149,9 @@ export interface ClipboardEvent extends SyntheticEvent {
 
 export type FocusEvent = SyntheticEvent;
 
-export enum MouseButton {
-    Primary,
-    Auxiliary,
-    Secondary
-}
-
 export interface MouseEvent extends SyntheticEvent {
     altKey: boolean;
-    button: MouseButton;
+    button: number;
     clientX: number;
     clientY: number;
     ctrlKey: boolean;
@@ -1204,9 +1198,6 @@ export interface TouchEvent extends SyntheticEvent {
     pageX?: number;
     pageY?: number;
     touches: TouchList;
-    button?: number; // Mac, web native events
-    isRightButton?: boolean; // UWP only
-    isMiddleButton?: boolean;
 }
 
 export interface WheelEvent extends SyntheticEvent {
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -703,7 +703,6 @@ export interface TapGestureState extends GestureState {
     clientY: number;
     pageX: number;
     pageY: number;
-    button: MouseButton;
 }
 
 export enum GestureMouseCursor {
@@ -733,6 +732,7 @@ export interface GestureViewProps extends CommonStyledProps<ViewStyleRuleSet>, C
     onPanHorizontal?: (gestureState: PanGestureState) => void;
     onTap?: (gestureState: TapGestureState) => void;
     onDoubleTap?: (gestureState: TapGestureState) => void;
+    onContextMenuGesture?: (gestureState: TapGestureState) => void;
 
     // We can set vertical or horizontal as preferred
     preferredPan?: PreferredPanGesture;
@@ -1205,6 +1205,7 @@ export interface TouchEvent extends SyntheticEvent {
     pageY?: number;
     touches: TouchList;
     isRightButton?: boolean; // UWP only
+    isMiddleButton?: boolean;
 }
 
 export interface WheelEvent extends SyntheticEvent {
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -703,6 +703,7 @@ export interface TapGestureState extends GestureState {
     clientY: number;
     pageX: number;
     pageY: number;
+    isRightButton?: boolean; // UWP only, for desktop context menu
 }
 
 export enum GestureMouseCursor {
@@ -1197,6 +1198,7 @@ export interface TouchEvent extends SyntheticEvent {
     pageX?: number;
     pageY?: number;
     touches: TouchList;
+    isRightButton?: boolean; // UWP only
 }
 
 export interface WheelEvent extends SyntheticEvent {
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -1204,6 +1204,7 @@ export interface TouchEvent extends SyntheticEvent {
     pageX?: number;
     pageY?: number;
     touches: TouchList;
+    button?: number; // Mac, web native events
     isRightButton?: boolean; // UWP only
     isMiddleButton?: boolean;
 }
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -732,7 +732,7 @@ export interface GestureViewProps extends CommonStyledProps<ViewStyleRuleSet>, C
     onPanHorizontal?: (gestureState: PanGestureState) => void;
     onTap?: (gestureState: TapGestureState) => void;
     onDoubleTap?: (gestureState: TapGestureState) => void;
-    onContextMenuGesture?: (gestureState: TapGestureState) => void;
+    onContextMenu?: (gestureState: TapGestureState) => void;
 
     // We can set vertical or horizontal as preferred
     preferredPan?: PreferredPanGesture;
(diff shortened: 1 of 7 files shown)


src/common/Types.ts
```diff
@@ -703,6 +703,7 @@ export interface TapGestureState extends GestureState {
     clientY: number;
     pageX: number;
     pageY: number;
+    isRightButton?: boolean; // UWP only, for desktop context menu
```
erictraut: For consistency with MouseEvent, this should be called "button", it should return the integer corresponding to the button, and it should be implemented for all mouse-based platforms (not just UWP).

erictraut: Also, since this is an addition to the public interface, it will need to be documented (in GestureView.md) and added to the ReactXP test app (GestureViewTest.tsx).

thezero:
Changes:

- Removed isRightButton property from original PR
- Added `onContextMenuGesture(gestureState: Types.TapGestureState)` to GestureView with multiplatform support

src/common/Types.ts
```diff
@@ -1148,9 +1149,15 @@ export interface ClipboardEvent extends SyntheticEvent {
...
+export enum MouseButton {
+    Primary,
+    Auxiliary,
+    Secondary
+}
+
 export interface MouseEvent extends SyntheticEvent {
     altKey: boolean;
-    button: number;
+    button: MouseButton;
```
erictraut: This is a breaking interface change. While I like the use of an enumeration here, it's not worth the pain associated with breaking existing code. Please revert back to "number".

src/common/Types.ts
```diff
@@ -1197,6 +1204,9 @@ export interface TouchEvent extends SyntheticEvent {
     pageX?: number;
     pageY?: number;
     touches: TouchList;
+    button?: number; // Mac, web native events
```
erictraut: We shouldn't reflect this differently on Mac/web/UWP. We should use "button" on all three platforms. Please remove isRightButton and isMiddleButton.

src/common/Types.ts
```diff
@@ -1148,9 +1149,15 @@ export interface ClipboardEvent extends SyntheticEvent {
 
 export type FocusEvent = SyntheticEvent;
 
+export enum MouseButton {
+    Primary,
```
erictraut: Please remove this enumeration since its use in MouseEvent will result in a breaking change.

src/native-common/GestureView.tsx
```diff
@@ -501,7 +502,21 @@ export abstract class GestureView extends ViewBase<Types.GestureViewProps, {}> {
     }
 
     private _sendTapEvent(e: Types.TouchEvent) {
-        if (this.props.onTap) {
+        const button = EventHelpers.toMouseButton(e);
+        if (button === Types.MouseButton.Secondary) {
+            // always handle secondary button, even if context menu is not set - it shouldn't trigger onTap
```
erictraut: Nit - for consistency, capitalize comment and end in period.

src/native-common/GestureView.tsx
```diff
@@ -515,6 +530,16 @@ export abstract class GestureView extends ViewBase<Types.GestureViewProps, {}> {
     }
 
     private _sendDoubleTapEvent(e: Types.TouchEvent) {
+        // if user did a double click with different mouse buttons, eg. left (50ms) right
```
erictraut: Nit - for consistency, capitalize comment and end in period.

src/web/GestureView.tsx
```diff
@@ -183,6 +184,25 @@ export class GestureView extends RX.ViewBase<Types.GestureViewProps, {}> {
         }
     }
 
+    private _sendContextMenuEvent = (e: React.MouseEvent<any>) => {
+        if (this.props.onContextMenuGesture) {
+            const clientRect = this._getGestureViewClientRect();
+
+            if (clientRect) {
+                const tapEvent: Types.TapGestureState = {
```
erictraut: The TapGestureState doesn't derive from SyntheticEvent, so it doesn't offer the standard event methods preventDefault and stopPropagation. I think that's a problem here. I think we have two options:
1. Change the TapGestureState to inherit from SyntheticEvent and implement these two methods - and fill in the other fields that go along with SyntheticEvent
2. Always call e.preventDefault() and e.stopPropagation() for onContextMenu in a gesture view and document it as such.

Option #1 is more general and arguably more consistent, but it's a big change. For that reason, I am leaning to option #2. Let's discuss.

src/common/Types.ts
thezero: That's not up to me - there are attributes coming directly from native code. While I agree it would be great to have them unified, it would require changes in react-native-windows.

src/web/GestureView.tsx
thezero: I think #2 is consistent with all the other events GestureView is handling - none of them bubble down. CMIIW, but you can't get onPress on a parent. So the only thing missing here is possibly `e.stopPropagation()` as `e.preventDefault` is a few lines below already.

src/common/Types.ts
```diff
@@ -732,6 +732,7 @@ export interface GestureViewProps extends CommonStyledProps<ViewStyleRuleSet>, C
     onPanHorizontal?: (gestureState: PanGestureState) => void;
     onTap?: (gestureState: TapGestureState) => void;
     onDoubleTap?: (gestureState: TapGestureState) => void;
+    onContextMenuGesture?: (gestureState: TapGestureState) => void;
```
erictraut: This should be "onContextMenu" for consistency. We don't add the word "Gesture" in any other cases.

src/native-common/utils/EventHelpers.ts
```diff
@@ -189,6 +183,19 @@ export class EventHelpers {
         return mouseEvent;
     }
 
+    toMouseButton(e: Types.TouchEvent): number {
+        const nativeEvent = e as any;
+        if (nativeEvent.button !== undefined) {
+            return nativeEvent.button;
+        } else if (nativeEvent.isRightButton) {
```
erictraut: Are you sure "isRightButton" shouldn't be capitalized? It was in the old code. Same with IsMiddleButton. Was the old code incorrect?

src/web/GestureView.tsx
erictraut: Yes, that's a good argument. Please add the call to e.stopPropagation.

src/common/Types.ts
thezero: This can't be `onContextMenu` because of conflicting interface. As discussed in the chat you created:

> `onContextMenu` is defined as `(e: MouseEvent)` which 1) doesn't make sense for touch interface 2) would be hard to fake due to how GestureView works now and would require serious refactoring of the component.
> 
> All the other callbacks for GestureView are `(gestureState: TapGestureState)` so my proposal is to keep up with this and add `onContextMenuGesture(gestureState: TapGestureState).` Thoughts?

erictraut: Yeah, I realize I used the name "onContextMenuGesture" in the chat, but now that I'm reviewing the change, I think "onContextMenu" is better for consistency. There's no conflict with ViewPropsShared because GestureProps has no inheritance relationship with ViewPropsShared.

The GestureView documentation also needs to be updated to include the new prop.

thezero: Yes, good point about documentation. I'll add it.

> There's no conflict with ViewPropsShared because GestureProps has no inheritance relationship with ViewPropsShared.

src/native-common/GestureView.tsx(42,52): error TS2344: Type 'GestureViewProps' does not satisfy the constraint 'ViewProps'.
  Types of property 'onContextMenu' are incompatible.
    Type '((gestureState: TapGestureState) => void) | undefined' is not assignable to type '((e: MouseEvent) => void) | undefined'.

`GestureView extends ViewBase<Types.GestureViewProps, {}>` -> `ViewBase<P extends Types.ViewProps, S>` -> `ViewProps extends ViewPropsShared`. Therefore `GestureViewProps` need to be interface compatible with `ViewPropsShared`.

erictraut: Yikes, that's a bug. GestureView shouldn't be extending ViewBase like that. Neither should ScrollView. I'm surprised that the TypeScript compiler didn't complain about that because ViewBase requires that "P extends Types.ViewProps", and GestureViewProps doesn't properly extend this interface. I'm working on a fix, and I'll let you know when I'm done.

erictraut: OK, I just pushed a fix for the bug. Turns out it was a problem for some implementations of GestureView, ScrollView and WebView. I've fixed all of them.

src/native-common/utils/EventHelpers.ts
```diff
@@ -239,6 +233,19 @@ export class EventHelpers {
         return mouseEvent;
     }
 
+    toMouseButton(e: Types.TouchEvent): number {
+        const nativeEvent = e as any;
+        if (nativeEvent.button !== undefined) {
+            return nativeEvent.button;
+        } else if (nativeEvent.isRightButton || nativeEvent.IsRightButton) {
```
thezero: As mentioned in previous PR, I tested current implementations and it seems that UpperCase `IsRightButton` is not produced by any native code, but I'd rather keep backwards compatibility.

src/web/GestureView.tsx
```diff
@@ -182,6 +183,27 @@ export class GestureView extends React.Component<Types.GestureViewProps, Types.S
         }
     }
 
+    private _sendContextMenuEvent = (e: React.MouseEvent<any>) => {
+        if (this.props.onContextMenu) {
+            e.preventDefault();
+            e.stopPropagation();
```
thezero: I moved these a bit earlier to keep this behavior on par with the rest on reactxp - events are being cancelled before their synthentic handlers. At the same time if onContextMenu is not defined, I don't think we should prevent the event from bubbling down.

PR #613, opened 2018-04-28
Add automatic prevention of layout-only view optimizations in some cases

reseul:
A: Automatically prevent layout only optimizations mainly in RN for Windows for some cases:

- RX.Button (this is currently also achieved by a default "background: transparent" style, a losing solution in long term).
- RX.View with responder attached
- RX.View with handlers for specific mouse direct events

This is not a complete solution. Some left out cases:
- overflow: hidden/visible - there's a long story here, looking the other way for now
- handlers for bubbled events: no easy solution here.

B: Fixed RXPTest dependencies

### DIFF
--- a/samples/RXPTest/windows/rxptest.sln
+++ b/samples/RXPTest/windows/rxptest.sln
@@ -5,6 +5,9 @@ MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "rxptest", "rxptest\rxptest.csproj", "{B1DF5A70-CE68-416F-94DE-B16C6F796B27}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ReactNative", "..\node_modules\react-native-windows\ReactWindows\ReactNative\ReactNative.csproj", "{C7673AD5-E3AA-468C-A5FD-FA38154E205C}"
+	ProjectSection(ProjectDependencies) = postProject
+		{2EACF721-73B5-46AE-9775-4A8674D05A9C} = {2EACF721-73B5-46AE-9775-4A8674D05A9C}
+	EndProjectSection
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "ReactNative.Shared", "..\node_modules\react-native-windows\ReactWindows\ReactNative.Shared\ReactNative.Shared.shproj", "{EEA8B852-4D07-48E1-8294-A21AB5909FE5}"
 EndProject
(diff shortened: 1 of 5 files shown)


erictraut:
I don't understand why this change is needed. If RNW includes this optimization, it should be smart enough to apply it only when it's inappropriate to do so. Are you trying to work around bugs in the RNW implementation? If so, I'd prefer to just fix them at their source.

reseul:
"Layout only" optimization exists in RN for Android, and RN for Windows inherited it during the initial porting.
It prevents the creation of "intermediate" native views that would just have layouting (and no visual) role.
Optimizations are supposed to be transparent, but this implementation left it somewhat in the middle.

Current implementation does detect "layout only" views by checking whether all properties fall into a hardcoded set of "layout only" properties, for some of them the values being taken into account as well.
For other scenarios a "collapsable" property is exposed for external (to native) callers to override the behavior. AnimatedImplementation uses that flag when native animations are involved, for example.

My rationalization was "RNA/RNW expose the 'collapsible' flag and RXP doesn't, so RXP should do its best to synthesize it".

Fixing in RN could be attempted, I guess:
- would require changes in both RNA & RNW for most of the cases 
- the presence of "direct handlers" that require the view exist can be detected by using boolean-type "pseudo-properties" on native side.
- detecting a view that can get a responder can be done similarly by defining an "onResponderGrant" pseudo-property in RN, but that starts to look weird already.
- "overflow: hidden/overflow" is dealt with incorrectly, but fixing would interfere badly with a reactxp peculiarity (the use of the "hidden" default) 
- other less obvious scenarios may exist, harder to grasp/prove, though that may be an argument for exposing the flag to apps.

Currently apps use "background: transparent" and other "do nothing" tricks like this, not very optimal.




---
In reply to: [385126180](https://github.com/Microsoft/reactxp/pull/613#issuecomment-385126180) [](ancestors = 385126180)

erictraut:
I'm not aware of any "holes" in the Android implementation — cases where the optimization is not transparent. Are you saying that such holes exist in the Android implementation and/or the Windows implementation? If so, we should try to address them. This optimization should be transparent, and I don't see any good reason why it can't be. Dealing with "holes" piecemeal at the ReactXP level is not the a good approach because it leaves users vulnerable to unexpected behaviors.

erictraut:
@reseul Any thoughts on my comment above?

reseul:
Android does have holes, mainly related to the lack of a consistent way to find out whether what seems to be looking like a layout-only view (based solely on properties) is needed for some UI event handling on the JS side. There are also the subtle sabotaging done by the default "overflow: hidden" style applied to all views such that, if RNW were to work by the book, almost no view would end up optimized.

I could try to dig again (cause details are getting fuzzier), but:
- I don't always find most satisfying/elegant solutions.
- App code already had the workarounds in place (for Android) and they were extremely fast with filling the gaps for the windows flavor
- No one seems to be actually crying about this..

Perhaps I'll fix the case of direct mouse events (enter/leave) the RN for Windows added on RNW side sometime, but even that is not a high pri.

So I guess you may close this one.

PR #619, opened 2018-05-08
useSafeInsets prop on RX.View to support rendering within safe area b…

Fitzpasd:
…oundaries for applicable devices.

This is becoming an important scenario as users look to support the iPhone X, and I can see this prop also being used on Android when P is released with their 'notch' support.

All sample projects updated and tested.

### DIFF
--- a/docs/docs/components/view.md
+++ b/docs/docs/components/view.md
@@ -157,8 +157,11 @@ activeOpacity: number = undefined; // iOS and Android only
 // handlers
 underlayColor: string = undefined; // iOS and Android only
 
-// If true, renders children within the safe area boundaries of a device, i.e. with 
-// padding with ensure the children don't cover navigation bars, toolbars etc.
+// When true
+//  - renders children within the safe area boundaries of a device, i.e. with 
+//    padding with ensure the children don't cover navigation bars, toolbars etc.
+//  - Applies a style of { flex: 1, alignSelf: 'stretch' } to this view.
+//  - Some ViewProps may be ignored.
 useSafeInsets: boolean = false; // iOS only
 ```
 
--- a/docs/docs/components/view.md
+++ b/docs/docs/components/view.md
@@ -156,6 +156,10 @@ activeOpacity: number = undefined; // iOS and Android only
 // Background color that will be visible on touch on views that have onPress
 // handlers
 underlayColor: string = undefined; // iOS and Android only
+
+// If true, renders children within the safe area boundaries of a device, i.e. with 
+// padding with ensure the children don't cover navigation bars, toolbars etc.
+useSafeInsets: boolean = undefined; // iOS only
 ```
 
 ## Styles
--- a/docs/docs/components/view.md
+++ b/docs/docs/components/view.md
@@ -159,7 +159,7 @@ underlayColor: string = undefined; // iOS and Android only
 
 // If true, renders children within the safe area boundaries of a device, i.e. with 
 // padding with ensure the children don't cover navigation bars, toolbars etc.
-useSafeInsets: boolean = undefined; // iOS only
+useSafeInsets: boolean = false; // iOS only
 ```
 
 ## Styles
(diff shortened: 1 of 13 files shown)


src/native-common/View.tsx
```diff
@@ -392,11 +398,18 @@ export class View extends ViewBase<Types.ViewProps, Types.Stateless> {
     }
 
     render() {
-        let PotentiallyAnimatedView = this._isButton(this.props) ? RN.Animated.View : RN.View;
+        let ViewToRender = RN.View;
+
+        if (this._isButton(this.props)) {
+            ViewToRender = RN.Animated.View;
+        } else if (this.props.useSafeInsets) {
+            ViewToRender = RN.SafeAreaView;
```
hnager: Is SafeAreaView an Animated.View?

hnager: I guess it doesn't matter since this was previously RN.View, but this means that buttons won't honor safe area insets?

samples/hello-world-js/src/MainPanel.js
```diff
@@ -73,28 +73,30 @@ export default class MainPanel extends RX.Component{
...
-                        Edit App.js to get started
-                    </RX.Text>
-                    <RX.Link style={ styles.docLink } url={ 'https://microsoft.github.io/reactxp/docs' }>
-                        View ReactXP documentation
-                    </RX.Link>
-
-                    <RX.Button style={ styles.roundButton } onPress={ this.props.onPressNavigate }>
-                        <RX.Text style={ styles.buttonText }>
-                            See More Examples
+            <RX.View useSafeInsets={true}>
```
erictraut: Won't this require a style with flex:1 and alignSelf: 'stretch'? Same with other RX.View containers you added to other samples.

src/native-common/View.tsx
erictraut: "useSafeInsets" is defined in ViewProps, not in ViewPropsShared, so it's not available to RX.Animated.View.

docs/docs/components/view.md
```diff
@@ -156,6 +156,10 @@ activeOpacity: number = undefined; // iOS and Android only
 // Background color that will be visible on touch on views that have onPress
 // handlers
 underlayColor: string = undefined; // iOS and Android only
+
+// If true, renders children within the safe area boundaries of a device, i.e. with 
+// padding with ensure the children don't cover navigation bars, toolbars etc.
+useSafeInsets: boolean = undefined; // iOS only
```
erictraut: The default value should be documented as "false", not "undefined".

samples/ImageList/src/ImageListPanel.tsx
```diff
@@ -43,7 +43,7 @@ class ImageListPanel extends RX.Component<{}, ImageListPanelState> {
 
     render() {
         return (
-            <RX.View style={ [_styles.listContainer, RX.StatusBar.isOverlay() && _styles.statusSpacer] }>
+            <RX.View useSafeInsets={true} style={ [_styles.listContainer, RX.StatusBar.isOverlay() && _styles.statusSpacer] }>
```
erictraut: Nit: include spaces within curly braces.

samples/hello-world-js/src/MainPanel.js
Fitzpasd: Since `flex: 1` is always recommended/required, I've added an override in the View class itself to always apply that style.  I didn't think to also apply alignSelf: 'stretch', should I?  Which dev-error would that overcome?

Fitzpasd: Also, do you think it's worth adding a note to the doc that those styles are auto-applied?

erictraut: Yeah, you should add alignSelf: 'stretch' as well. Otherwise the layout engine, by default, sizes the non-axis direction according to the parent's alignItems attribute and the width/height of the children. 

And yeah, the documentation should reflect this style override. Also, the doc should probably also indicate that other props may be ignored when this prop is true.

PR #620, opened 2018-05-08
Pass 'animated' param to setBarStyle on iOS

Fitzpasd:
Looks like this is unintentionally always set to true

### DIFF
--- a/src/ios/StatusBar.ts
+++ b/src/ios/StatusBar.ts
@@ -19,7 +19,7 @@ export class StatusBar extends RX.StatusBar {
     }
 
     setBarStyle(style: 'default' | 'light-content' | 'dark-content', animated: boolean): void {
-        RN.StatusBar.setBarStyle(style, true);
+        RN.StatusBar.setBarStyle(style, animated);
     }
 
     setHidden(hidden: boolean, showHideTransition: 'fade' | 'slide'): void {


Fitzpasd:
@erictraut - I was thinking of adding a `getCurrentHeight()` method to RX.StatusBar, that can pipe into `RN.StatusBar.currentHeight`, that will return the value on Android.  The issue is that this does nothing on iOS.  If we would like to add this method, should we doc it as Android-only, or should we look to add our own implementation for iOS?  iOS is 44 for iPhone X, and 20 for the other devices.

Thought?

hnager:
I would be really careful about adding anything for iOS. Status bar height on iOS can be 20, 44, or 0 (when hidden). To get the height on iOS you would have to query safe area insets on the native side UIApplication.sharedApplication.delegate.window.safeAreaInsets, and would also likely need some stimulus for updating the value, unless you plan to make the interface async.

erictraut:
@fitzpasd, where would such an API be used? On Android (unlike iOS), the status bar doesn't overlap the app's view, so the app shouldn't need to know or care about its height. On iOS, I can see how this would be more important, but the safe view area solves that problem.

hnager:
On Android it seems as if RX.UserInterface.measureWindow() actually includes the status bar height, so knowing the size can come in handy, depending on how you are using measureWindow. I wonder if @Fitzpasd ran in to the same scenario or if it's something else.

Fitzpasd:
@erictraut Setting the status bar to translucent on Android does make it overlap your view.

erictraut:
Ah, I didn't realize that setting it to translucent changed the behavior.

I'm not crazy about having a SafeAreaView on iOS and an API for measuring the status bar on Android. Could we implement the same SafeAreaView mechanism on Android to get the same affect?

hnager:
You could find a way to combine them, although I don't the safe area api (as implemented in RN) allows you to query for the current values, it just applies some padding (? I assume) to move children based on the current values. That would defeat the purpose of making a transparent status bar on Android. The RN safe area implementation seems limited to me in its current state, but I may be misunderstanding what is exposed.

erictraut:
If you use a translucent bar on Android, you probably want to avoid placing any controls beneath the status bar. That's where the useSafeArea feature would come into play. It would apply an inset (padding) at the top to avoid the status bar area. I assume that the reason you'd want to use a translucent bar is so your app's background color/pattern will appear behind it.

erictraut:
We considered implementing an API for querying inset values (including the status bar) instead of the useSafeArea prop. We decided against it because the inset values can change based on situations outside of the app's control, like orientation.

PR #628, opened 2018-05-15
Change Linking.getInitialUrl() to return null, not undefined

Fitzpasd:
`null` is returned by [the RN implementations](https://facebook.github.io/react-native/docs/linking.html#getinitialurl) when an initial url can't be found, not `undefined`.

### DIFF
--- a/src/common/Interfaces.ts
+++ b/src/common/Interfaces.ts
@@ -110,7 +110,7 @@ export abstract class Popup {
 
 export abstract class Linking {
     // Incoming deep links
-    abstract getInitialUrl(): SyncTasks.Promise<string|undefined>;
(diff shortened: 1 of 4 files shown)


erictraut:
I don't think we should change the reactxp interface here. We try to avoid "null" in general and favor undefined. If RN is returning null in some cases, the common-native implementation in reactxp should convert it to undefined.

Fitzpasd:
Sure, I'll make that change in another PR then.

erictraut:
Thanks. I just published 1.1.2-rc.2. I'd like to publish the final 1.1.2 in a few days. If you're able to get this in by then, that'd be great.

PR #631, opened 2018-05-17
added support for Dialog trait to trigger yes-dont-hide importantForA…

zholobov:
…ccessibility behavior. also fixed bug with not accounting for i4a===undefined for the trigger.


zholobov:
@erictraut could you take a look and if it's okay include this in upcoming public release too? this blocks major accessibility use case for us.

erictraut:
@zholobov, looks good. This will be included in upcoming 1.1.2 release.

PR #632, opened 2018-05-17
Added "disableContextMenu" property to RX.Text.

reseul:
This allows the disabling the opening of the OS provided context menu on right clicking selected text.
The fix is in native-common, but it's just the Windows (UWP) version of RN that interprets the property.

### DIFF
--- a/docs/docs/components/text.md
+++ b/docs/docs/components/text.md
@@ -71,6 +71,13 @@ selectable: boolean = false;
 onPress?: (e: SyntheticEvent) => void = undefined;
 onContextMenu?: (e: SyntheticEvent) => void = undefined;
 
+// Disables the default context menu displayed by the OS.
+// By default Windows displays a default context menu on right clicking a selectable 
+// text with a non empty selection, usually with a Copy menu item.
+// Applications that want to provide their own context menu can use this property
+// to prevent the default behavior
+disableContextMenu: boolean = false; // Windows only
+
 // See below for supported styles
 style: TextStyleRuleSet | TextStyleRuleSet[] = [];
 ```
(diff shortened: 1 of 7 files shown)


docs/docs/components/text.md
```diff
@@ -71,6 +71,13 @@ selectable: boolean = false;
 onPress?: (e: SyntheticEvent) => void = undefined;
 onContextMenu?: (e: SyntheticEvent) => void = undefined;
 
+// Disables the default context menu displayed by the OS.
+// By default Windows displays a default context menu on right clicking a selectable 
+// text with a non empty selection, usually with a Copy menu item.
+// Applications that want to provide their own context menu can use this property
+// to prevent the default behavior
+disableContextMenu: boolean = false; // Windows only
```
erictraut: I don't understand why we need this. Isn't this implied if there's an onContextMenu handler provided? Also, this should be disabled by default if the text isn't marked selectable, which it isn't by default.

reseul: 1. Yes, it applies just for the selectable texts.
2. I'm all in for minimizing API changes/needless expansion.
The problem is that onContextMenu is usually set on some parent rather than the RX.Text itself.
Usually the text has no onPress/onContextMenu set, so it keeps itself out of becoming a responder. The honor usually goes to a parent RX.View.
To accommodate this I'd need to add a context field set by whoever is a responder candidate with a non falsy onContextMenu, and have RX.Text check that. I really have to do this at ReactXP level, RN(W) doesn't see the handlers that granularly (right vs left press).
Should I go this way?




---
In reply to: [189486529](https://github.com/Microsoft/reactxp/pull/632#discussion_r189486529) [](ancestors = 189486529)

PR #633, opened 2018-05-17
fix UWP accessibility announce temporary same way Android code path h…

zholobov:
…acked/fixed it (for consistency)


zholobov:
@erictraut if it's not too late please review and include to 1.1.2 release - this fix slipped through cracks.

PR #635, opened 2018-05-18
Fixed import from RNSvgPackage to SvgPackage

michaelperel:
The Android hello-world-js sample would not compile because RNSvgPackage has been updated to SvgPackage. This change was not made in the hello-world-js sample, so this PR makes this change.


deregtd:
Thanks!

PR #652, opened 2018-05-30
Fix  "isNavigatingWithKeyboard" for the multi root-view case

reseul:
1. Propagate "isNavigatingWithKeyboard" across all RootView instances. The flag is global, the truth is in UserInterface instance, and RootView instances continue to drive the changes as before.
2. Fixed the initializing of  "isNavigatingWithKeyboard" when subscribing to changes.
3. Removed some dev logging conflicting with the advertised support of multiple root views.

### DIFF
--- a/src/web/Button.tsx
+++ b/src/web/Button.tsx
@@ -41,13 +41,6 @@ const _styles = {
 const _longPressTime = 1000;
 const _defaultAccessibilityTrait = Types.AccessibilityTrait.Button;
 
-let _isNavigatingWithKeyboard: boolean;
-
-UserInterface.keyboardNavigationEvent.subscribe(isNavigatingWithKeyboard => {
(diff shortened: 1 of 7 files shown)


src/web/Button.tsx
```diff
@@ -41,11 +41,12 @@ const _styles = {
 const _longPressTime = 1000;
 const _defaultAccessibilityTrait = Types.AccessibilityTrait.Button;
 
-let _isNavigatingWithKeyboard = false;
+let _isNavigatingWithKeyboard: boolean;
```
erictraut: Why are we keeping redundant copies of this piece of state in a bunch of modules? It's cheap to access the "truth" version by calling UserInterface.isNavigatingWithKeyboard whenever we need it.

reseul: Yeah, it's treated very carefully, with caching, suppressing of repeated events, etc., a little overdone.
I'll simplify.

PR #656, opened 2018-06-01
Added getMetadata image API

RomanLysogor:
Added new Image API method that exposes RN.Image.getSize. It allows to prefetch image and get imageInfo (currently only dimensions) without extra call.

### DIFF
--- a/docs/docs/components/image.md
+++ b/docs/docs/components/image.md
@@ -70,7 +70,7 @@ getNativeWidth(): number;
 prefetch(url: string): Promise<boolean>;
 
 // Similarly to [prefetch], this method loads a remote image and stores it in
-// a cache. If prefetching was succesfull it will also get image dimensions. It will be
+// a cache. If prefetching was successful it will also get image dimensions. It will be
 // useful if you need [getNativeHeight] or [getNativeWidth] after image
 // was loaded because you will get this info together with prefetching and before
 // you actually need to show the image.
--- a/docs/docs/components/image.md
+++ b/docs/docs/components/image.md
@@ -68,4 +68,11 @@ getNativeWidth(): number;
 // image only has to be fetched from the local cache rather than over the
 // network.
 prefetch(url: string): Promise<boolean>;
+
+// Similarly to [prefetch], this method loads a remote image and stores it in
+// a cache. If prefetching was succesfull it will also get image dimensions. It will be
+// useful if you need [getNativeHeight] or [getNativeWidth] after image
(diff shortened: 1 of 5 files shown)


docs/docs/components/image.md
```diff
@@ -68,4 +68,11 @@ getNativeWidth(): number;
 // image only has to be fetched from the local cache rather than over the
 // network.
 prefetch(url: string): Promise<boolean>;
+
+// Similarly to [prefetch], this method loads a remote image and stores it in
+// a cache. If prefetching was succesfull it will also get image dimensions. It will be
```
erictraut: Minor: should be "successful" (only one "l").

src/native-common/Image.tsx
```diff
@@ -46,6 +46,29 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
         return defer.promise();
     }
 
+    static getMetadata(url: string): SyncTasks.Promise<Types.ImageMetadata> {
+        const defer = SyncTasks.Defer<Types.ImageMetadata>();
+
+        Image.prefetch(url).then(success => {
+            if (!success) {
+                defer.reject('Prefetching url ' + url + ' not succeeded.');
```
erictraut: Minor: should be "did not succeed".

src/web/Image.tsx
```diff
@@ -154,6 +154,30 @@ export class Image extends React.Component<Types.ImageProps, ImageState> {
         return defer.promise();
     }
 
+    static getMetadata(url: string): SyncTasks.Promise<Types.ImageMetadata> {
+        const defer = SyncTasks.Defer<Types.ImageMetadata>();
+
+        const img = new (window as any).Image();
+        img.src = url;
+
+        img.onload = ((event: Event) => {
```
erictraut: Are onload, onerror and onabort guaranteed to be called even if the operation completes before the handlers are specified? If not, we should move "img.src = url" further down so the operation isn't started until after the handlers are installed.

docs/docs/components/image.md
```diff
@@ -68,4 +68,11 @@ getNativeWidth(): number;
...
+// useful if you need [getNativeHeight] or [getNativeWidth] after image
+// was loaded because you will get this info together with prefetching and before
+// you actually need to show the image.
+getMetadata(url: string): Promise<ImageMetadata>
```
erictraut: The new API should be exercised in a test case. Please add to the RXPTest app. You can either add a new top-level test or extend one of the existing Image tests with a new sub-test.

PR #659, opened 2018-06-04
Wire tooltip on buttons,images,links,views

shamy91:
Wire title to tooltip property on Image, View, Button and Link

### DIFF
--- a/src/native-common/Image.tsx
+++ b/src/native-common/Image.tsx
@@ -89,9 +89,9 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
                 accessibilityLabel={ this.props.accessibilityLabel }
                 onLoad={ this.props.onLoad ? this._onLoad : undefined }
                 onError={ this._onError }
-                shouldRasterizeIOS= { this.props.shouldRasterizeIOS }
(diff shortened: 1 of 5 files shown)


src/native-common/Image.tsx
```diff
@@ -91,6 +91,7 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
                 onError={ this._onError }
                 shouldRasterizeIOS= { this.props.shouldRasterizeIOS }
                 { ...additionalProps }
+                tooltip = { this.props.title }
```
erictraut: Minor: Please move this above the expansion of additionalProps. Also, remove spaces around equals sign. (There's also an extra space after the equals for shouldRasterizeIOS — please remove it while you're in there.)

src/typings/react-native.d.ts
```diff
@@ -123,6 +123,9 @@ declare module 'react-native' {
         onLoadEnd?: (e: SyntheticEvent<Image>) => void;
         onLoadStart?: Function;
         onProgress?: Function;
+
+        //windows
```
erictraut: Minor: All comments should have a space after "//" and "Windows" should be capitalized for consistency.

PR #663, opened 2018-06-05
Add drag and drop handlers on Mac OS.

alregner:
Enables drag and drop support on OSX.

### DIFF
--- a/src/macos/View.tsx
+++ b/src/macos/View.tsx
@@ -26,8 +26,7 @@ export class View extends ViewCommon {
                 this._internalProps.allowDrop = true;
 
                 this._internalProps[name] = (e: React.SyntheticEvent<View>) => {
-                    const dndEvent: any = EventHelpers.toMouseEvent(e);
-                    dndEvent.dataTransfer = (e.nativeEvent as any).dataTransfer;
+                    const dndEvent = EventHelpers.toDragEvent(e);
                     handler(dndEvent);
                 };
             }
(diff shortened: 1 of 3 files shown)


src/macos/View.tsx
```diff
@@ -0,0 +1,50 @@
...
+        // Drag and drop related properties
+        for (const name of ['onDragEnter', 'onDragOver', 'onDrop', 'onDragLeave']) {
+            const handler = this._internalProps[name];
+
+            if (handler) {
+                this._internalProps.allowDrop = true;
+
+                this._internalProps[name] = (e: React.SyntheticEvent<View>) => {
+                    handler({
+                        dataTransfer: (e.nativeEvent as any).dataTransfer,
```
erictraut: You are not filling in any of the other fields in the SyntheticEvent. DragEvent derives from MouseEvent, which derives from SyntheticEvent. Each of these adds a bunch of fields including mouse positions, modifier key states, etc.

alregner: I used EventHelpers.toMouseEvent to copy missing fields, I can leave setting dataTransfer fields here or just move it all to a new EventHelpers method, since the same issue is in the Windows implementation. What do you prefer?

erictraut: Might as well create a new EventHelpers method that copies the dataTransfer after it calls toMouseEvent. That way we minimize code duplication.

alregner: Done.

PR #664, opened 2018-06-05
onBlur event parameter

Fitzpasd:
Bringing back this event and fixing the test case so it shows up in the history case.

### DIFF
--- a/samples/RXPTest/src/Tests/TextInputInteractiveTest.tsx
+++ b/samples/RXPTest/src/Tests/TextInputInteractiveTest.tsx
@@ -235,14 +235,16 @@ class TextInputView extends RX.Component<RX.CommonProps, TextInputViewState> {
                     onSubmitEditing={ this._onSubmitEditingTest6 }
                     multiline={ true }
                 />
-                <RX.ScrollView style={ _styles.eventHistoryScrollView }>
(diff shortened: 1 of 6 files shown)


samples/RXPTest/src/Tests/TextInputInteractiveTest.tsx
```diff
@@ -235,14 +235,16 @@ class TextInputView extends RX.Component<RX.CommonProps, TextInputViewState> {
...
                 />
-                <RX.ScrollView style={ _styles.eventHistoryScrollView }>
-                    <RX.Text style={ _styles.eventHistoryText }>
-                        { this.state.test6EventHistory.length ?
-                            this.state.test6EventHistory.join('\n') :
-                            'Event history will appear here'
-                        }
-                    </RX.Text>
-                </RX.ScrollView>
+                <RX.View style={ _styles.eventHistoryScrollView }>
```
Fitzpasd: I couldn't get the ScrollView to show up on Web without adding this container.  Didn't have time to dig into why.  Should we create an issue to investigate?

erictraut:
This change looks good to me. Rather than merging it immediately, I'd like to wait. Until this change is published, it will effectively break the test app for those users who don't go through the trouble of building ReactXP from scratch. I'll merge it when we're getting ready to release the next version (1.3.x).

PR #665, opened 2018-06-06
Add BezierEasing as NPM dependency

a-tarasyuk:


erictraut:
@a-tarasyuk, I needed to revert this change because RN's bundler doesn't appear to be compatible with the layout of the bezier-easing package layout. Webpack works fine, but the RN bundler reports that it can't find the module. I spent a little time digging into it but couldn't find the root cause or an obvious solution. This may be why RN directly copied the code. If you want to investigate further and find a solution, that'd be great. If not, we can leave it as it was previously.

Issue #600, opened 2018-04-17
Support for zIndex


junseld:
Is it possible to add support for `zIndex`? 
or what are the drawbacks/problems to do so?

related issue (react-native supports zIndex.): https://github.com/Microsoft/reactxp/issues/452

I'm not sure if you just are ok with the workarounds and didn't have the time to add it or if there is another problem I missed.

junseld:
btw I currently just use a workaround:
```
const style = RX.Styles.createViewStyle({ 
    ['zIndex' as any]: 1 
}
```

erictraut:
I'd prefer not to add zindex. In my experience, there's always a cleaner way to implement something without using zindex. I'd be more inclined to add it if zIndex were implemented in a consistent fashion across all supported implementations of RN, but that's not currently the case. It has been added to iOS and Android, but they have different behaviors and limitations.

If you have a specific use case in mind for zindex, let's discuss and see if I can convince you that there's a cleaner way to accomplish your objective.

junseld:
ah ok didn't know they had different behaviors.
I just wanted my titlebar to show its shadow (without column-reversing their render order)
after some try&error it worked with that workaround I postet (but didn't test it for iOS yet).

currently I mainly need it for that.

erictraut:
Yeah, column reversing is an easy workaround in your case.

Or if your titlebar is a static height, you can render in normal-column order but leave space at the top. Then include your actual title bar at the end of the item list and use "position: absolute" so it is positioned where you want it.

junseld:
ok thanks. yeah for now setting the zIndex here works pretty good.
But I will keep it in mind for similar cases in the future, thanks.

should I close this issue? I know know why you didn't add zIndex and can understand it (even if I would still like to have it :P)

shameister:
Hey Eric,

 we ran into a similar issue. We are creating a custom component, which shows an input field and auto-suggestions, when typing. We like to reuse our component in several parts of our code, by just using `<TextInputFilterable>`.

![bildschirmfoto 2018-06-11 um 18 13 29](https://user-images.githubusercontent.com/98413/41243730-31d20efc-6da3-11e8-8b5b-58d72ee7cd04.png)
![bildschirmfoto 2018-06-11 um 18 13 31](https://user-images.githubusercontent.com/98413/41243729-31a1d7aa-6da3-11e8-90ba-4e13f4dff1ab.png)


[...]

PR #669, opened 2018-06-08
pointerUp event on Input API

martinbelaska:
Introducing pointerUp event which is fired on the Input API to supplement keyDown/keyUp events.


erictraut:
I don't understand the need for this functionality. Can you elaborate? 

In general, it's a bad idea to expose two mechanisms that do the same thing in a framework. This looks like a redundant piece of functionality.

Pointer events already have a well-defined behavior relating to the view hierarchy. They bubble up from the lower-most view to the upper-most until a view handles it and requests the propagation to stop. A separate global event mechanism raises questions about how this interacts with the existing mechanism — which order are they dispatched and whether the receiver can stop the propagation.

martinbelaska:
We would like to introduce pop-ups, which are not blocking UI and users can still operate the application, while the pop is being shown. This event will be used to close such a pop-up, when user clicks somewhere in the application. 

We could not implement this with current feature set. If the pointer event is handled somewhere on the child, it does not bubble up to the root view and there is no way how to find out that the user actually clicked somewhere.

Regarding the order - global pointerUpEvent is fired after common pointer events and they are independent on each other. With given order, "global" callback could be unsubscribed when the common pointer event is handled somewhere in the child.

erictraut:
ReactXP already implements extensive support for popups. Let's see if we can meet your requirements with those before we add this functionality. What requirement(s) were you unable to meet with the current RX.Popup APIs?

martinbelaska:
What we are missing is "non-modality". We want users to be able to control app while the pop up is shown and dismiss the popup on the first click/tap somewhere out of the popup.

We also have custom animation, but I think it's something what we can sacrifice. The default one could be used there.

erictraut:
OK, that makes sense. I think that a "non-modal" popup would be generally useful. From an interface perspective, we could simply add a field to the PopupOptions to enable it. 

Implementing this might be challenging though. Your PR is a good start, but there are holes that we'd need to address. 
1. There's currently no way in the Windows UWP implementation of RN to install a global "pointer up" handler. We'll look at adding that support.
2. It's not clear how we can handle touch gestures correctly here. Simple taps may be easy to deal with, but more complex gestures are probably not possible.

We will continue to discuss and investigate our options here.

As for your requirements for custom animation, can you explain that one further? Perhaps we could address that through another PopupOption - a callback that is triggered just after the popup is displayed in its final position. You could use that callback to trigger any animations within the popup view. Thoughts?

erictraut:
Incidentally, support for a non-modal popup could also be the answer for [another issue](https://github.com/Microsoft/reactxp/issues/600) that is being discussed.

martinbelaska:
In my PR, I took a shortcut for UWP. I have introduced nativeCommonPointerUpEvent which is dispatched from the common native RootView and also from pressable components. It is fired on the common native input API. Desktop native input API simply listen to this event. When this event occurs, pointerUpEvent is fired. It works reliably on my PC.

Support for gestures is good point. In this iteration, I would personally go just with support of simple taps and clicks. Gesture-end event seems to me as a bit different case (which maybe deserves own implementation of the event on input API).

Regarding the animation, I think that callback after the popup is rendered could be used in our particular case, assuming that we will display transparent pop up and animations will be applied on it.

erictraut:
There are a several problems I see with the PR.

1. On native platforms, it attempts to intercept pointer up events by adding code to all of the RX primitives that intercept taps. This is pretty fragile because it doesn't handle the case where other RN elements or other reactxp extensions are used. The result will be inconsistent behavior depending on where the user taps.

2. On the web, I don't think the proposed implementation will work because the window-level mouseup handler isn't marked as "useCapture", so it will only catch mouseup events that bubble all the way to the window. If any component stops the bubbling (by calling stopPropagation), it will not trigger the event. 

3. We could pass true to the "useCapture" parameter to help address issue 2 above, but that will cause another problem. The order of events would then be inconsistent between native and web. The native implementation triggers the new event after the event is bubbled up the hierarchy, but the web would trigger it before.

I think we should abandon this PR and pursue the alternate approach of exposing support for a "non-modal" popup — i.e. a popup that doesn't swallow clicks/presses in the background. I've started the work to implement such a solution, but I ran into the issues highlighted above. So I think we need to continue to investigate other ways to be reliably notified of a tap or click without intercepting it or preventing it from being swallowed. And we need solutions for both web and native.

martinbelaska:
Sure, if non-modal popups will be implemented differently then we do not have to keep this PR opened. Are you planning to continue working on your 'erictr/nonmodalpopup' branch?

Regarding the events order, could not it be handled directly in the popup component? If we will have 2 pointerup event callback handlers registered, one will be on the level of window and the second one on the level of popup component, both marked as useCapture, then I think that we should be able to determine if event has occurred inside or outside of the popup component. (Only top layer callback executed = click outside of the popup. If the top layer callback and then component callback were executed = tap inside of the popup.)

erictraut:
I've been thinking about how to make this work, but I haven't come up with a solution yet. 

Your suggestion might work on the web, but it's pretty hacky and fragile. That's because we'd need to install the window-level handler with "useCapture" and the popup handler with "useCapture" false. That means the window-level handler will be called first, even if the popup was clicked. Then the popup handler would fire (or not). Within the window-level handler, we'd need to set a short timer that determines whether the popup was clicked and respond accordingly (either dismiss the popup or not). Since this is running on a timer, the DOM could have completely changed in the meantime, which is why I'm somewhat concerned about fragility. I think we could make it work though.

On the native platforms, I haven't come up with any good solution because there's no way to capture taps or clicks at the window level before they are bubbled up through the view hierarchy. I don't think the approach you used in your PR is robust enough for the reasons I mentioned above in point #1. Do you have any other ideas?

martinbelaska:
No, no other ideas. I agree that deferring callback is quite fragile approach. In general, I think that there is no silver bullet to cover all the platforms and it needs to be implemented separately for native and for web.

erictraut:
I think we're blocked on this issue then. Until we have a non-fragile solution, I don't think we should attempt to add support for non-modal popups or for global pointerUp events. As a workaround, you can write your own platform-specific code for your project if you feel comfortable with the limitations and fragility of the approach.

PR #672, opened 2018-06-10
Rename filesGlob to include

a-tarasyuk:


m1ck2:
A-tarasyuk to include what

a-tarasyuk:
@m1ck2
 
> If a segment of a glob pattern includes only * or .*, then only files with supported extensions are included (e.g. .ts, .tsx, and .d.ts by default with .js and .jsx if allowJs is set to true). - [docs#tsconfig-json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html)

@m1ck2 

`filesGlob` is not part of TypeScript [Microsoft/TypeScript-Handbook/issues/334](https://github.com/Microsoft/TypeScript-Handbook/issues/334#issuecomment-230184744
), [Microsoft/TypeScript/pull/5980](https://github.com/Microsoft/TypeScript/pull/5980)

PR #674, opened 2018-06-11
disallow using prefix I in interfaces

a-tarasyuk:
This PR includes standardizing to interface names. 

**Warn:** This change can cause an issue to users who export public interfaces with the prefix `I`. Better to merge it in some of the _major_ releases and add the note in release changes.


erictraut:
There are parts of it that are incorrect, such as the modifications to ".d.ts" files for dependent libraries like react-native. The other changes are to internal implementation names. I don't think it provides sufficient benefit to be worthwhile. We are already careful to standardize all public interface names exported by reactxp and its related extensions.

a-tarasyuk:
Ok, I got. A little bit confusing to have in one file interfaces with `I` and without `I`, for instance 

[`interface IUpdateLayoutAnimationConfig {}`](https://github.com/Microsoft/reactxp/blob/master/src/typings/react-native.d.ts#L976), [`interface AnimationConfig {}`](https://github.com/Microsoft/reactxp/blob/master/src/typings/react-native.d.ts#L937
), 
[`interface IWillAnimatePhaseInfo {}`](https://github.com/Microsoft/reactxp/blob/master/src/web/listAnimations/MonitorListEdits.tsx#L118),[`interface MountedChildrenRef {}`](https://github.com/Microsoft/reactxp/blob/master/src/web/listAnimations/MonitorListEdits.tsx#L126)


If interfaces with the `I` have special meaning I think need to add the comment about that even if these interfaces only for internal usage. What do you think?

erictraut:
I consider react-native.d.ts external to reactxp, even though we have our own private copy checked in. The goal is to eventually switch over to the public @types version. So we shouldn't make any name changes within that file, even in the name of clarity.

PR #677, opened 2018-06-12
Eliminated use of custom react-native.d.ts file in favor of the public version.

erictraut:
Extensions are in the file react-native-extensions.d.ts.

I've made several attempts to do this over the past two years, but the public react-native.d.ts was always too incomplete or too buggy. It has now reached a point where it's feasible to switch to it. This will get us off the constant treadmill of trying to stay up to date with changes to the public interfaces.


src/windows/TextInput.tsx
```diff
@@ -17,11 +17,17 @@ import { TextInput as TextInputBase } from '../native-common/TextInput';
 export class TextInput extends TextInputBase implements FocusManagerFocusableComponent {
 
     protected _render(props: RN.TextInputProps): JSX.Element {
+        // Work around the fact that the react-native type declaration file doesn't
+        // currently define tabIndex and incorrectly defines the onFocus prop.
+        const undefinedProps: any = {
+            tabIndex: this.getTabIndex(),
+            onFocus: (e: React.FocusEvent<any>) => this._onFocusEx(e, props.onFocus)
```
reseul: >onFocus: (e: React.FocusEvent<any>) => this._onFocusEx(e, props.onFocus) [](start = 12, length = 72)

This line gives me an "npm run build" error when run in a standalone enlistment:

Argument of type '((e: NativeSyntheticEvent<TextInputFocusEventData>) => void) | undefined' is not assignable to parameter of type '((e: FocusEvent<any>) => void) | undefined'.
  Type '(e: NativeSyntheticEvent<TextInputFocusEventData>) => void' is not assignable to type '((e: FocusEvent<any>) => void) | undefined'.
    Type '(e: NativeSyntheticEvent<TextInputFocusEventData>) => void' is not assignable to type '(e: FocusEvent<any>) => void'.
      Types of parameters 'e' and 'e' are incompatible.
        Type 'FocusEvent<any>' is not assignable to type 'NativeSyntheticEvent<TextInputFocusEventData>'.
          Types of property 'nativeEvent' are incompatible.
            Type 'FocusEvent' is not assignable to type 'TextInputFocusEventData'.
              Property 'text' is missing in type 'FocusEvent'.

src/windows/ScrollView.tsx
```diff
@@ -9,28 +9,31 @@
...
         //   order for the UI to acknowledge your interaction.
         const keyboardShouldPersistTaps = 'always';
 
+        let nativeProps = {
+            ...props,
+            onKeyDown: onKeyDownCallback,
+            keyboardShouldPersistTaps: keyboardShouldPersistTaps,
+            tabNavigation: this.props.tabNavigation,
+            disableKeyboardBasedScrolling: true
+    } as RN.ScrollViewProps;
```
reseul: > as RN.ScrollViewProps [](start = 5, length = 22)

Is this tradeoff worth it?

src/windows/TextInput.tsx
erictraut: Not sure why this didn't show up before. It's fixed now.

src/typings/react-native-extensions.d.ts
```diff
@@ -0,0 +1,111 @@
...
+*/
+
+import React = require('react');
+import RN = require('react-native');
+
+declare module 'react-native' {
+    //
+    // Focusable view related declarations
+    // ----------------------------------------------------------------------
+    type FocusableWindowsProps<P={}> = P & {
```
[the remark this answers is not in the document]
erictraut: Yeah, you're right. Moved them back.

src/windows/ScrollView.tsx
erictraut: Are you talking about this particular case, or is this a broader question? 

In any case, I think the answer is yes. We've been maintaining this Frankenstein "react-native.d.ts" file for some time now, and it's getting pretty unmaintainable. The use of the public one requires a few more ugly casts in our code, but I still think it's worth it.

reseul: It was a broader Q.

I was thinking whether we can find a better way to describe the new "extended" properties, currently they look like magic (for ex. we use "tabNavigation" because we "know" RNW supports that, but it's not documented anywhere inside ReactXP).
I made an experiment:

1. I added this to the extension file:

    interface ExtendedScrollViewProps extends RN.ScrollViewProps {
         // Windows only
         onKeyDown? : Function;
         onKeyUp? : Function;
         tabNavigation? : 'local' | 'cycle' | 'once';
         disableKeyboardBasedScrolling?: boolean;       
    }

2. Converted the code to:

        let nativeProps: RN.ExtendedScrollViewProps = {
            ...props as RN.ScrollViewProps, // unrelated hack, forces a Types.ScrollViewProps into RN.ScrollViewProps
            onKeyDown: onKeyDownCallback,
            keyboardShouldPersistTaps: keyboardShouldPersistTaps,
            tabNavigation: this.props.tabNavigation,
            disableKeyboardBasedScrolling: true
    };

It seems to work, at least in this case.


---
In reply to: [194625530](https://github.com/Microsoft/reactxp/pull/677#discussion_r194625530) [](ancestors = 194625530)

src/native-common/FrontLayerViewManager.tsx
```diff
@@ -216,8 +216,11 @@ export class FrontLayerViewManager {
         return null;
     }
 
-    private _onBackgroundPressed = (e: RN.SyntheticEvent<any>) => {
-        e.persist();
+    private _onBackgroundPressed = (e: RN.GestureResponderEvent) => {
```
a-tarasyuk: I've added `persist` to `SyntheticEvent` interface - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26474. `RN.GestureResponderEvent` should contain `persist` now.

src/macos/TextInput.tsx
```diff
@@ -42,7 +42,7 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
 
     private _selectionStart: number = 0;
     private _selectionEnd: number = 0;
-    private _mountedComponent: RN.ReactNativeBaseComponent<any, any>|null = null;
+    private _mountedComponent: any;
```
berickson1: shouldn't this be typed as RN.Text?

src/native-common/Button.tsx
```diff
@@ -82,16 +82,17 @@ export class Button extends ButtonBase {
...
+    touchableHandleResponderGrant!: (e: RN.GestureResponderEvent) => void;
+    touchableHandleResponderMove!: (e: RN.GestureResponderEvent) => void;
+    touchableHandleResponderRelease!: (e: RN.GestureResponderEvent) => void;
+    touchableHandleResponderTerminate!: (e: RN.GestureResponderEvent) => void;
 
-    private _isMounted = false;
+    protected _isMounted = false;
     protected _isMouseOver = false;
     protected _isHoverStarted = false;
+    protected _buttonElement: any = null;
```
berickson1: Similar to before (and throughout), shouldn't this be RN.Animated.View?

src/native-common/Image.tsx
```diff
@@ -79,12 +79,12 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
 
     render() {
         // Check if require'd image resource
-        let imageSource: RN.ImageSource | number;
+        let imageSource: RN.ImageURISource | number;
```
a-tarasyuk: There is  type https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L3402
```typescript
export type ImageSourcePropType = ImageURISource | ImageURISource[] | ImageRequireSource;
```

src/native-common/Image.tsx
```diff
@@ -101,20 +101,23 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
         }
 
         const additionalProps = this._getAdditionalProps();
+        const extendedProps: RN.ExtendedImageProps = {
+            source: imageSource,
```
a-tarasyuk: `source` exists in `ImagePropsBase` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L3503

src/native-common/Image.tsx
```diff
@@ -101,20 +101,23 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
...
+            shouldRasterizeIOS: this.props.shouldRasterizeIOS,
```
a-tarasyuk: This `prop` related to [`RN.View`](https://github.com/facebook/react-native/blob/master/Libraries/Components/View/ViewPropTypes.js#L101), and there is no in [`RN.Image`](https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageProps.js#L24)

src/native-common/Image.tsx
```diff
@@ -101,20 +101,23 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
...
-                style={ this.getStyles() }
-                source={ imageSource }
-                resizeMode={ resizeMode }
+                ref={ this._onMount as any }
+                style={ this.getStyles() as RN.StyleProp<RN.ImageStyle> }
+                resizeMode={ resizeMode as any }
                 resizeMethod={ this.props.resizeMethod }
                 accessibilityLabel={ this.props.accessibilityLabel }
-                onLoad={ this.props.onLoad ? this._onLoad : undefined }
+                onLoad={ this.props.onLoad ? this._onLoad as any : undefined }
```
a-tarasyuk: Based on [`ImageProps` ](https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageProps.js#L32) `onLoad` does not accept any arguments, hm... 

```
onLoad?: ?() => void,
```

src/macos/TextInput.tsx
erictraut: Yes, I can type this as RN.Text | null.

src/native-common/Button.tsx
erictraut: Unfortunately, I can't type it as RN.Animated.View because this isn't a class definition, only a type definition (i.e. it can be used as a JSX tag but not as a variable type).

src/native-common/Image.tsx
erictraut: Yes, but it's a required (not optional) prop. If I move it out of here, TypeScript marks it as an error because I'm not specifying the source.

src/native-common/Image.tsx
erictraut: Good point. Fixed.

src/native-common/Image.tsx
erictraut: Good point. This is a long-standing bug in reactxp. I'll remove it from the RX.Image interface and the associated documentation since it clearly doesn't do anything in RN iOS.

src/native-common/Image.tsx
erictraut: Yeah, that's a bug in the type definition file. If you look at the RN.Android and RN.IOS sources, you'll see that it returns the dimensions.

src/native-common/AccessibilityUtil.ts
```diff
@@ -10,12 +10,15 @@
...
 
 import { AccessibilityUtil as CommonAccessibilityUtil, AccessibilityPlatformUtil } from '../common/AccessibilityUtil';
 
 import Types = require('../common/Types');
 
-const liveRegionMap: { [key: string]: string } = {
+type AccessibilityLiveRegionValue = 'none' | 'polite' | 'assertive';
+
+const liveRegionMap: { [key: string]: AccessibilityLiveRegionValue } = {
     [Types.AccessibilityLiveRegion.None]: 'none',
```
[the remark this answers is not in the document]
erictraut: No, it's not needed here because we're using a TypeScript-style map in this case. We couldn't use this approach in the other location because TypeScript maps support keys that are either strings or numbers, not other types.

PR #679, opened 2018-06-12
UWP announcement fixes: queuing, hiding fake element, identical subsequent announcements

zholobov:


src/native-common/RootView.tsx
```diff
@@ -125,17 +126,23 @@ abstract class BaseRootView<P extends BaseRootViewProps> extends React.Component
...
-                    accessibilityLiveRegion={ AccessibilityUtil.accessibilityLiveRegionToString(Types.AccessibilityLiveRegion.Polite) }
-                />
+                { announcerView }
             </RN.Animated.View>
         );
 
         return this.renderTopView(content);
     }
 
+    renderAnnouncerView(): JSX.Element {
```
erictraut: This should be a private or protected method and should begin with an underscore.

src/windows/RootView.tsx
```diff
@@ -55,6 +56,12 @@ class RootViewUsingStore extends RootViewBase {
             return super.renderTopView(content);
         }
     }
+
+    renderAnnouncerView(): JSX.Element {
```
erictraut: See note above. This should be protected and start with an underscore.

zholobov:
@erictraut fixed! I miss "override" keyword.

PR #680, opened 2018-06-13
On web, we should set user presence to Active only if the app has focus

RamyaSenk:
Currently when the app doesnt have focus, and the user hovers on the app -> the user presence is set to true. This is incorrect assumption. The user should be considered active or present in the app, only if the app has focus. The handleWakeUp callback gets called even when the user hovers over the app. 

Fix - Adding a callback to detect when the app has focus. And only call handle wake up when the app has focus. 

Here handle wake up is necessary when the app has focus and the user becomes idle. Then hovering over the app should set the present state to true.


src/web/UserPresence.ts
```diff
@@ -18,21 +18,24 @@ if (typeof(document) !== 'undefined') {
...
-            this._isPresent = ifvisible.now();
+            this._isPresent = this._isAppFocused = ifvisible.now();
 
             ifvisible.on('wakeup', this._handleWakeup.bind(this));
             ifvisible.on('idle', this._handleIdle.bind(this));
             ifvisible.on('focus', this._handleFocus.bind(this));
             ifvisible.on('blur', this._handleBlur.bind(this));
 
             window.addEventListener('blur', this._handleWindowBlur.bind(this));
+            window.addEventListener('focus', this._handleFocus.bind(this));
```
a-tarasyuk: nit: I think we can use the following syntax instead of explicitly use `.bind` in each callback

```typescript
window.addEventListener('focus', this._handleFocus);
```

```typescript
private _handleFocus = (): void => {
  ...
}
```

erictraut: In this case, we can't use this approach because the code is within the object's constructor. The instance variables (e.g. "_handleFocus") may not be initialized at the time the constructor runs.

src/web/UserPresence.ts
```diff
@@ -18,21 +18,24 @@ if (typeof(document) !== 'undefined') {
 
 export class UserPresence extends RX.UserPresence {
     private _isPresent: boolean;
+    private _isAppFocused: boolean;
```
erictraut: I'm concerned about adding another state variable here. The ifvisible module is already maintaining its notion of whether the app's main window is focused, and we're now maintaining a separate piece of state that tracks this. This is confusing and fragile.

src/web/UserPresence.ts
erictraut: Looking at the source for ifvisible, it looks like that module already installs a window-level "focus" and "blur" handler and uses them to report "focus" and "blur" events, which we subscribe to. So this appears to be redundant. See https://github.com/serkanyersen/ifvisible.js/blob/master/src/ifvisible.js. Ideally, we could simplify the existing logic by removing the _handleWindowBlur method and relying only on ifvisible to perform all of the work here.

mariarek:
Hi Ramya, I wanted to point you to my old PR: https://github.com/Microsoft/reactxp/pull/474
- we were hitting an issue where setting the focus would not start ifvisible's idle timer, thus the userPresence would **never** change back to false, until the user moves his mouse or performs some other action that causes ifvisible to kick its timer.

src/web/UserPresence.ts
mariarek: Hi, just wanted to point out that we're using https://www.npmjs.com/package/ifvisible (1.1.0), not https://www.npmjs.com/package/ifvisible.js - since this package was not updated and does not contain the fix for properly clearing timeouts.

So the actual source is here: https://github.com/muhammaddadu/ifvisible/blob/master/src/ifvisible.js

Also, IIRC ifvisible was not handling window.focus/window.blur as we needed, which is the reason we added our own. Cannot remember the details, but found this issue in the original repo: https://github.com/serkanyersen/ifvisible.js/issues/42

RamyaSenk: @erictraut currently the only issue i see is not being able to detect that window is out of focus. It is also used in other places like to detect app activation state. 

Also it has logic to determine that user is inactive in the app - has a timeout for mouse moves. Which is valuable. 

Do you still think we need to reimplement this?

RamyaSenk: I will reimplement ifvisibile here based on your comments in this PR. https://github.com/Microsoft/reactxp/pull/681

RamyaSenk:
Implemented the changes we discussed in this PR: 
https://github.com/Microsoft/reactxp/pull/681

In summary, I think we're talking about three changes:

1. Modify the behavior of UserPresence on web/Windows/Mac to always report False if the app's main window is not in focus (regardless of whether mouse activity is seen within the window).

2. Modify the behavior of App Activity on web/Windows/Mac to report Inactive when the app's main window is not in focus but is not minimized (or the tab is backgrounded).

3. Update the documentation for both UserPresence and AppActivityState so their intent and behavior is clear.

src/web/UserPresence.ts
RamyaSenk: I think this comment is not valid anymore. I have removed using the same callback for two events..

src/web/UserPresence.ts
RamyaSenk: Removed using ifvisible module and implemented a Utils class AppVisibilityUtils that handles this now.

src/web/UserPresence.ts
RamyaSenk: Reimplemented ivisible in our own utils class AppVisibilityUtils. Please have a look.

RamyaSenk:
@mariarek in the current implementation, i made sure it restarts idle timer when the app is focused. Let me know if you have any concerns.

src/web/UserPresence.ts
```diff
@@ -6,40 +6,36 @@
...
         } else {
             this._isPresent = false;
         }
     }
 
     isUserPresent(): boolean {
         // Handle test environment where document is not defined.
-        if (typeof(document) !== 'undefined') {
-            return ifvisible.now();
+        if (typeof (document) !== 'undefined') {
```
erictraut: Shouldn't this method always simply return this._isPresent?

src/web/UserPresence.ts
```diff
@@ -6,40 +6,36 @@
...
-        if (typeof(document) !== 'undefined') {
-            this._isPresent = ifvisible.now();
-
-            ifvisible.on('wakeup', this._handleWakeup.bind(this));
-            ifvisible.on('idle', this._handleIdle.bind(this));
-            ifvisible.on('focus', this._handleFocus.bind(this));
-            ifvisible.on('blur', this._handleBlur.bind(this));
-
-            window.addEventListener('blur', this._handleWindowBlur.bind(this));
+        if (typeof (document) !== 'undefined') {
```
erictraut: Nit: Space should be removed between "typeof" and "(document)".

src/web/UserPresence.ts
erictraut: Nit: Space should be removed after "typeof".

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+*
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT license.
+*
+* Web-specific helpers for firing focus/activity related events
+*/
+
+import SubscribableEvent from 'subscribableevent';
+
+const idleTimeInMs = 10 * 1000;
```
erictraut: Shouldn't this be 60 * 1000?

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export class AppVisibilityUtils {
+    private _isIdle = false;
+    private _timers: Array<number> = [];
```
erictraut: Why are you allocating multiple timers? Isn't a single timer sufficient?

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    hasFocusAndActive() {
+        return document.hasFocus() && !this._isIdle;
+    }
+
+    hasFocus() {
+        return document.hasFocus();
+    }
+
+    private _trackIdleStatus () {
```
erictraut: Nit: Remove space after "_trackIdleStatus".

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+        document.addEventListener('mousemove', this._wakeUpAndSetTimerForIdle);
+        document.addEventListener('keyup', this._wakeUpAndSetTimerForIdle);
+        document.addEventListener('touchstart', this._wakeUpAndSetTimerForIdle);
+        document.addEventListener('scroll', this._wakeUpAndSetTimerForIdle);
+
+        this._wakeUpAndSetTimerForIdle();
+    }
+
+    private _wakeUpAndSetTimerForIdle = () => {
+        this._timers.map(clearTimeout);
```
erictraut: See above. I don't understand why we need an array of timers. A single timer should be sufficient.

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+        this._timers.push(setTimeout(() => {
+            if (this.hasFocus()) {
+                this._onIdle();
+            }
+        }, idleTimeInMs));
+    }
+
+    private _onFocus = () => {
+        this.onFocusedEvent.fire();
+        this._wakeUpAndSetTimerForIdle();
```
erictraut: Event should be fired after wakeUpAndSetTimerForIdle. If the event handler calls back into this module, it should get the updated value for hasFocusAndActive().

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private _onBlur = () => {
+        this.onBlurredEvent.fire();
+        this._onIdle();
```
erictraut: Same here. The event should be fired only after _onIdle is called.

erictraut: By calling _onIdle, you're going to fire the "onIdle" event as well as the "onBlurred" event. Is that intentional? Wouldn't it be better to fire only the "onBlurred" event here since we're not really hitting an idle condition?

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,96 @@
+/**
+* AppVisibilityUtils.ts
```
erictraut: This is so much cleaner and simpler than the ifvisible implementation! Thanks for doing this.

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+*
```
erictraut: You can remove the ifvisible dependency in package.json.

src/web/utils/AppVisibilityUtils.ts
RamyaSenk: yes sorry i was testing it.

src/web/utils/AppVisibilityUtils.ts
RamyaSenk: Hmm i reused the logic from ifvisible. But your right, since we clear it before everytime we allocate it - i can just use one.

src/web/utils/AppVisibilityUtils.ts
RamyaSenk: Agreed, i will remove it.

src/web/utils/AppVisibilityUtils.ts
RamyaSenk: I was debating about this. I copied the way ifvisible works. 

Essentially my thinking was, there could be listeners for only wakeup and idle events. And the app is idle when not blurred. Ifivisble does the same as well - when it goes to blur, it also fires idle event. Which i think makes sense - in which case each event has correct information about the status. 

Same goes for wakeup - current i fire "wakeup" when app is focused and when there is activity after idle. 

And same for app foreground and app background. We fire blur event, when app is backgrounded as well. And focused even when app is foregrounded. So they are independent pair of events. 

What do you think?

src/web/utils/AppVisibilityUtils.ts
RamyaSenk: Yes thanks - will make that change.

src/web/utils/AppVisibilityUtils.ts
RamyaSenk: Yes i did that.

src/web/utils/AppVisibilityUtils.ts
RamyaSenk: yes i agree. No problem, thanks for pointing me to it. :)

src/web/utils/AppVisibilityUtils.ts
erictraut: OK, sounds reasonable. Just wanted to make sure it was intentional.

src/web/UserPresence.ts
RamyaSenk: Yes, i will fix that.

RamyaSenk: Done.

src/web/UserPresence.ts
RamyaSenk: Removed.

src/web/UserPresence.ts
RamyaSenk: Removed.

src/web/utils/AppVisibilityUtils.ts
RamyaSenk: Fixed.

RamyaSenk:
@erictraut made the changes your requested. Please have a look.

src/web/utils/AppVisibilityUtils.ts
```diff
@@ -0,0 +1,98 @@
...
+    
+    onFocusedEvent = new SubscribableEvent<() => void>();
+    onBlurredEvent = new SubscribableEvent<() => void>();
+    onAppForegroundedEvent = new SubscribableEvent<() => void>();
+    onAppBackgroundedEvent = new SubscribableEvent<() => void>();
+    onIdleEvent = new SubscribableEvent<() => void>();
+    onWakeUpEvent = new SubscribableEvent<() => void>();
+
+    constructor() {
+        window.addEventListener('focus', this._onFocus);
```
berickson1: I suspect that this is going to cause crashes in JS unit test frameworks that don't have a DOM since you're acting on the window and document object without checking if they exist

Based on existing comment in UserPresence.ts
`// Handle test environment where document is not defined.`

RamyaSenk: Thanks Brent. Added it as a part of this pr https://github.com/Microsoft/reactxp/pull/710/files

PR #684, opened 2018-06-14
Made ValueListener and related methods private

erictraut:
ValueListener was inadvertently being exported as a public interface from Animated. Made it private, along with several methods in the Value class that were also intended to be private.

### DIFF
--- a/src/web/Animated.tsx
+++ b/src/web/Animated.tsx
@@ -70,7 +70,7 @@ export abstract class Animation {
 
 // Interface for a component that wants to know when the value
 // of an Animated.Value changes or is about to be animated.
-export interface ValueListener {
+interface ValueListener {
     setValue(valueObject: Value, newValue: number | string): void;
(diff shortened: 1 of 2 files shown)


src/web/Animated.tsx
```diff
@@ -149,32 +149,32 @@ export class Value extends Types.AnimatedValue {
     }
 
     // Add listener for when the value gets updated.
-    addListener(listenerToAdd: ValueListener): void {
+    _addListener(listenerToAdd: ValueListener): void {
```
berickson1: Shouldn't this (and the rest) have the private modifier? They're still exposed as public on ValueListener

erictraut: That won't work because these methods on Value need to be callable from within the AnimatedComponentGenerated class. In other words, they need to have public accessibility within reactxp's implementation but not part of its public interface. Unfortunately, TypeScript doesn't have any way of expressing this.

PR #685, opened 2018-06-14
Improved docs for usage of ImageSvg extension

shameister:
The docs was not complete and little incorrect, when it comes to using it.

1) I added possibility to install via yarn

2) Change the tag syntax to the one in the original code

3) Added needed import, so the next developer has not to search for it.


erictraut:
Thanks for the contribution!

Issue #687, opened 2018-06-14
Docs: Example for importing the Network-API missing


shameister:
While trying to use the [Network-API](https://microsoft.github.io/reactxp/docs/apis/network) I am missing an example. Most important would be the line, whats the class to import.

I found [this class definition](https://github.com/Microsoft/reactxp/blob/master/src/web/Network.ts) and was able to implement my needed behavior using `RX.Network`. But the class definition found gave me the feeling I am supposed to use the more concrete implementation.

Or… am I on the wrong track?

erictraut:
I just added a sample use case to the documentation.

Note that all of the APIs in ReactXP are used within the RXPTest sample app, so that's a good place to go if you want more detailed examples of how to use them.

PR #688, opened 2018-06-14
Added example, how to use the connectivityChangedEvent

shameister:
Maybe, I have not done it the right way. [Please keep this in mind](https://github.com/Microsoft/reactxp/issues/687)


erictraut:
I added additional documentation. Your sample was more complex than necessary.

PR #690, opened 2018-06-14
Added screen reader support to FocusManager/Windows

reseul:
1. Bought the Windows implementation in sync with web with regards to hiding restricted/limited focusable controls from screen readers (prior code supported just the limitation of keyboard based navigation).
This is done by adding a way for FocusManager to override the "importantForAccessibility" property using a similar mechanism like the tabIndex one.
2. Addressed the hiding of non-keyboard focusable controls like RX.Text. To support this kind of controls I added an "accessible only" category, basically a lightweight version of FocusManager tracked control that only supports overriding "importantForAccessibility".
    - An optimization is used to prevent tracking ALL RX.Text instances in an app. This is based on the fact that hiding a node through an "importantForAccessibility==='no-hide-descendants'" hides the whole child subtree, making the tracking of those child controls unnecessary.
    - Current implementation is set up for Windows. Web could follow/adapt same path.
3. Moved some web-only state from StoredComponent class to the web derivation + some other small fixes for some regressions.

This PR is sent before a rebase, will do that tomorrow + PR feedback fixes.

### DIFF
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -40,10 +40,6 @@ export interface StoredFocusableComponent {
     restricted: boolean;
     limitedCount: number;
     limitedCountAccessible: number;
-    origTabIndex?: number;
-    origAriaHidden?: string;
-    curTabIndex?: number;
-    curAriaHidden?: boolean;
     removed?: boolean;
     callbacks?: FocusableComponentStateCallback[];
 }
--- a/src/common/utils/FocusManager.ts
+++ b/src/common/utils/FocusManager.ts
@@ -80,7 +80,7 @@ export abstract class FocusManager {
     // Whenever the focusable element is mounted, we let the application
     // know so that FocusManager could account for this element during the
     // focus restriction.
-    addFocusableComponent(component: FocusableComponentInternal, accessibleOnly?: boolean) {
+    addFocusableComponent(component: FocusableComponentInternal, accessibleOnly: boolean = false) {
         if (component.focusableComponentId) {
             return;
         }
@@ -92,7 +92,7 @@ export abstract class FocusManager {
             id: componentId,
             numericId: numericComponentId,
             component: component,
-            accessibleOnly: !!accessibleOnly,
+            accessibleOnly: accessibleOnly,
             restricted: false,
             limitedCount: 0,
             limitedCountAccessible: 0,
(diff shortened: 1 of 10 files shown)


src/common/utils/FocusManager.ts
```diff
@@ -83,7 +80,7 @@ export abstract class FocusManager {
     // Whenever the focusable element is mounted, we let the application
     // know so that FocusManager could account for this element during the
     // focus restriction.
-    addFocusableComponent(component: FocusableComponentInternal) {
+    addFocusableComponent(component: FocusableComponentInternal, accessibleOnly?: boolean) {
```
erictraut: Perhaps it would be better to define accessibleOnly as a required parameter but with a default value of "false"? It seems unintuitive for a parameter like this to be completely optional.

erictraut: Also, a comment would be useful explaining what "accessibleOnly" means here.

src/windows/Button.tsx
```diff
@@ -26,6 +28,7 @@ let FocusableAnimatedView = RNW.createFocusableComponent(RN.Animated.View);
 
 export interface ButtonContext extends ButtonContextBase {
     isRxParentAContextMenuResponder?: boolean;
+    isRxParentAFocusableInSameFMRealm?: boolean;
```
erictraut: Nit: Casing is internally inconsistent in this variable name. "Rx" is a camel-cased abbreviation but "FM" isn't. I don't have a strong feeling about this. It would be good to document what this context variable means though. I'm not familiar with the term "FMRealm".

src/windows/Button.tsx
```diff
@@ -217,14 +228,23 @@ export class Button extends ButtonBase implements React.ChildContextProvider<But
...
 
-    updateNativeTabIndex(): void {
+    getImportantForAccessibility(): string | undefined {
+        // Focus Manager may override this
+        // We force a default of YES if no property is provided, consistent with the base class
+        return AccessibilityUtil.importantForAccessibilityToString(this.props.importantForAccessibility,
+            Types.ImportantForAccessibility.Yes);
+    }
+
+    updateNativeTabIndexAndIFA(): void {
```
erictraut: What is "IFA"? I'm not familiar with that abbreviation. Could use a comment.

src/windows/Button.tsx
reseul: I think I'll change to "isRxParentAFocusableInSameFocusManager", sounds clearer.

src/windows/Button.tsx
reseul: IFA == "importantForAccessibility". I think I'll rename everything to "updateNativeAccessibilityProps".

PR #691, opened 2018-06-14
Add hideDeleteButton prop to TextInput for use in UWP apps

mschmuckler:
Adds a prop that can be set to true (default false) to hide the 'X' delete button in TextInput boxes in UWP apps.


erictraut:
Thanks for adding the documentation & test case. Looks good!

PR #692, opened 2018-06-14
Expose the ScrollView decelrationRate property

Fitzpasd:
This iOS only property is typically always set to 'fast' when you're using `snapToInterval`.  In our testing, the default value of 'normal' doesn't give a good UX on iOS, so without this property the snapToInterval functionality is less useful.


erictraut:
Rather than expose another platform-specific property (which we generally try to avoid unless there's a really compelling reason), do you think it would be reasonable to auto-set this prop to 'fast' if snapToInterval is being used?

Fitzpasd:
Ah, that's a good idea.  I'll make that change.

PR #694, opened 2018-06-14
Remove shouldRasterizeIOS from RXTests/RX.Image

a-tarasyuk:


erictraut:
Thanks for catching this.

PR #701, opened 2018-06-16
Expose clearButtonMode RX.TextInput property

reseul:
This piggybacks on existing iOS property, and the UWP implementation will follow suit.

### DIFF
--- a/docs/docs/components/textinput.md
+++ b/docs/docs/components/textinput.md
@@ -53,9 +53,6 @@ disableFullscreenUI: boolean = false; // Android-specific
 // Can text be edited by the user?
 editable: boolean = true;
 
-// UWP-only (Windows-only) prop for hiding the 'X' delete button
-hideDeleteButton: boolean = false;
-
 // iOS-only prop for controlling the keyboard appearance
 keyboardAppearance: 'default' | 'light' | 'dark';
 
--- a/docs/docs/components/textinput.md
+++ b/docs/docs/components/textinput.md
@@ -117,7 +117,7 @@ spellCheck: boolean = [value of autoCorrect];
 // iOS and Windows only property for controlling when the clear button should appear on the right side of the text view.
 // Default behavior is dependent on platform: equivalent to 'never' on iOS, and 'always' on Windows.
 // Button is hidden when text view is empty, regardless of the property value.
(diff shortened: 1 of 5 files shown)


docs/docs/components/textinput.md
```diff
@@ -117,6 +114,11 @@ secureTextEntry: boolean = false;
 // Should spell checking be applied to contents?
 spellCheck: boolean = [value of autoCorrect];
 
+// iOS and Windows only property for controlling when the clear button should appear on the right side of the text view.
+// Default behavior is dependent on platform: equivalent to 'never' on iOS, and 'always' on Windows.
```
hnager: Would it be so bad to have the default be consistent? We could explicitly set 'never' if the incoming prop is undefined in nativecommon/TextInput

reseul: We could, it's just that I'm torn. The default UWP behavior is something Windows users got accustomed to.

hnager: I don't feel strongly about it... it's just an implementation detail that a dev needs to know and think about.

docs/docs/components/textinput.md
```diff
@@ -117,6 +114,11 @@ secureTextEntry: boolean = false;
...
+// Button is hidden when text view is empty, regardless of the property value.
+clearButtonMode: 'never' | 'while-editing' | 'unless-editing' | 'always'
```
erictraut: Nit: missing a semicolon at the end.

erictraut: We could define another value 'default' and map that to the default value for each platform. Or we could just specify that an undefined value maps to the default behavior. I'm fine with either approach.

reseul: Howard would prefer the default to be 'never', aligned with the S4L design that abhors the clear buttons.
I'm more inclined to give platform some "personality" choice..
Currently it is coded as/assumes "undefined means whatever platform chooses as the default"

docs/docs/components/textinput.md
mschmuckler: I'm nearly done with the react-native-windows implementation of it. Default mode (if no value is provided) will be the behavior as it is right now, so nothing to worry about for devs who aren't aware of this prop. I've also ensured the 4 modes behave exactly like their iOS counterparts.

PR #704, opened 2018-06-18
Remove onFocus, onBlur from extended props

a-tarasyuk:


src/typings/react-native-extensions.d.ts
```diff
@@ -25,8 +25,6 @@ declare module 'react-native' {
     }
 
     interface ExtendedTextInputProps extends RN.TextInputProps {
-        onFocus?: (e: RN.NativeSyntheticEvent) => void;
```
reseul: @a-tarasyuk I'm not flagging an issue, just a potential one. Whenever you remove lines from extended typings, pls double check whether those properties have been added to RN after 0.46... 
ReactXP supports RN starting from 0.46 (another way of saying that a particular app uses that RN version...)

a-tarasyuk: Thanks, yes I always check definitions in @types/react-native

PR #707, opened 2018-06-19
Change Image onLoad callback types / Update @types/react-native

a-tarasyuk:


src/native-common/Image.tsx
```diff
@@ -100,7 +100,7 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
                 resizeMode={ resizeMode as any }
                 resizeMethod={ this.props.resizeMethod }
                 accessibilityLabel={ this.props.accessibilityLabel }
-                onLoad={ this.props.onLoad ? this._onLoad as any : undefined }
+                onLoad={ this._onLoad }
```
erictraut: This was written this way for performance reasons. We don't want to generate a bunch of extra callback traffic over the RN bridge when we're loading a large number of images. If we pass undefined, RN doesn't bother sending any message when the image loads. We should have documented this so it was clear why it was written this way. 

I like the other parts of your change - removing the "any" casts, etc.

a-tarasyuk:
@erictraut thanks for your review. I've pushed changes, you can check.

PR #710, opened 2018-06-20
Ramyasen/activestate web

RamyaSenk:
Adding inactive state in web makes it a bit inconsistent between react native platforms and electron/web. In RN active means app is foregrounded. Adding inactive for web means - foregrounded is active or inactive. And subscribing to this events will get fired way more times in web, making the app do more work to handle this case. 

Currently this brings an inconsistency which makes it not compatible with RN platforms and anyone consuming this change will have a lot of things to fix. 

Currently reverting this change.


docs/docs/apis/app.md
```diff
@@ -21,8 +21,7 @@ enum AppActivationState {
 
     // App is inactive
     // On RN mobile platforms, it is an intermediate state between when app transitions between foreground and background.
-    // On web platforms, app is inactive when the app window is not focused.
-    // On native desktop platforms, this is currently not being used.
+    // On desktop platforms, this is currently not being used.
```
berickson1: should this be desktop or web?

RamyaSenk: Desktop - we dont use it any other place other than React native mobile.

PR #716, opened 2018-06-23
Change TextInput onSelectionChange types

a-tarasyuk:


erictraut:
Nice. Thanks for doing this cleanup work. It's helping not only reactxp but also the broader react-native community.

PR #717, opened 2018-06-24
Support for writingDirection property of text style on web. 

magrotek:
Support for writingDirection property of text style on web. At this moment it is ignored, because there is no such a style property in css.

### DIFF
--- a/src/web/Text.tsx
+++ b/src/web/Text.tsx
@@ -84,14 +84,19 @@ export class Text extends TextBase {
 
         const isAriaHidden = AccessibilityUtil.isHidden(this.props.importantForAccessibility);
 
+        const textStyle = this._getStyles() as Types.TextStyle; 
+        const directionValue = textStyle && textStyle['writingDirection'];
+        const directionProperty = directionValue ? {'dir': directionValue} : {};
+
         if (this.props.selectable || typeof this.props.children !== 'string') {
             return (
                 <div
-                    style={ this._getStyles() as any }
+                    style={ textStyle as any }
                     aria-hidden={ isAriaHidden }
                     onClick={ this.props.onPress }
                     id={ this.props.id }
                     onContextMenu={ this.props.onContextMenu }
+                    {...directionProperty}
                 >
                     { this.props.children }
                 </div>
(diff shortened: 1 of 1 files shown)


erictraut:
Can you provide an explanation of what you're trying to accomplish with this change? This approach isn't consistent with reactxp's philosophy of exposing features in a cross-platform manner. I can help explore a better way of doing this once I understand the intent behind it.

erictraut:
Do you have a need to mix RTL and LTR text within the same app? If so, I'm curious why. 

ReactXP already has [APIs](https://microsoft.github.io/reactxp/docs/apis/international.html) for controlling the writing direction at the app level.

The iOS implementation of RN has support for the "writingDirection" style on Text elements, but other RN implementations (including Android) don't support this, so I don't think it makes sense to expose it as a cross-platform style attribute at this time.

magrotek:
Yeah, I am aware the we have International class in ReactXP, but for web all of its methods are not implemented yet. Is there some another API in ReactXP that I should use for RTL languages for web client?

erictraut:
Ah, I didn't realize these weren't implemented on the web. I just created a [pull request](https://github.com/Microsoft/reactxp/pull/718) that implements these methods. Please review it and give it a try. I've tested it in the RXPApp already, and it appears to work.

To test it within your app, clone the branch from the PR, build reactxp ("npm i" followed by "npm run build"). Then copy the contents of the "dist" directory to the "./node_modules/reactxp/dist" directory in your app.

If this solves your problem, I'll merge it into ReactXP and publish it shortly.

magrotek:
@erictraut, thanks for that! Yours implementation of International class for web seems to satisfy my needs. But it requires that the whole app is prepared for rtl and right alignment on its root level. I was thinking that giving possibility to apply this more locally could be very convenient. I also thought that we want eventually to make styles available in ReactXP to work on every platform.

erictraut:
I don't think there's a broad need for intermixing writing directions within an app. For now, I'd like to avoid adding this to the core. It would be easy to write a simple reactxp extension to provide this capability.

magrotek:
You are right. I haven't thought about this. Thank you for this tip!

PR #720, opened 2018-06-26
Adding Keycode for F10 in order for Shift+F10 to be handled. 

roopasp:
Fixing issue with the Shift +F10 combination that should also bring up the context menu. 
The keyCode for F10 was not added to the Event Helper hence it was getting missed.

### DIFF
--- a/src/native-common/utils/EventHelpers.ts
+++ b/src/native-common/utils/EventHelpers.ts
@@ -64,9 +64,53 @@ export class EventHelpers {
                     case 'Application':
                         keyCode = 500;
                         break;
+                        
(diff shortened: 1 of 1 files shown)


src/native-common/utils/EventHelpers.ts
```diff
@@ -65,6 +65,10 @@ export class EventHelpers {
                         keyCode = 500;
                         break;
 
+                    case 'F10':
```
erictraut: If you're going to add F10, you should also add all the other function keys.

roopasp:
Added keyCodes for F1-F12.

PR #731, opened 2018-06-28
Erictr/todosample

erictraut:
This is a major revamp of the TodoList sample app. It incorporates a bunch of techniques we've used internally. See README for a full list.

Please review to make sure I'm not releasing anything that shouldn't go external.


samples/TodoList/src/ts/services/ServiceRegistrar.ts
```diff
@@ -0,0 +1,23 @@
+/**
+* ServiceRegistrar.tsx
```
deregtd: You moved this code here, but also have it in the base appbootstrapper

erictraut: Not sure what you mean. The ServiceRegistrar is registering all services. The AppBootstrapper is starting a few of them (well, actually all of them in the current sample).

rolodex:
Has the Visual Studio sln been tested? It's failing due to SQLite plugin not being found.

erictraut:
The react-native-sqlite-storage module needs to be updated to include Windows support. I used a non-published prerelease version to develop the sample. I'm waiting for @rigdern to add the code to the public repo and publish a new version. Once he does, the TodoList sample should work fine. Adam, do you have an update on this?

rigdern:
I had to make some changes to the UWP react-native-sqlite-storage code while preparing it for the public repo. My goal is to send a react-native-sqlite-storage PR within the next couple of days.

PR #732, opened 2018-06-29
Make auto as valid resizeMode value

a-tarasyuk:


src/web/Image.tsx
```diff
@@ -375,41 +376,43 @@ export class Image extends React.Component<Types.ImageProps, ImageState> {
         this._mountedComponent = component;
     }
 
-    private _getStyles() {
-        let combinedStyles = Styles.combine([_styles.defaultContainer, this.props.style]) as any;
+    private _getStyles(): CSSProperties {
+        const { resizeMode = 'contain' } = this.props;
```
erictraut: I'm not familiar with this construct. Does this assign the value of 'contain' if 'this.props.resizeMode' is undefined? If so, your change doesn't deal with the case where the caller specified some unsupported value for the resizeMode prop. The previous code properly fell back to the default of 'contain'. Your logic simply passes is through unmodified.

Incidentally, the documentation indicates that the default value of this prop is 'stretch', which doesn't match the code. It's probably best to change the documentation to say that the default is 'contain' rather than changing the code at this point, in case app code already depends on this behavior.

a-tarasyuk: @erictraut thanks for your comments.

> If so, your change doesn't deal with the case where the caller specified some unsupported value for the resizeMode prop.

I think we can use propTypes to disallow unsupported values instead of adding this logic explicit to code

```typescript
static propTypes = {
    resizeMode: PropTypes.oneOf(['auto', 'contain', 'cover', 'stretch', 'repeat'])
};
```

What do you think about it?

> Are you planning to fix that as well?
Yes, I'll make new PR for that, after discussion. In the PR just wanted to fix the current bug, in order to have small PR to review. Does that make sense?

erictraut: We don't use PropTypes for enforcement anywhere else in ReactXP, so I'm hesitant to use this approach in this one case. (We tend not to use PropTypes because we believe strongly in static type checking versus runtime checks.) I guess that my comment above doesn't affect users of TypeScript. 

In this case, I'd prefer to stick with the previous approach of falling back to the default if the user provides an unsupported input value.

a-tarasyuk: >  I guess that my comment above doesn't affect users of TypeScript.

You are right. PropTypes has to check it in JS version. Anyway, I will add this logic explicitly in code.

a-tarasyuk:
> Incidentally, the documentation indicates that the default value of this prop is 'stretch', which doesn't match the code.

It was the previous behavior.

@erictraut Pushed changes + Updated docs.

PR #745, opened 2018-07-10
WebView.onMessage receives additional props which are not defined in Types.WebViewMessageEvent

a-tarasyuk:


src/native-common/WebView.tsx
```diff
@@ -1,19 +1,18 @@
...
+ * Licensed under the MIT license.
+ *
+ * A control that allows the display of an independent web page.
+ */
+
 import React = require('react');
 import RN = require('react-native');
 
-import RX = require('../common/Interfaces');
+import * as Types from '../common/Types';
```
erictraut: For consistency, we should always use "RX.Types" rather than importing "Types" separately. This file uses a mixture of the two currently.

a-tarasyuk: @erictraut Thanks, I've checked https://github.com/Microsoft/reactxp/blob/master/src/common/Interfaces.ts#L18 and you are right this file contains re-export of Types. 

Also, there are a lot of files which containt RX and Types, for instance https://github.com/Microsoft/reactxp/blob/40ef7634b5886672a4edd059218b51bd579f7c23/src/web/Input.ts, https://github.com/Microsoft/reactxp/blob/1ccb82370df0b701a9e8eb8b735b2c5ee3cc4cb9/src/native-common/Alert.ts, https://github.com/Microsoft/reactxp/blob/b51c6d101f652c39e72210f1cdae5dee66fb5a72/src/native-common/Platform.ts. Should these files be changed to use only one RX import?

erictraut: If they're importing common/Types, they yes, ideally they should be using RX.Types instead. It's not a big deal, but if you happen to be changing a file anyway...

Issue #747, opened 2018-07-10
VirtualListView should pass testId through to RX.ScrollView


luisnaranjo733:
@erictraut Added support for the RN testID prop a few weeks ago on all of the XP views.

It looks like the VirtualListView got missed in that PR. Notice how the VLV doesn't pass the testId prop to the RX.Scrollview [here](https://github.com/Microsoft/reactxp/blob/master/extensions/virtuallistview/src/VirtualListView.tsx#L1142).

- [Example](https://imgur.com/caJL1WO) that renders RX.View and VirtualListView and tries to pass testId to each
- [Image](https://imgur.com/R2uWAkl) of what that code sample looks like when rendered
- XCode layout inspector [screenshot](https://imgur.com/b7x0JHd) that shows that the testId was successfully passed to the native view from RX.Text
- XCode layout inspector [screenshot](https://imgur.com/cZyoZY9) that shows that the testId was _not_ successfully passed to the native view from VirtualListView

luisnaranjo733:
By the way, I haven't checked any of the other extensions to see if they were also missed. I imagine they are also not passing this prop through yet.

erictraut:
Yeah, I didn't make this change to any of the extensions. Should be easy to add if someone wants to submit a PR. Or I can do it when I publish the next release.

luisnaranjo733:
Cool. Thanks!

PR #748, opened 2018-07-10
Pass testId from VirtualListView to ScrollView

luisnaranjo733:
I believe this will *almost* fix issue #747.

It seemed like a small change, so I figured I'd go ahead and make a PR myself.

I tested by building the root xp project and the extension, and then dropping the transpiled files from the extensions dir in dist/ into the vlv package in my project's node_modules.

 ~~I used XCode's layout inspector to confirm that the testId is being propagated from XP to RN to Native.~~
What I saw was the testId being propagated correctly when I use RX.ScrollView directly, *not* when I use VirtualListView. It looks like this change either isn't enough, or I'm going about testing this wrong.

### DIFF
--- a/extensions/virtuallistview/src/VirtualListView.tsx
+++ b/extensions/virtuallistview/src/VirtualListView.tsx
@@ -1141,6 +1141,7 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
         return (
             <RX.ScrollView
                 ref={ _scrollViewRef }
+                testId={this.props.testId}
                 onLayout={ this._onLayoutContainer }
                 onScroll={ this._onScroll }
                 keyboardDismissMode={ this.props.keyboardDismissMode }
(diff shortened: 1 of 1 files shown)


luisnaranjo733:
Hmm actually I was wrong when I thought this fix was enough.

Consider the following:
```
        <RX.View style={styles.scrollViewContainer}>
          <RX.ScrollView style={styles.scrollView1} snapToInterval={50} scrollsToTop={true} testId={"xpscrollviewtestid"}>
            <RX.View style={styles.listContainer}>{numberItems}</RX.View>
          </RX.ScrollView>
        </RX.View>

        <VirtualListView
          testId={"vlvTestId"}
          itemList={this.state.items}
          renderItem={this._renderItem}
          animateChanges={true}
          skipRenderIfItemUnchanged={true}
        />
```

When I inspect this rendered code in XCode, I can see "xpscrollviewtestid" in the native RCTScrollView, indicating that testId was succesfully passed to RN.

However, I do not see "vlvTestId" in the RCTScrollView from VirtualizedListView. I wonder if manually copying the transpiled files into my node_modules/reactxp-virtuallistview/dist/ folder is not enough to get the changes to take into effect. 

Or am I missing something? Is there something else that needs to happen in addition to this code change?

erictraut:
Thanks for the change. The documentation also needs to be updated, but I can handle that separately.

PR #749, opened 2018-07-11
Set size attribute of input HTML element to 1.

MiguelCa-MSFT:
The input HTML element has size 20 by default. This means that it has to allocate at least 20 characters, which makes the input to have a minimum width that can fit those characters (the final minimum width in pixels depends on the browser and the font. More info: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input). This behavior differs from mobile implementation that doesn't have a minimum width by default.

Since size attribute has to be greater than 0, setting it to 1 is the closest we can get to have same behavior across platforms.


MiguelCa-MSFT:
Here there is an example of the behavior we are trying to fix with this change:
https://codepen.io/anon/pen/vaYQJw?editors=1100

PR #752, opened 2018-07-16
s#1528727 Accessibility fix for Clipboard

krbartov:


krbartov:
My changes resolves Accessibility bug, which was causing focus loss after copying text using textarea.

src/web/Clipboard.ts
```diff
@@ -15,7 +16,7 @@ export class Clipboard extends RX.Clipboard {
         let node = Clipboard._createInvisibleNode();
         // Replace carriage return /r with /r/n, so that pasting outside browser environment
         // (eg in a native app) preserves this new line
-        text = text.replace(/\r/g, '\r\n');
+        text = _.escape(text).replace(/\r/g, '<br/>');
```
mshoho: That regexp is not sufficient for all platforms. Perhaps it should be something like `/\n\r|\r\n|\n|\r/g`. Also the comment above it is obsolete after this change.

src/web/Clipboard.ts
```diff
@@ -6,6 +6,7 @@
 *
 * Web-specific implementation of the cross-platform Clipboard abstraction.
 */
+import _ = require('lodash');
```
a-tarasyuk: You can add `escape` to **lodashMini** _(src/common/lodashMini.ts)_ and import it from **lodashMini** instead of import entire **lodash**., or import only `escape`, like so

```typesscript
import escape = require('lodash/escape')
```
or 
```typesscript
import { escape } from 'common/lodashMini';
```

src/web/Clipboard.ts
```diff
@@ -54,12 +55,12 @@ export class Clipboard extends RX.Clipboard {
         return node;
     }
 
-    private static _copyNode(node: HTMLTextAreaElement) {
+    private static _copyNode(node: HTMLSpanElement) {
         const selection = getSelection();
+        var range = document.createRange();
```
RamyaSenk: You will have to remove selection before creating new one.. 

const selection = getSelection();
selection.removeAllRanges();

const range = document.createRange();
range.selectNodeContents(node);
selection.addRange(range);

document.execCommand('copy');
selection.removeAllRanges();

src/web/Clipboard.ts
krbartov: Thanks for your comment, Marat. 
I have updated the regexp so it works for all three cases (cr, lf, and crlf). Haven't heard of lfcr, so I did not added this case.

mshoho: Well, theoretically: https://en.wikipedia.org/wiki/Newline. Anyway yours should work for our platforms.

src/web/Clipboard.ts
krbartov: Thanks for your comment. I have updated PR accordingly.

src/web/Clipboard.ts
```diff
@@ -6,16 +6,17 @@
...
 import SyncTasks = require('synctasks');
 
 export class Clipboard extends RX.Clipboard {
     public setText(text: string) {
         let node = Clipboard._createInvisibleNode();
-        // Replace carriage return /r with /r/n, so that pasting outside browser environment
+        // Replace carriage returns or newlines with br, so that pasting outside browser environment
         // (eg in a native app) preserves this new line
-        text = text.replace(/\r/g, '\r\n');
+        text = escape(text).replace(/\r\n?|\n/g, '<br />');
```
catalinmiron: isn't `/\r\n?|\n/g` equal with `/\r|\n/g`?

krbartov: I am not regexp expert, but IMO it is not. I think that /\r|\n/g won't cover CLRF case, but CL or RF separately.

catalinmiron: Totally agree. Already talked with @mshoho.

PR #760, opened 2018-07-23
Use semver in reactxp dependencies

berickson1:
This prevents consumers from accidentally importing and bundling two versions of the same library

For example, if I have version 1.0.0 in reactxp and 1.0.1 in my app, I would bundle both versions.  If reactxp has semver, it allows 1.0.1 to be bundled once since it's compatible with the referenced version in reactxp

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,10 +9,10 @@
     "tslint": "tslint --project tsconfig.json -r tslint.json -r ./node_modules/tslint-microsoft-contrib --fix || true"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


berickson1:
I believe most people go with that approach. Otherwise RX prevents consumers from choosing/locking to specific versions of packages. I actually ran into a problem with this in the client repo

berickson1:
Ok, updated. I'll leave the @types modules in place since I'm not sure what the impact (if any) will be to move them to dev deps

PR #761, opened 2018-07-26
Update usages of findDOMNode to indicate that it returns null

berickson1:
Fix a bug in web animations that results in a crash due to this


src/web/Animated.tsx
```diff
@@ -395,14 +395,20 @@ function createAnimatedComponent<PropsType extends Types.CommonProps>(Component:
 
             let attrib = this._findAnimatedAttributeByValue(this._animatedAttributes, valueObject);
             if (attrib) {
-                let cssValue = this._generateCssAttributeValue(attrib, valueObject, valueObject._getValue());
-                (this._getDomNode().style as any)[attrib] = cssValue;
+                const domNode = this._getDomNode();
+                if (domNode) {
```
berickson1: The crash I fixed is due to missing null check here

PR #776, opened 2018-08-14
Brent pointed out a problem in my last change, where first-runs of no…

deregtd:
…n-cached images would throw an error as well as a load.  Reworked it to only show the img tag when the blob url is loaded, if there's an onload prop.

### DIFF
--- a/src/web/Image.tsx
+++ b/src/web/Image.tsx
@@ -229,10 +229,12 @@ export class Image extends React.Component<Types.ImageProps, ImageState> {
         // Only make the xhr request if headers are specified and there was no cache hit.
         const performXhrRequest = !!props.headers && !cachedXhrBlobUrl;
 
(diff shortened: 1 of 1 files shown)


src/web/Image.tsx
```diff
@@ -250,14 +252,15 @@ export class Image extends React.Component<Types.ImageProps, ImageState> {
             return;
         }
 
-        this.setState({
-            displayUrl: URL.createObjectURL(blob)
-        });
-
         // Save the newly fetched xhr blob url in the cache.
         XhrBlobUrlCache.insert(this.props.source, this.state.displayUrl);
```
berickson1: I think this change breaks the XhrBlobUrlCache. I suspect the cache was relying on synchronous setStates, but at this point in code this.state.displayUrl will be undefined.  I think you need to create the objectURL above , stuff it into cache then call setState

It's also possible that the cache behaviour is already busted in some cases

PR #779, opened 2018-08-17
Use localhost:8080 so /etc/hosts changes are unnecessary

mikehardy:
I really appreciate the examples, but it's best to run things like this on localhost:8080 so no privileged access (edits to /etc/hosts) is required under normal development conditions


erictraut:
Yes, that's a good suggestion. The example I derived this from required the use of an SSL cert, and the instructions provided a way to create a self-signed cert. When I removed this requirement, I should have also switched to localhost. Thanks for the contribution.

PR #780, opened 2018-08-17
Expose the 'collapsable' prop on View

Fitzpasd:
Android removes View instances as a performance optimization in some scenarios, e.g. if they are only used for layout and don't draw anything.  If you have code that wants to get accurate layout information for one of those instances, then it requires setting that property to false.

The code already passes all props down, so this change just exposes it in the types and docs.

### DIFF
--- a/docs/docs/components/view.md
+++ b/docs/docs/components/view.md
@@ -50,6 +50,10 @@ ariaRoleDescription?: string = undefined; // Web only
 // Block touches for this component and all of its children
 blockPointerEvents: boolean = false; // iOS and Android only
 
+// Set to false to prevent the optimization that may cause this 
+// view to be removed from the hierarchy if it doesn't draw anything.
+collapsable: boolean = true; // Android only
+
 // Disable default opacity animation on touch on views that have
 // onPress handlers
 disableTouchOpacityAnimation: boolean = false;  // iOS and Android only
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -618,6 +618,7 @@ export interface ViewPropsShared extends CommonProps, CommonAccessibilityProps {
     blockPointerEvents?: boolean; // Native-only prop for disabling touches on self and all child views
     shouldRasterizeIOS?: boolean; // iOS-only prop, if view should be rendered as a bitmap before compositing
     viewLayerTypeAndroid?: ViewLayerType; // Android only property
+    collapsable?: boolean; // Android only prop
 
     restrictFocusWithin?: boolean; // Web-only, during the keyboard navigation, the focus will not go outside this view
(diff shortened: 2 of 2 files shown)


docs/docs/components/view.md
```diff
@@ -50,6 +50,10 @@ ariaRoleDescription?: string = undefined; // Web only
 // Block touches for this component and all of its children
 blockPointerEvents: boolean = false; // iOS and Android only
 
+// Set to false to prevent the optimization that may cause this 
+// view to be removed from the hierarchy if it doesn't draw anything.
+collapsable: boolean = true; // Android only
```
hnager: @reseul you may have some thoughts on exposing this (or maybe not, but I know we've talked about this prop in the past)

erictraut:
I'd prefer not to expose this in ReactXP. It's specific to a single platform, and it's a hack to begin with.  

Can you provide a scenario where it's necessary to expose those? It seems to me that collapsibility should be an internal implementation detail — something that callers should not depend upon or care about. Are there bugs in the implementation that force you to override it in some circumstances?

hnager:
The primary scenario, that we’ve run in to, is dealing with refs that have their associated views optimized out. Passing that ref to something like measureLayoutRelativeToWindow is problematic because there is no view that maps to the ref. There are (hack) workarounds that you can apply to prevent those views from getting optimized out. Personally I think that RX should do a better job at figuring out when views should stick around vs exposing this prop. There may also be cases where an overflow: hidden view is optimized out which can cause child views to be visibile when they would otherwise be clipped by the parent.

Fitzpasd:
I have some alternatives, so I'm fine with not exposing this if people have already thought this through.

I'm writing a component that wraps any other component and at some point will measure itself relative to the window.  Measuring a collapsed component returns undefined. 

On Android, the component itself will be collapsed since it doesn't draw anything, and it's also possible that the root child component will be collapsed.  So I need to prevent one of them being collapsed.  I could:
1. Expose the prop in RXP 
2. Add an `onLayout` prop to the View, which also prevents it being collapse. 
3. Use `cloneElement` to clone the child and set the collapsable prop on it.  This is probably what I'll do if we don't want to take this change.

erictraut:
These all sound like workarounds for what I'd argue is a bug in RN. If a ref is specified, then it's not safe for RN's implementation to remove it. I recommend filing a bug against RN — or better yet, submitting a PR that addresses it — rather than providing a hacky workaround at the RX layer. 

If you really need to use a workaround, onLayout is already exposed.

reseul:
@Fitzpasd Transparent background...

erictraut:
Yeah, transparent background is another good workaround.

PR #785, opened 2018-08-24
Expose selection from RN.Text

gmangov:
In order to be able to do partial select/copy/paste, we'll expose the selected text and event from RNW. We need to expose this via ReactXP so that the application code can use them.

### DIFF
--- a/src/web/Text.tsx
+++ b/src/web/Text.tsx
@@ -182,7 +182,7 @@ export class Text extends TextBase {
     }
 
     getSelectedText(): string {
-        return ''; // Needs implementation (requires support from RN).
+        return ''; // Not implemented yet.
     }
 }
 
--- a/src/web/Text.tsx
+++ b/src/web/Text.tsx
@@ -180,6 +180,10 @@ export class Text extends TextBase {
             }
         }
     }
+
+    getSelectedText(): string {
+        return ''; // Needs implementation (requires support from RN).
+    }
 }
 
 export default Text;
--- a/docs/docs/components/text.md
+++ b/docs/docs/components/text.md
@@ -111,6 +111,9 @@ requestFocus(): void;
 
 // Blurs the component.
 blur(): void;
+
+// When selection is enabled, retrieves the selected text.
+getSelectedText(): string; // Windows only
 ```
 
 ## Sample Usage
--- a/samples/RXPTest/src/Tests/TextTest.tsx
+++ b/samples/RXPTest/src/Tests/TextTest.tsx
@@ -24,7 +24,7 @@ const _styles = {
         color: CommonStyles.explainTextColor
     }),
     resultContainer: RX.Styles.createViewStyle({
-        alignSelf: 'stretch',
+        alignSelf: 'stretch',        
         marginHorizontal: 36,
         marginVertical: 12
     }),
@@ -75,21 +75,41 @@ const _styles = {
         fontSize: 16,
         lineHeight: 48,
         textDecorationLine: 'underline'
+    }),
+    selectTextContainer: RX.Styles.createViewStyle({
+        margin: 12,
+        flexDirection: 'row',
+        alignItems: 'center'
+    }),
+    testSelectText: RX.Styles.createTextStyle({
+        fontSize: CommonStyles.generalFontSize
+    }),
+    selectTextButton: RX.Styles.createButtonStyle({
+        backgroundColor: '#ddd',
+        borderWidth: 1,
+        margin: 20,
+        padding: 12,
+        borderRadius: 8,
+        borderColor: 'black'
     })
 };
 
 interface TextViewState {
     test5ColorIndex: number;
+    selectedText: string;
 }
 
 const _dynamicColors = ['black', 'red', 'green', 'blue'];
 
 class TextView extends RX.Component<RX.CommonProps, TextViewState> {
+    private _selectionText: RX.Text | undefined;
+
     constructor(props: RX.CommonProps) {
         super(props);
 
         this.state = {
-            test5ColorIndex: 0
+            test5ColorIndex: 0,
+            selectedText: ''
         };
     }
 
@@ -283,6 +303,30 @@ class TextView extends RX.Component<RX.CommonProps, TextViewState> {
                         { 'Text with a red shadow.' }
                     </RX.Text>
                 </RX.View>
+
+                <RX.View style={ _styles.selectTextContainer }>
+                    <RX.Text 
+                        style={ _styles.testSelectText } 
+                        selectable={ true }
+                        ref={ (comp: any) => { this._selectionText = comp; } }
+                        >
+                        { 'Select from this text.' }
+                    </RX.Text>
+                    <RX.Button
+                        style = { _styles.selectTextButton }
+                        onPress={ this._onPressCopySelectedText }
+                    >
+                        <RX.Text>
+                            { 'Copy selected.' }
+                        </RX.Text>
+                    </RX.Button>
+                    <RX.Text style={ _styles.testSelectText }>
+                        { 'Copied selected text:' }
+                    </RX.Text>
+                    <RX.Text style={ _styles.testSelectText }>
+                        { this.state.selectedText }
+                    </RX.Text>
+                </RX.View>
             </RX.View>
         );
     }
@@ -290,6 +334,15 @@ class TextView extends RX.Component<RX.CommonProps, TextViewState> {
     private _onPressTest5 = () => {
         let newIndex = (this.state.test5ColorIndex + 1) % _dynamicColors.length;
         this.setState({ test5ColorIndex: newIndex });
+    }
+
+    private _onPressCopySelectedText = () => {
+        if (this._selectionText) {
+            var selectedText: string = '';
+            // TODO Enable when ReactXP dependency is updated.
+            // selectedText = this._selectionText.getSelectedText();
+            this.setState({ selectedText: selectedText });
+        }
     }
 }
 
--- a/src/common/Interfaces.ts
+++ b/src/common/Interfaces.ts
@@ -253,6 +253,7 @@ export abstract class Text extends React.Component<Types.TextProps> implements F
     abstract focus(): void;
     abstract requestFocus(): void;
     abstract blur(): void;
+    abstract getSelectedText(): string;
 }
 
 export abstract class TextInput extends React.Component<Types.TextInputProps> implements FocusableComponent {
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -569,7 +569,6 @@ export interface AnimatedImageProps extends ImagePropsShared {
 // | example     |
 export interface TextPropsShared extends CommonProps {
     selectable?: boolean;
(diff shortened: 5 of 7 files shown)


gmangov:
@erictraut is added to the review. #Closed

erictraut:
Ah, I missed the fact that this prop is for RX.Text rather than RX.TextInput. RX.Text is not selectable by default. Can you provide a concrete example of where you need to report the selection range for an RX.Text field? 

This functionality isn't available on the web, so that makes the bar very high for including it in a cross-platform interface.

If we do end up including an onSelectionChange callback for RX.Text, it should be defined with the same signature as the onSelectionChange callback for RX.TextInput.

gmangov:
@erictraut We need this functionality because we want to provide partial select/copy/paste from a message. The select occurs in RNW RichTextBlock control and these needs to be exposed. The web provides for Window.getSelection() that is used for this purpose but this is not available for iOS/Android/UWP etc where selection occurs in the native control. As far as the API, unlike the TextInput that contains plain text, the content of the RichTextBlock and other similar controls is a document consisting of runs, spans, images, etc. that is built by sequential calls for styling and embedded elements. At that level there is no information that will provide for mapping of selection start/end back to original message that resulted in the document rendered. So, we need to build the string and return it back with the event (similar to what Window.getSelection() does).

erictraut:
OK, thanks for the explanation. 

The proposed solution isn't sitting well with me because it's specific to Windows. It doesn't appear to apply to iOS or Android, and a web implementation would need to take a different approach. 

My intuition is that there has to be a better way to expose this functionality. I haven't had time to think through the options though. Let's do some brainstorming in this forum to see if we can collectively come up with a better approach.

gmangov:
@erictraut Few more clarifications. We need this one in order to allow for users to select/copy/paste partial messages (only on RNW for now). We have multiple levels of embedded <RX.Text> tags (with views, etc.). On RNW only the top <RX.Text> results in a native control (RichTextBlock), the rest of the calls just add elements to the control's document (and the control has the ability to render formatted text, embedded images, etc). The control supports content selection (this is what is being exposed here) and all the information need for generating text representation of the selection is already available in the native control, so the text can be generated and sent in the selection event. On iOS, the RX.Text is rendered with RCTTextView(UIView) with embedded <RX.Text, etc. UIView children that are separate controls but conceptually it is the same since if the functionality is implemented on iOS, it will be needed to support selecting child views of a RN.Text, building a text representation of these and firing an event when the selection occurs with the text representation of the selection.

deregtd:
Do you actually want events for the selectionchange, or just the ability to get the current selection, across all platforms?  With a getter for current selection, you could just fire a timer for checking when it changes, if you pass the this.props.onSelectionChanged on web.  We do the same thing for onLayout on View on web, as ugly as that is.

gmangov:
@deregtd Since the intended functionality is "on demand" a getter may be ok (when user initiates copy/paste). But AFAIK event-based approach is superior to polling with timer and a getter.

erictraut:
Thanks for the details, George. I think I understand the problem you're trying to solve. Let's think through the possible solutions.
1. Expose a callback prop on RX.Text that is called whenever the selection changes (the approach you've implemented in your PR)
2. Expose an async method on RX.Text that can be called to retrieve the current text selection if there is one
3. Expose an async method in RX.UserInterface that retrieves the current selection if there is one; we'd probably need to include an optional rootViewId to handle multi-window cases
4. Recognize that this problem is probably unique to Skype and punt on the problem in ReactXP; Skype could solve the problem by creating an internal module with the desired behavior, and you could wrap messages within an instance of this module on the UWP client

My concerns with your proposed approach are:
1. The solution doesn't apply to any platform other than UWP, so it goes against of ReactXP's cross-platform philosophy
2. It introduces an interface inconsistency by exposing a prop that has the same name as a prop on RX.TextInput but with a different parameters

David's suggestion covers options 2 and 3. Option 2 is easier to implement but is also specific to UWP. Option 3 could presumably be implemented on web as well.

At this point, I think my preference is option 4. This is something you could do with minimal work on your side, and you wouldn't be blocked on a new major version of ReactXP being published. 

Thoughts?

gmangov:
@erictraut Tnx! I am not entirely clear on few things: 1. Why the solution doesn't apply to any other platform? On iOS/Android and web we also want to provide functionality where the content of RX.Text can be partially selected. In fact RX.Text exposes "selectable" property and RN.Text in addition to "selectable" exposes "selectionColor" property. There is not much value in allowing selection if it can't be copied. Bug the default Ctrl+C on UWP doesn't take into account emoticons, etc. as mentioned above. So to me it looks like this "partial selection of RX.Text content" feature is not just UWP but also applies to other platforms (and it already works on web since it has support from the browser)
2. Regarding 3. my concern is that it assumes that in the window it can be only one selection at a time. It will do for now but I am not sure that this is the general case. 
3. Regarding 4. I am not clear on exactly how this will be wired to RNW. At RNW level selection occurs in the native control (in this case RichTextBlock). We have some modules in Skype that implement specific behavior e.g. animated sprites. Is this something similar? If so, it seems like a kind of complicated solution since it will require all "selectable" RX.Texts to be replaced with some "selectable element" and it will require moving the entire logic of the RNW.Text into the skype module (if I understand correctly). Why do you think it is going to require minimal work?
4. As far as your second concern, I completely agree. In fact RNW already has same event already with different params so I just followed suit. But I guess this can be addressed by using a different name.
How does that sound?


---
In reply to: [417506091](https://github.com/Microsoft/reactxp/pull/785#issuecomment-417506091) [](ancestors = 417506091)

gmangov:
@erictraut I looked into RX.UserInterface and that seams like feasible solution to me. It will address the issue with the difference between how web and other clients treat selection and will provide for an easy integration. Unless I get a better understanding about 4., 3. is what I would prefer.....

---
In reply to: [417510761](https://github.com/Microsoft/reactxp/pull/785#issuecomment-417510761) [](ancestors = 417510761,417506091)

erictraut:
You said "On iOS/Android and web we also want to provide functionality where the content of RX.Text can be partially selected." Does that mean that you're committing to make this change in the iOS and Android source base and work with FB to get it integrated into a future version of RN? If not, you should assume it won't happen.

You said "There is not much value in allowing selection if it can't be copied." Most (all?) platform environments already provide system-level functionality for copying selected text onto the system clipboard. In every situation we've seen so far, this was sufficient. For your use case, I don't think this approach is sufficient.

I agree that my second concern can be eliminated by coming up with a different name for the prop, but that still leaves my first concern. I don't see any way to implement this on the web. And unless you're committed to making it happening (and getting FB to agree with the approach), I don't expect that it will ever happen on iOS or Android.

I'll follow up with you privately on option 4, which is still my preference, and I think it's the easiest way for you to get this functionality into the Skype client as soon as possible.

gmangov:
@erictraut Even if now for us this functionality is 1st priority on UWP, there is no reason of why similar functionality wouldn't be needed for iOS/Android/Web. The React Native documentation says that for RN.Text control "selectable property lets the user select text, to use the native copy and paste functionality." but since complex text renders to view with multiple children on these platforms they will experience the same issue as UWP so if functionality is implemented there since the system provided copy/paste is not going to take into account the tree but will only get the text from a single native control. So when the time comes, we'll work into getting this functionality to ReactNative. But more than that - even now, ReactXP includes conditionals for functionality available on one or two platforms only and that eventually will be implemented in the future. So, if the functionality makes sense (like in this case) I see no reason of it not being put in place.
When you say "I don't see any way to implement this one web" I read this as "implement it at the RX.Text level", is this correct. Because if we go with option 3. that is something that web supports and that also is doable with selection events from RN.Text.
Please reach to me regarding 4. since that is not clear to me and we need to finalize on this since it is becoming a blocker for us, tnx!  

---
In reply to: [417981756](https://github.com/Microsoft/reactxp/pull/785#issuecomment-417981756) [](ancestors = 417981756)

gmangov:
@erictraut I made changes along the discussion we had, please review, feedback, tnx!

src/web/Text.tsx
```diff
@@ -180,6 +180,10 @@ export class Text extends TextBase {
             }
         }
     }
+
+    getSelectedText(): string {
+        return ''; // Needs implementation (requires support from RN).
```
erictraut: I presume this can be implemented on the web. In any event, the comment is misleading.

gmangov:
@erictraut Updated text.md and added test to RXPTest

deregtd:
If you want to add web support, I dug up some ancient code of mine that works across browsers to get the selection range:

```
function GetHighlightedObjs()
{
	if (window.getSelection)
	{
		//Firefox + Opera + Safari
		var txt = window.getSelection();

		//Pull all needed info from the selection DOM object
		var startc = txt.anchorNode;
		var endc = txt.focusNode;
		var sels = txt.anchorOffset;
		var sele = txt.focusOffset;

		//Make sure it's inside reader and not empty
		var valid = false;
		if (!txt.isCollapsed)
		{
			var test = startc;
			while (test)
			{
				if (test.id == "reader") valid = true;
				test = test.parentNode;
			}
		}
		if (!valid)
			return null;

		//Make sure it wasn't a reverse selection
		var swap = false;
		if (startc == endc)
		{
			if (sels > sele) swap = true;
		}
		else
		{
			var test = startc;
			while ((test) && (test != endc)) test = GetNextRawNode(test);
			if (!test) swap = true;
		}

		if (swap)
		{
			//Swap start and finish nodes + offsets
			test = startc;
			startc = endc;
			endc = test;

			test = sels;
			sels = sele;
			sele = test;
		}

		return new Array(startc, endc, sels, sele);
	}
	else if (document.selection)
	{
		//IE

		//ensure not empty
		if (document.selection.createRange().text == "")
			return null;

		//get start node
		var sels = document.selection.createRange();
		sels.collapse(true);
		var snode = sels.parentElement();

		//Make sure it's inside reader and not empty
		var valid = false;
		var test = snode;
		while (test)
		{
			if (test.id == "reader") valid = true;
			test = test.parentNode;
		}
		if (!valid)
			return null;

		//calculate start offset
		var soffset = -1;
		while (sels.parentElement() == snode)
		{
			soffset++;
			sels.moveStart("character", -1);
		}

		//get end node
		var sele = document.selection.createRange();
		sele.collapse(false);
		var enode = sele.parentElement();

		//calculate end offset
		var eoffset = -1;
		while ((sele.parentElement() == enode) && (sele.htmlText.indexOf("DIV>") < 1) && ((sele.htmlText.length < 5) || (sele.htmlText.indexOf("</P>") < sele.htmlText.length - 4)))
		{
			eoffset++;
			sele.moveStart("character", -1);
		}

		return new Array(snode, enode, soffset, eoffset);
	}
}

function GetNextRawNode(snode)
{
	var inode = snode;

	//First check children
	if (inode.hasChildNodes())
		return inode.childNodes[0];

	//Then check siblings
	if (inode.nextSibling)
		return inode.nextSibling;

	//Walk parents until we find one with a sibling
	while (inode.parentNode)
	{
		inode = inode.parentNode;

		if (inode.nextSibling)
			return inode.nextSibling;
	}

	return null;
}
```

PR #792, opened 2018-09-02
Stopped using the custom react-native-svg.d.ts, since react-native-sv…

deregtd:
…g is distributed with a good d.ts file automatically these days.  Updated Types and SvgRect/Path to use a bunch more supported fields on all platforms.  Lots more to add down the road if we want.


extensions/imagesvg/src/common/Types.ts
```diff
@@ -12,16 +12,40 @@ export interface ImageSvgStyle extends RXTypes.ViewStyle {
 }
 export declare type ImageSvgStyleRuleSet = RXTypes.StyleRuleSet<ImageSvgStyle>;
 
-export interface SvgCommonProps {
-    key?: string | number;
+type NumberProp = string | number;
```
erictraut: Why are you supporting strings here? No where else in ReactXP are strings supported for numeric props. I'd prefer to stay consistent.

extensions/imagesvg/src/common/Types.ts
```diff
@@ -12,16 +12,40 @@ export interface ImageSvgStyle extends RXTypes.ViewStyle {
...
+    // These typings are incompatible between web and react-native-svg.  Requires further understanding...
+    // strokeDasharray?: ReadonlyArray<number> | string,
+    strokeDashoffset?: NumberProp;
+    strokeLinecap?: Linecap;
+    strokeLinejoin?: Linejoin;
+    strokeMiterlimit?: NumberProp;
+}
+
+export interface ImageSvgProps extends RXTypes.CommonStyledProps<ImageSvgStyleRuleSet> {
+    // Wacky subset of available props that only does anything on web..?
```
erictraut: Not sure what you mean by "wacky subset...that only does anything on web". These were implemented previously for both web and native. Is the issue that they're defined for images rather than paths?

extensions/imagesvg/src/common/Types.ts
```diff
@@ -31,12 +55,18 @@ export interface ImageSvgProps extends SvgCommonProps, RXTypes.CommonStyledProps
...
     d?: string;
 }
-export interface SvgRectProps extends SvgCommonProps {
-    width: number;
-    height: number;
-    x: number;
-    y: number;
+
+export interface SvgRectProps extends CommonPathProps {
+    x?: NumberProp;
```
erictraut: See comment above about NumberProp. I'd prefer to keep these as "number".

extensions/imagesvg/src/common/Types.ts
```diff
@@ -12,16 +12,40 @@ export interface ImageSvgStyle extends RXTypes.ViewStyle {
...
+    fillRule?: FillRule;
+}
+  
+export interface StrokeProps {
+    strokeColor?: string;   // Different from react-native-svg and web impl -- they just use "stroke"
+    strokeWidth?: NumberProp;
+    strokeOpacity?: NumberProp;
+    // These typings are incompatible between web and react-native-svg.  Requires further understanding...
+    // strokeDasharray?: ReadonlyArray<number> | string,
+    strokeDashoffset?: NumberProp;
```
erictraut: By ReactXP naming standard, this should be "strokeDashOffset". Not sure why "offset" is lowercase here. Same true for LineCap, LineJoin, and MiterLimit.

docs/docs/extensions/imagesvg.md
```diff
@@ -41,18 +41,23 @@ viewBox: string = undefined;
...
+fillColor: string;      // Same as "fill" in react-native-svg and web
+fillOpacity: string|number;
+fillRule: 'evenodd' | 'nonzero';
 
 // Color, width and opacity of stroke; default values are provided by SVG
-strokeColor: color;
-strokeWidth: number;
-strokeOpacity: number;
+strokeColor: string;   // Same as "stroke" in react-native-svg and web
+strokeWidth: string|number;
```
erictraut: See notes below about naming conventions and types.

extensions/imagesvg/src/common/Types.ts
deregtd: Because both the web and the native implementations support number|string, and it's much faster than having to do Number(x) everywhere.

extensions/imagesvg/src/common/Types.ts
deregtd: I copied it all out of react-native-svg.

deregtd:
I'm going to change to making a new filesvg extension made to load and display svg files rather than polluting the more simplistic imagesvg made for dynamic svg generation/rendering.

Issue #768, opened 2018-08-03
Text forces pointer cursor


pronebird:
Hi,

Adding `onPress` handler on `<Text>` nodes forces `pointer` cursor. I suppose we should let the developer decide what the desired cusor is:

https://github.com/Microsoft/reactxp/blob/master/src/web/Text.tsx#L130

Same as with `Buttons` before, `Text` overwrites the original `cursor` 😅

zarubond:
Here it a PR to fix the problem [https://github.com/Microsoft/reactxp/pull/793](https://github.com/Microsoft/reactxp/pull/793) .

PR #793, opened 2018-09-04
#768 disable cursor type override

zarubond:
Currently the cursor type can be overridden by the ReactXP despite developers´s explicit cursor selection. This PR allows the ReactXP to change the cursor only if the developer have not selected one.


erictraut:
Thanks for the contribution. Change looks good. It will appear in the next release of ReactXP.

PR #796, opened 2018-09-07
Fix regression introduced by #753

reseul:
Right click stopped working in UWP due to the change in isRightMouseButton. 
The expected parameter is a synthetic event as received from RN implementations, with a nativeEvent payload.
Change #753 cast that to the native payload expected by toMouseButton, which is incorrect. That broke UWP and I seriously doubt it worked for Mac.
I kept the use of toMouseButton by fixing isRightMouseButton to pass the right payload.

More, using Types.TouchEvent (or any Types.* definition) to describe the nativeEvent is wrong in my opinion. It's proven by the first thing toMouseButton used to do: casting to any, not to mention by the regression itself.
So I switched the parameter type to "any" and I removed SOME of the useless casts. I didn't touch the gesture code because it's complex and I may not understand it well, perhaps @thezero could fix when he has time. 

Cc: @alregner

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -109,7 +109,7 @@
     "brace-expansion": {
       "version": "1.1.11",
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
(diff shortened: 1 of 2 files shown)


erictraut:
Yeah, I'm guessing it's an older npm. I think it's safe to revert that file.

Issue #300, opened 2017-10-06
RX.Accessibility.announceForAccessibility() leaves behind an hidden object which is reachable at least on MacOSX voice over


marco-ms:
Steps to reproduce the issue:
- On MacOSX have voice over enabled
- Have some code that calls RX.Accessibility.announceForAccessibility('some string here') only once
- Use Voice Over commands like Control+Option+[Left|Right] to navigate
- There is an hidden object that is left behind that still has the same text 'some string here' announced before (see screenshot)

A workaround to remove that hidden object is to call RX.Accessibility.announceForAccessibility('') with an empty string.

<img width="498" alt="screen shot 2017-10-05 at 11 18 38 am" src="https://user-images.githubusercontent.com/29441719/31292786-93f31d7a-aa89-11e7-8e8a-358dd88c3b62.png">

erictraut:
@marco-ms, do you know if this is still an issue?

marco-ms:
@erictraut yes it is in the 0.51 branch as well.
The problem is due to the hidden div created for announceForAccessibility

https://github.com/Microsoft/reactxp/blob/0.51.x/src/web/RootView.tsx line 334
Looking at this link[1] it seems possible that the role='text' is making it focusable, so I tried locally an example and using tabindex='-1' but it didn't fix the issue, the div is still focusable.

[1] http://accessibleculture.org/articles/2010/05/tabindex/

roger4ever:
The problem still exists, it's caused by the hidden div created for announceForAccessibility. In addition, since the div has a default aria-role of group, the announced string will become "Group, ${announcement}"

The fix should be pretty simple, all you need is to add a few attributes to the hidden div:
aria-hidden="true" tabindex="-1" aria-label="" role="presentation" aria-live="assertive"

I believe the above attributes will hide the div from the screen reader object tree, and prevent the "group" being announced

PR #798, opened 2018-09-07
[Web] #300 Fix the bug where AccessibilityAnnouncer creates a div which is reachable by screen readers

roger4ever:
#300  Fix the bug where AccessibilityAnnouncer would create a div which is reachable by screen readers, the role of the div "Group" was also included in the announcement.

The combination of aria-hidden={ 'true' }, role={ 'presentation' } and tabIndex={ -1 } ensures that this div and its children are hidden from screen readers, they become unreachable and their roles won't be announced anymore.

However, this does mean that the screen reader will not be reading { announcement } at line 120 anymore. Instead it will be reading the attribute aria-label. The announcement is still triggered by the change of { announcement }.

### DIFF
--- a/src/web/AccessibilityAnnouncer.tsx
+++ b/src/web/AccessibilityAnnouncer.tsx
@@ -112,6 +112,10 @@ export class AccessibilityAnnouncer extends React.Component<{}, AccessibilityAnn
                 aria-live={ AccessibilityUtil.accessibilityLiveRegionToString(Types.AccessibilityLiveRegion.Assertive) }
                 aria-atomic={ 'true' }
                 aria-relevant={ 'additions text' }
+                aria-hidden={ 'true' }
+                role={ 'presentation' }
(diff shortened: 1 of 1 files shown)


mshoho:
@erictraut please revert this PR, it breaks the announcement completely.

@roger4ever I am wondering how did you test it? aria-hidden=true hides the element from the screen readers completely, including the ones with aria-live.

erictraut:
OK, I've reverted it in master.

roger4ever:
@mshoho I tested on Windows with Narrator and NVDA, they worked well. I didn't have a Mac, so I borrowed on and tested today and figured that this doesn't work on Mac with Voice Over.

I found a solution which simply makes the aria-hidden property a state, which is toggled when there's a new announcement. I created another PR #801, since I can't re-activate this one.

Thanks for pointing out the issue.

PR #800, opened 2018-09-10
Fix bug: if there are only cached popups and no visible one the main …

zholobov:
…view remains inaccessible for screen readers as if a visible popup is shown. Also the cached popups are potentially accessible to screen readers.

### DIFF
--- a/src/native-common/FrontLayerViewManager.tsx
+++ b/src/native-common/FrontLayerViewManager.tsx
@@ -179,6 +179,17 @@ export class FrontLayerViewManager {
         );
     }
 
+    private _getOverlayContext(rootViewId?: string | null): PopupStackContext {
+        return _.findLast(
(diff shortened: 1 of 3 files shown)


zholobov:
@ladipro @erictraut can you please review?

src/native-common/FrontLayerViewManager.tsx
```diff
@@ -179,6 +179,17 @@ export class FrontLayerViewManager {
         );
     }
 
+    private _getOverlayContext(rootViewId?: string | null): PopupStackContext {
```
erictraut: The return type here should be PopupStackContext | undefined. That should allow you to remove the incorrect "as PopupStackContext" cast below.

ladipro:
@zholobov thank you for fixing this!

PR #801, opened 2018-09-10
[Web] #300 Fix the bug where AccessibilityAnnouncer creates a div which is reachable by screen readers

roger4ever:
This is a continuation of #798, the previous PR was closed and I cannot reopen it.

#300 Fix the bug where AccessibilityAnnouncer would create a div which is reachable by screen readers, the role of the div "Group" was also included in the announcement.

The combination of aria-hidden={ 'true' }, role={ 'presentation' } and tabIndex={ -1 } ensures that this div and its children are hidden from screen readers, they become unreachable and their roles won't be announced anymore.

The aria-hidden={'true'} will hide the object from aria and disable the aria-live function. So it's changed to a state ariaHidden, which will be set to true when there's no announcement through _startClearAnnouncementTimer, and set to false when there's a new announcement.

However, this does mean that the screen reader will not be reading { announcement } at line 126 anymore. Instead it will be reading the attribute aria-label. The announcement is still triggered by the change of { announcement }.

### DIFF
--- a/src/web/AccessibilityAnnouncer.tsx
+++ b/src/web/AccessibilityAnnouncer.tsx
@@ -119,9 +119,7 @@ export class AccessibilityAnnouncer extends React.Component<{}, AccessibilityAnn
                 aria-atomic={ 'true' }
                 aria-relevant={ 'additions text' }
                 aria-hidden={ this.state.ariaHidden }
-                role={ 'presentation' }
-                tabIndex={ -1 }
-                aria-label={ this.state.announcementText }
+                role={ _isMac ? '' : 'Status' }
             >
                 { announcement }
             </div>
--- a/src/web/AccessibilityAnnouncer.tsx
+++ b/src/web/AccessibilityAnnouncer.tsx
@@ -112,6 +112,10 @@ export class AccessibilityAnnouncer extends React.Component<{}, AccessibilityAnn
                 aria-live={ AccessibilityUtil.accessibilityLiveRegionToString(Types.AccessibilityLiveRegion.Assertive) }
                 aria-atomic={ 'true' }
                 aria-relevant={ 'additions text' }
+                aria-hidden={ 'true' }
+                role={ 'presentation' }
+                tabIndex={ -1 }
+                aria-label={ this.state.announcementText }
             >
                 { announcement }
             </div>
--- a/src/web/AccessibilityAnnouncer.tsx
+++ b/src/web/AccessibilityAnnouncer.tsx
@@ -19,6 +19,9 @@ export interface AccessibilityAnnouncerState {
     // Screen Reader text to be announced.
     announcementText: string;
 
+    // Hides the div from aria tree when it's not used so that it can't be accessed from screen reader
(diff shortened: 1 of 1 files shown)


mshoho:
I need to test this. I have one concern in mind to check out. Please, give me a little time.

mshoho:
@roger4ever I have tried it on Mac.

It doesn't work with the presence of `role={ 'presentation' }`.
When I remove that role, it reads the text twice because of aria-label.

Also `tabIndex={ -1 }` makes no sense for this element. Negative tabIndex means that the element is focusable programmatically (and VoiceOver sees the elements with negative tabIndex and focuses them).

roger4ever:
On Windows, Narrator announces "Group, {announcement}" without role={ 'presentation' }. I couldn't find a way to hide the role from the screen reader. Although adding role={ 'presentation' } hides the object completely, using tabIndex={ -1 } makes it focusable programmatically so that screen reader is still able to announce the event when the contents are updated.

If I keep going this route, the only way I can think of is to separate Windows and Mac. But I think the attempt to hide the object from the visual tree while making screen reader be able to detect the announcement is hacky, and has a potential to create a regression if screen reader behavior changes in the future.

I suggest to add role={ 'Status' } for Windows. Because "Status {announcement}" is better than "Group {annoucement}" anyway. And it shouldn't affect Mac behaviour. 
@mshoho I updated the file, what do you think?

erictraut:
@mshoho, any thoughts on @roger4ever's last question? I'd like to figure out what to do with this PR — either get it to a point where it's ready to merge or abandon it.

mshoho:
@erictraut I will test it this week. But as we have other problems with the announcer on web (the recent Chrome seems to have aria-live broken completely), I see what else could be done and we're about to try the Speech Synthesis API which is available in all modern browsers and might give a more reliable result.

mshoho:
The updated version does work on Mac, but doesn't work reliably with a combination of Electron+NVDA...

roger4ever:
Hmm... What exactly is not working? Are you talking about the Electron for building native apps? I only changed src/web/AccessibilityAnnouncer.tsx so it shouldn't affect native app generation.

The only thing I added is a ariaHidden property, so if there are still issues then I don't know what else can be done. But this will be the version that I will use in our app, until React XP has a better web announcer like the  Speech Synthesis API you mentioned.

mshoho:
@roger4ever Electron is Chromium and uses web. So, basically it doesn't work in Chrome+NVDA.

mshoho:
We've had quite a painful history with this function and it's very fragile. I suspect Chrome+NVDA doesn't catch the aria-hidden change fast enough. What we can try is to remove aria-hidden first, and update the label only after some time. Which complicates it. Plus there are known combinations (like Chrome+VoiceOver on Mac) where aria-live is broken on a browser level.

The main problem with the SpeechSynthesis one is that there is no way to speak something for the screen reader only with it. So it requires a UI switch.

erictraut:
@roger4ever, any further thoughts on this PR? If not, I'm going to suggest that we close it.

PR #805, opened 2018-09-12
web: Change import require declaration to ECMAScript 2015

a-tarasyuk:


erictraut:
@a-tarsyuk, thanks for doing this. Looks great!

PR #812, opened 2018-09-14
Allow view to be draggable

mihaimedintu:
Allow drag property to views


erictraut:
RX.View was originally intended to be a simple container. It has grown a bit bloated over time, so I've been reluctant to add more functionality to it. However, it already contains support for an RX.View acting as a drop target, so there's a good argument to be made that it should be a draggable target too. 

That said, this PR presents a lot of questions we need to answer before I feel comfortable incorporating it.

The drop target support was originally added so users could drag items from outside of the app and drop them within the app (e.g. files or text). It was not added for internal drags. There are better ways to implement internal drags using app logic and RX primitives. Am I correct in assuming that this new drag functionality is intended to support dragging items from inside the app to outside?

The PR implements onDragStart. This is only one of the three events that [HTML exposes for draggable targets](https://www.w3schools.com/tags/ev_ondragstart.asp). What about onDrag and onDragEnd? Are these not required because you're focusing only on dragging items outside the app?

What happens if the caller specifies draggable=true but doesn't provide an onDragStart handler? Is that combination valid? If not, why not simply infer that something is draggable if it has on onDragStart handler?

How does the app logic specify the type of information being dragged? Is this done through a dataTransfer object in the event with a setData method? If so, what data types are guaranteed to be supported? Just 'text' or 'text/plain'? Are other file types allowed? 

Is there a way to set the [drag feedback image](https://www.w3schools.com/tags/ev_ondragstart.asp) and the [drag effect](https://www.w3schools.com/tags/ev_ondragstart.asp)?

How do drags initiated through this mechanism interact with internal drop targets (i.e. other RX.View components that have an onDragEnter and onDrop callback defined)? Can they receive such a drag?

Finally, if we do incorporate this new functionality, you'll need to include documentation (add to the existing View.md file) and unit tests (add a new test case to the RXPTest sample app).

shameister:
Hello Eric,

we are also at the point, where we like to introduce what you call 'internal dragging'. 

> There are better ways to implement internal drags using app logic and RX primitives.

In our case, we like to move calendar entries on the y-axis of a time-scale. Which components and events would you use, to implement the desired behavior? By browsing the docs, I was not able to find the right ones, so far. I'll be happy to get your advise.

bildschirmfoto 2018-09-17 um 08 32 49

Best regards,
Sascha

erictraut:
Complex internal drag operations (those that require pointer capture because the drop target is in a different part of the view hierarchy) are not possible to implement today with ReactXP's APIs. For simple internal drag operations (those that involve moving a RX.View within its parent container), you should be able to use an animated transform style attribute to track the mouse location during the drag. This will guarantee that the view remains beneath the pointer for the duration of the drag. 

The closest sample I can think of is the [GestureView test](https://github.com/Microsoft/reactxp/blob/master/samples/RXPTest/src/Tests/GestureViewTest.tsx).

There hasn't been demand for complex internal drag actions in ReactXP so far. This is probably because this type of gesture doesn't work well for touch interfaces. Most devices (with the exception of Macs) support touch, so new UIs tend to be designed with touch in mind. For a calendar app, I can see how dragging with either a mouse or a finger makes sense.

mihaimedintu:
Hi Eric,
The functionality that was implemented here came as a need to drag and drop a view from one part of the application to the other (different views). Although we have plans to add also out-of-app drop possibility, it is not included in this PR. We have tested the behavior on drag and drop in our application and we can see that dragging and dropping inside the application works good with existing onDragEnter/onDragOver/onDragLeave. I choose this implementation mainly because react supports these operations and I thought that it would be simpler than using gesture views.

Draggable and onDragStart are two separate properties, that is why I implemented it like that. As a method of precaution, I think that we can combine these two and if you specify onDragStart, the element will get automatically also draggable attribute. Also, onDrag, onDragEnd and onDragExit are supported by React, so I believe there is no harm in adding support for these methods.

Regarding the data that is being dragged, the information is indeed set through dataTransfer. We tried with simple text (JSON.stringify on an object) and there were no problems getting the data on the other side. Also, we have some examples where you can start dragging files and it is much more straight forward than with text. With text you need to know the exact keys of the dataTransfer text, but with files you can just browser through the files property of dataTransfer.

For now there is no way to set the feedback image of the dragged element, but this is something that we have in mind and we are considering adding it soon.

Regarding tests, I would like to write any tests necessary, but I need to figure out how they work :). There are also tests missing for onDragEnter/onDragOver/onDragLeave, so I think also implementing those will be a good thing.

mihaimedintu:
Hi Eric,

I updated the code and removed the draggable property. If the onDragStart property is available, then draggable is set automatically.

I also added some tests for drag and drop or files from outside of app and for views inside of app.

erictraut:
Looks great! Thanks for the changes.

PR #825, opened 2018-09-25
Remove code duplication for render in RX.Text

gmangov:
User override method to get extended properties in order to avoid duplicating render for RX.Text in derived components.


gmangov:
@erictraut is added to the review. #Closed

gmangov:
@erictrout A small change proposed by radus to avoid code duplication

PR #841, opened 2018-10-04
Make buttons drag and droppable

RamyaSenk:
Currently we have ability to drag and drop views only. Extending the ability to make buttons drag and droppable. 

For web - added the properties to button. 
For native platforms - Buttons are internally views. So the properties will be available as it is already existing.

### DIFF
--- a/samples/RXPTest/src/Tests/DragAndDropTest.tsx
+++ b/samples/RXPTest/src/Tests/DragAndDropTest.tsx
@@ -37,7 +37,7 @@ const _styles = {
     dragAndDropContainer: RX.Styles.createViewStyle({
         flexDirection: 'row',
         flex: 1,
-        height: 70
+        height: 100
     }),
     dragAndDropView: RX.Styles.createViewStyle({
         flex: 1,
@@ -50,8 +50,14 @@ const _styles = {
     }),
     dragAndDropObject: RX.Styles.createViewStyle({
         height: 40,
-        width: 40,
+        width: 80,
         backgroundColor: CommonStyles.successTextColor,
+        marginTop: 5
+    }),
+     dragAndDropButton: RX.Styles.createViewStyle({
+        height: 40,
+        width: 80,
+        backgroundColor: CommonStyles.buttonBackgroundColor,
         marginTop: 5
     })
 };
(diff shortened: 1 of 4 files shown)


samples/RXPTest/src/Tests/DragAndDropTest.tsx
```diff
@@ -185,55 +214,101 @@ class DragAndDropView extends RX.Component<RX.CommonProps, DragAndDropViewState>
         );
     }
 
-    private _onDragEnterForFiles = (e: RX.Types.DragEvent) => {
+    private _getFilesDropped(e: RX.Types.DragEvent) {
```
a-tarasyuk: Seems `.map` is more appropriate to use in this case.

```typescript
private _getFilesDropped(e: RX.Types.DragEvent): FileInfo[] {
    return _.map(e.dataTransfer.files, (file) => ({
        fileName: file.name,
        fileSize: Math.round(fileData.size / 1024),
    }));
}
```

What do you think? ;)

erictraut:
I don't think we should extend drag and drop functionality to buttons. It adds unnecessary interface and implementation complexity to the RX.Button class. You can already easily place a button inside a drag-and-droppable view with just a few extra lines of code if you want this behavior.

RamyaSenk:
@erictraut thanks for your comment. 

What additional implementational complexity has this added? We are just assigning the props to the appropriate handler.

The problem that we have with adding another view around the button to make i drang and droppable is, in a long list like a list of messages - we will be adding x number of views to be rendered. And this causes performance regressions in the message stream, which is not desired.

erictraut:
The RX.Button is designed to have a specific and limited role in ReactXP. It's not meant to have all the capabilities of a general view. It is a simple container that handles presses and touches. I was already hesitant to add drag and drop to RX.View, but I reluctantly did so because there was some precedence for it. I don't see any justification for adding it to RX.Button. 

If I were to apply your logic, I'd be able to justify collapsing every component in RX into a single RX.ViewThatDoesEverything component that takes a union of all parameters supported by today's RX components. Sorry, but I have to hold the line on this one. You can wrap a RX.Button in an RX.View or place an RX.View in a button to compose additional behaviors.

RamyaSenk:
@erictraut Buttons in web have some additional support form browser. It is not the same as a general view. We do have performance issues when wrapping things with extra views in our app. I feel like this will again add to those performance issues. ReactJs supports this for buttons.. 

And how come there was issues adding drag and drop to RX.View? Views have these functionality in react.

RamyaSenk:
I dont necessarily agree, since  it is available in react js and my understanding is react xp is n abstraction of it. But seems like my assumption is wrong and react xp is designed differently. I will abandon the pull request, thanks for having a look.

Issue #832, opened 2018-10-01
RX.Alert does not pass the Android specific AlertOptions object down to RN


luisnaranjo733:
The default behavior on Android is for dialogs to close when the user touches away from them. RN.Alert has an AlertOptions object that you can pass which allows you to configure this behavior, as well as get an onDismiss callback.

ReactXP does not pass this prop down to RN, making this behavior unreachable with RX.Alert.
I know that XP tries to keep away from platform specific stuff, but this is a pretty commonly desired feature on Android. Does this make the cut for getting added to RX.Types.AlertOptions so it can be plumbed down to RN?

erictraut:
The ```PopupOptions``` interface supports an option called ```preventDismissOnPress```. I think it would make sense to add an option with the same name to ```AlertOptions```. It could be implemented on all platforms except for iOS. Does that sound good? If so, feel free to implement it (along with documentation and test code) and submit a PR.

luisnaranjo733:
I like that.
I went ahead and opened a [PR](https://github.com/Microsoft/reactxp/pull/843/files).

PR #843, opened 2018-10-07
Pass the cancelable prop down to RN.Alert

luisnaranjo733:
Addresses issue #832

### DIFF
--- a/docs/docs/apis/alert.md
+++ b/docs/docs/apis/alert.md
@@ -64,6 +64,9 @@ interface AlertOptions {
     // Optional: the id of the root view this alert is associated with.
     // Defaults to the view set by UserInterface.setMainView().
     rootViewId?: string;
+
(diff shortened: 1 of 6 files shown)


luisnaranjo733:
I've just implemented web. Before I add the test case I wanted to ask, what is the procedure for adding a test case?

RXPTest seems to use a published version of ReactXP, so my changes in this PR won't be visible to the test app right away.

luisnaranjo733:
Ok I've added a test case and verified that it works by manually copying the dist files over.

erictraut:
Looks good. Thanks for the contribution. I generally don't add new APIs on stable branches, so this change will need to wait until the next major release (1.5.0). I don't yet have a timeline for that, but we've typically been releasing a new major release every three months or so. It depends on how many other changes come in. It's only been a few weeks since we released 1.4.0.

PR #847, opened 2018-10-08
Simple abstraction to support RN.Animated.Event semantics

deregtd:
First whack at building a simple abstraction on top of the new RN.Animated.event() ability into the RX.ScrollView control. RN.Animated.event nominally lets you handle any parameter of the native callback object and map it into an animated value, but, in reality, for a scrollview, there's only x and y scroll position to take into account. As such, my theory is that we should simplify the abstraction down to simply passing in a scroll{X,Y}AnimatedValue property to the scrollview, and it can do the appropriate wacky mappings to the react native event objects to map it through.

After implementing this, I realized that the web's AnimatedValue class was making a bad assumption.  It was assuming that you could only interpolate a single value from another value, so they became coupled. This is broken in the common case with native combined with animated.event, where you have a single native-mapped animated.value that's mapped to a scroll position, and then interpolate several other values from that root value.

Opening PR to validate with @erictraut that the approach is fine, and then I'll update the sample app before merge if so.

### DIFF
--- a/src/web/Animated.tsx
+++ b/src/web/Animated.tsx
@@ -78,9 +78,8 @@ interface ValueListener {
 
 // The animated value object
 export class Value extends RX.Types.AnimatedValue {
-    private _value: number|string;
+    protected _value: number|string;
     private _listeners: ValueListener[];
-    private _interpolationConfig: { [key: number]: string|number } | undefined;
 
     // Initializes the object with the defaults and assigns the id for the animated value.
     constructor(value: number) {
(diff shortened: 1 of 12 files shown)


src/web/Animated.tsx
```diff
@@ -611,11 +685,7 @@ function createAnimatedComponent<PropsType extends RX.Types.CommonProps>(Compone
 
         // Generates the CSS value for the specified attribute given
         // an animated value object.
-        private _generateCssAttributeValue(attrib: string, valueObj: Value, newValue: number|string): string {
-            if (valueObj._isInterpolated()) {
-                newValue = valueObj._getInterpolatedValue(newValue as number);
-            }
-
+        private _generateCssAttributeValue(attrib: string, newValue: number|string): string {
```
erictraut: Nice simplification.

Nit: Add space around "|".

src/web/Animated.tsx
```diff
@@ -624,11 +694,7 @@ function createAnimatedComponent<PropsType extends RX.Types.CommonProps>(Compone
             return newValue;
         }
 
-        private _generateCssTransformValue(transform: string, valueObj: Value, newValue: number|string): string {
-            if (valueObj._isInterpolated()) {
-                newValue = valueObj._getInterpolatedValue(newValue as number);
-            }
-
+        private _generateCssTransformValue(transform: string, newValue: number|string): string {
```
erictraut: Nit: Add space around "|".

src/windows/ScrollView.tsx
```diff
@@ -11,32 +11,28 @@ import * as React from 'react';
 import * as RN from 'react-native';
 
 import { ScrollView as ScrollViewBase } from '../native-common/ScrollView';
-import { Types } from '../common/Interfaces';
 import EventHelpers from '../native-common/utils/EventHelpers';
 
 export class ScrollView extends ScrollViewBase {
 
-    protected _render(props: Types.ScrollViewProps): JSX.Element {
+    protected _render(nativeProps: RN.ScrollViewProps&React.Props<RN.ScrollView>): JSX.Element {
```
erictraut: Nit: Add space around "&"

PR #853, opened 2018-10-10
Emulate Cache-Control: max-stale on iOS

ladipro:
iOS does not seem to respect the max-stale header. It does not load
expired images from the cache unless the force-cache option (internally
NSURLRequestReturnCacheDataElseLoad) is passed.

We can't use force-cache always because it has the opposite problem of
loading only from the cache and using stale data even when the app is
online and the data served under the URL has changed.

The proposed solution is to fall back to force-cache only on encountering
the first error.

### DIFF
--- a/src/native-common/Image.tsx
+++ b/src/native-common/Image.tsx
@@ -181,7 +181,7 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
 
         if (!this.state.forceCache && this._shouldForceCacheOnError()) {
             // Some platforms will not use expired cache data unless explicitly told so.
-            // Let's try again with cache: 'force-cache'.
+            // Let's try again with cache: 'only-if-cached'.
             this.setState({ forceCache: true, lastNativeError: e.nativeEvent.error });
         } else if (this.props.onError) {
             if (this.state.forceCache) {
--- a/src/native-common/Image.tsx
+++ b/src/native-common/Image.tsx
@@ -31,7 +31,12 @@ export interface ImageContext {
     isRxParentAText?: boolean;
 }
 
-export class Image extends React.Component<Types.ImageProps, Types.Stateless> implements React.ChildContextProvider<ImageContext> {
+export interface ImageState {
+    forceCache?: boolean;
+    lastNativeError?: any;
+}
+
+export class Image extends React.Component<Types.ImageProps, ImageState> implements React.ChildContextProvider<ImageContext> {
     static childContextTypes: React.ValidationMap<any> = {
         isRxParentAText: PropTypes.bool.isRequired,
     };
@@ -59,7 +64,6 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
     protected _mountedComponent: RN.Image | null = null;
     private _nativeImageWidth: number | undefined;
     private _nativeImageHeight: number | undefined;
-    private _forceCache = false;
 
     protected _getAdditionalProps(): RN.ImageProperties | {} {
         return {};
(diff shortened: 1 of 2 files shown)


ladipro:
Forgot to add code to reset ```_forceCache``` back to false when the props change. Will do it shortly.

src/native-common/Image.tsx
```diff
@@ -109,6 +112,12 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
         );
     }
 
+    componentWillReceiveProps(nextProps: Types.ImageProps) {
+        if (!_.isEqual(this.props, nextProps)) {
```
a-tarasyuk: Do we need to check all props to set `this._forceCache`?

ladipro: We can be selective here but it doesn't hurt to retry without ```force-cache``` on all prop changes to keep the code simple and future-proof.

a-tarasyuk: ok. props can include callbacks _(not references, `onLoad: () => {}`)_ and this conditions could fail, also `isEqual` performs a deep comparison between two values - so it can impact on performance. Just need to keep in mind that.

ladipro: Great point, thanks. The web implementation also compares only a few props so I've done the same here.

ladipro:
@berickson1 you have some context on this problem. This is addressing the issue with inconsistent native caching behavior across platforms in S4L. Specifically bringing iOS in-line with Android.

src/native-common/Image.tsx
```diff
@@ -179,10 +195,27 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
         if (this.props.headers) {
             source.headers = this.props.headers;
         }
+        if (this._forceCache) {
```
berickson1: This looks to me like it's actually state. The only reason you make it here on changing is by calling forceUpdate, which is somewhat of an anti-pattern anyways

src/native-common/Image.tsx
```diff
@@ -179,10 +195,27 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
...
+            source.cache = 'force-cache';
```
berickson1: Can you explain why you use force-cache instead of only-if-cached?

force-cache will try to make a network request if there's no data in the cache, which we already expect won't work (since we're in the failure path)

src/native-common/Image.tsx
```diff
@@ -179,10 +195,27 @@ export class Image extends React.Component<Types.ImageProps, Types.Stateless> im
...
         return source;
     }
 
+    private _shouldForceCacheOnError(): boolean {
+        if (Platform.getType() !== 'ios') {
+            return false;
+        }
+        if (this.props.headers) {
+            for (let key in this.props.headers) {
+                if (key.toLowerCase() === 'cache-control' && this.props.headers[key].toLowerCase() === 'max-stale') {
```
berickson1: The spec indicates I can also provide a value
`max-stale=<x seconds>`
In this case, it seems you would ignore it.  At the very least you should document the divergence in behaviour from the spec 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

src/native-common/Image.tsx
ladipro: I wanted to avoid introducing state for perf reasons. Browsing through the rxp sources it doesn't look that uncommon to track state in fields and call forceUpdate.

src/native-common/Image.tsx
ladipro: Yes, since I don't have access to cache internals I can't easily implement the time bounded variant. I'll add a comment / update docs.

src/native-common/Image.tsx
ladipro: I was under the wrong impression that stateless components have better perf characteristics. Converted to state as suggested, thanks!

ladipro:
> nit. you can avoid defining the `constructor`, just add state like so
> 
> ```
> readonly state: ImageState = { forceCache: false, lastNativeError: undefined };
> ```

Awesome, thanks. And it also builds now :p

PR #856, opened 2018-10-14
Updated dev dependencies

victorsolaya:
Updated dev dependencias of React DOM, Native, Windows Native in the package json and package lock json

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,11 @@
     "react-native-windows": "^0.55.0-rc.0"
   },
   "devDependencies": {
+    "eslint": "^3.19.0",
+    "react": "^16.5.2",
+    "react-dom": "^16.5.2",
+    "react-native": "^0.57.3",
+    "react-native-windows": "^0.55.0-rc.0",
     "tslint": "5.11.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,6 +28,11 @@
     "react-native-windows": "^0.55.0-rc.0"
   },
   "devDependencies": {
+    "eslint": "^3.19.0",
```
erictraut: Why are you adding a devDependency on eslint? We don't use this in any of our scripts. Plus, eslint doesn't provide much (any?) value in a project written in typescript. If you want to use eslint, feel free to install it on your system globally, but I don't want to add unnecessary dependencies to ReactXP.

package.json
```diff
@@ -28,6 +28,11 @@
...
+    "react": "^16.5.2",
```
erictraut: Why are you adding these four "react*" packages as "devDependencies"? They're already properly specified in the "peerDependencies" section. I don't think they should be duplicated here.

erictraut:
@victorsolaya, any response to my questions above? If not, I'm going to close the PR.

PR #870, opened 2018-10-19
Filter out emulated Cache-Control: max-stale on iOS

ladipro:
The header has the opposite effect on iOS: instead of having the cache
return stale data it disables the cache altogether. The header is
emulated by the component as of #853. This change adds filtering so that
RN doesn't see the header.

### DIFF
--- a/src/native-common/Image.tsx
+++ b/src/native-common/Image.tsx
@@ -34,6 +34,7 @@ export interface ImageContext {
 export interface ImageState {
     forceCache?: boolean;
     lastNativeError?: any;
+    headers?: Types.Headers;
 }
 
 export class Image extends React.Component<Types.ImageProps, ImageState> implements React.ChildContextProvider<ImageContext> {
@@ -64,7 +65,7 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
     protected _mountedComponent: RN.Image | null = null;
     private _nativeImageWidth: number | undefined;
     private _nativeImageHeight: number | undefined;
-    readonly state: ImageState = { forceCache: false, lastNativeError: undefined };
+    readonly state: ImageState = { forceCache: false, lastNativeError: undefined, headers: this._buildHeaders() };
 
     protected _getAdditionalProps(): RN.ImageProperties | {} {
         return {};
@@ -121,7 +122,7 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
         const sourceOrHeaderChanged = (nextProps.source !== this.props.source ||
             !_.isEqual(nextProps.headers || {}, this.props.headers || {}));
         if (sourceOrHeaderChanged) {
-            this.setState({ forceCache: false, lastNativeError: undefined });
+            this.setState({ forceCache: false, lastNativeError: undefined, headers: this._buildHeaders() });
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/native-common/Image.tsx
```diff
@@ -202,7 +202,16 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
 
         const source: RN.ImageSourcePropType = { uri: this.props.source };
         if (this.props.headers) {
-            source.headers = this.props.headers;
+            const cacheControlHeader = this._getMaxStaleHeader();
+            if (cacheControlHeader) {
+                // Filter out Cache-Control: max-stale. It has the opposite effect on iOS: instead of having
```
ladipro: Apologies for not catching this when working on #853 
Basically it looks like NSURLRequest disables local cache if it sees a caching header it doesn't understand. So not only doesn't Cache-Control: max-stale do what it's spec'ed to, it actually hurts and has to be removed. cc @berickson1

src/native-common/Image.tsx
```diff
@@ -202,7 +202,16 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
...
+                // the cache return stale data it disables the cache altogether. We emulate the header by
+                // retrying with cache: 'only-if-cached'.
+                source.headers = _.clone(this.props.headers);
```
a-tarasyuk: nit: `_buildSource` is called directly in the `render` method. For images which have headers and the header `Cache-Control: max-stale` `_.clone` method will be called on each re-render. For instance, I want to build an application like ImageList, which will have 50 images on a page with the header `Cache-Control: max-stale`, for this case, the method `clone` will be called 50 times, it will take some time.

ladipro: Gotcha, yes, this code doesn't have to run on each render. I have added headers to state. Thanks!

src/native-common/Image.tsx
```diff
@@ -194,37 +195,46 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
...
     }
 
+    private _buildHeaders(): Types.Headers|undefined {
+        if (this.props.headers) {
+            const cacheControlHeader = this._getMaxStaleHeader();
+            if (cacheControlHeader) {
+                // Filter out Cache-Control: max-stale. It has the opposite effect on iOS: instead of having
+                // the cache return stale data it disables the cache altogether. We emulate the header by
+                // retrying with cache: 'only-if-cached'.
+                const headers = _.clone(this.props.headers);
```
a-tarasyuk: seems that `omit` is also acceptable in this case. 
```typescript 
return _.omit(this.props.headers, [cacheControlHeader])
```
what do you think?

ladipro: Nice and concise, thanks!

src/native-common/Image.tsx
```diff
@@ -121,7 +122,7 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
         const sourceOrHeaderChanged = (nextProps.source !== this.props.source ||
```
a-tarasyuk: fyi: [componentWillReceiveProps](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops) will be deprecated in the future. 

> Using this lifecycle method often leads to bugs and inconsistencies, and for that reason it is going to be deprecated in the future.

PR #872, opened 2018-10-19
Point RN MacOS TextInput at the Native common implementation

### LINKED ISSUE #871: MacOS TextInput clears input on unrelated prop change
If I have a TextInput object and I'm not using the value prop, changing any prop on TextInput causes the input text to be reset.

berickson1:
I reviewed the two and couldn't find any functionality differences (except for missing bug fixes)
Fixes #871


alregner:
> @alregner - can you comment as to why these files were separated? I don't see any differences between MacOS and RN native-common

Thanks for pointing this out. Originally this implementation added submitTextOnEnter property that was passed to native mac implementation. It was dropped later and now this is not needed anymore.

PR #873, opened 2018-10-19
Apply ScrollView focus workaround to MacOS

berickson1:
If a TextInput is located inside a ScrollView on RN MacOS, and the 'keyboardShouldPersistTaps' flag is false, the user will be unable to click on any other elements inside any ScrollView in the app.  Only once they finish editing (submit text via Enter or tab away) will clicks stop being swallowed by the scrollview.
Setting the keyboardShouldPersistTaps flag to true stops the ScrollView from swallowing these clicks


erictraut:
@berickson1, change looks fine to me. Please fix the merge conflict.

PR #880, opened 2018-10-25
feature/update hello world js

a-tarasyuk:


erictraut:
Nice! Thanks for the contribution.

reseul:
@a-tarasyuk nice! Good to know there's a combination of RNW/RN that works.
I'd like to fix RXPtest, just double checking you're not changing that so we don't cross paths.

a-tarasyuk:
@reseul :), No, I've not started work on RXPtest, so you can update it. Right now I'm working on updating [`hello-world`](https://github.com/Microsoft/reactxp/tree/master/samples/hello-world).

PR #899, opened 2018-11-06
Fix web accessibility in reactxp-virtuallistview

berickson1:
1) Ensure that there is a selectable item on vlv mount
2) Ensure that if the focuses item is removed, we pick a new item (instead of asserting)

### DIFF
--- a/extensions/virtuallistview/src/VirtualListView.tsx
+++ b/extensions/virtuallistview/src/VirtualListView.tsx
@@ -345,6 +345,15 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
             if (!(item.key in newItemMap)) {
                 // If we're deleting an item that's above the current render block,
                 // update the adjustment so we avoid an unnecessary scroll.
+                
+                // Update focused item if it's the one removed, if we're unable to, reset focus
+                if (item.key === this.state.lastFocusedItemKey) {
+                    if (!this._selectSubsequentItem(FocusDirection.Down, false) &&
+                            !this._selectSubsequentItem(FocusDirection.Down, false)) {
+                        this.setState({ lastFocusedItemKey: undefined });
+                    }
+                }
+
                 if (itemIndex < this._itemsAboveRenderBlock) {
                     this._heightAboveRenderAdjustment += this._getHeightOfItem(oldItems[itemIndex]);
                 }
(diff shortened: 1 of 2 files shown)


extensions/virtuallistview/src/VirtualListView.tsx
```diff
@@ -345,6 +345,15 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
             if (!(item.key in newItemMap)) {
                 // If we're deleting an item that's above the current render block,
                 // update the adjustment so we avoid an unnecessary scroll.
+                
+                // Update focused item if it's the one removed, if we're unable to, reset focus
+                if (item.key === this.state.lastFocusedItemKey) {
+                    if (!this._selectSubsequentItem(FocusDirection.Down, false) &&
```
a-tarasyuk: Both conditions are equivalent

```typescript
this._selectSubsequentItem(FocusDirection.Down, false) && !this._selectSubsequentItem(FocusDirection.Down, false)
```

I think it is a typo. ;)

extensions/virtuallistview/src/VirtualListView.tsx
```diff
@@ -1193,27 +1202,30 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
...
+            return true;
         }
 
         if (index === -1 && retry) {
             index = this._itemMap[this.state.lastFocusedItemKey];
 
             if (index === undefined) {
                 assert.ok(false, 'Something went wrong in finding last focused item');
-                return;
+                return false;
```
a-tarasyuk: nit: as I understand `return` needs here to avoid TS error? `assert` throws an error, and a code will not continue execution - https://codesandbox.io/s/383xk41nrm

> The [throw](throw https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw) statement throws a user-defined exception. Execution of the current function will stop (the statements after throw won't be executed), and control will be passed to the first catch block in the call stack. If no catch block exists among caller functions, the program will terminate.

berickson1: Typescript doesn't understand that assert throws.  Some projects also choose to strip asserts - this ensures graceful fallback in those cases

extensions/virtuallistview/src/VirtualListView.tsx
berickson1: Good catch

extensions/virtuallistview/src/VirtualListView.tsx
a-tarasyuk: Got it. Thanks.

PR #902, opened 2018-11-08
MacOs Accessibility - Keyboard focus + VLV

berickson1:
Keep this behind a flag until public react-native-mac fully supports this

### DIFF
--- a/src/common/Interfaces.ts
+++ b/src/common/Interfaces.ts
@@ -46,6 +46,8 @@ export abstract class AnimatedView extends AnimatedComponent<Types.AnimatedViewP
 }
 
 export abstract class App {
+    supportsExperimentalKeyboardNavigation = false;
+
     // Initialization
     initialize(debug: boolean, development: boolean): void {
         AppConfig.setAppConfig(debug, development);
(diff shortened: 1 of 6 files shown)


src/common/Interfaces.ts
```diff
@@ -46,6 +46,8 @@ export abstract class AnimatedView extends AnimatedComponent<Types.AnimatedViewP
 }
 
 export abstract class App {
+    supportsExperimentalKeyboardNavigation = false;
```
erictraut: Why add this flag? Is there a downside of unconditionally enabling it? RN.Mac isn't being used broadly, so as long as it works in your scenarios, I'd be OK with just enabling it rather than adding hacky flags that we'll need to remove later.

src/native-common/GestureView.tsx
```diff
@@ -619,7 +629,8 @@ export abstract class GestureView extends React.Component<Types.GestureViewProps
                 accessibilityComponentType={ accessibilityComponentType }
                 accessibilityLabel={ this.props.accessibilityLabel }
                 testID={ this.props.testId }
-                { ...this._panResponder.panHandlers }
+                {...this._panResponder.panHandlers}
```
erictraut: Nit: Please add spaces within the curly braces for consistency.

src/common/Interfaces.ts
berickson1: Unfortunately this flag is not available in the public variant of RN MacOS yet, therefore if it's applied to components, prop validation will fail due to lack of the prop inside ReactNativeViewAttributes.

PR #910, opened 2018-11-14
Fix for VoiceOver iOS 12 adding additional " , " to payload

thewulf7:
VoiceOver iOS 12 adding additional " , " to payload: fix to trim spaces and commas and compare such strings

### DIFF
--- a/src/ios/Accessibility.ts
+++ b/src/ios/Accessibility.ts
@@ -74,7 +74,7 @@ export class Accessibility extends NativeAccessibility {
 
         const postedAnnouncement = this._announcementQueue[0];
         // Handle retries if it's the announcement we posted. Drop other announcements.
-        if (payload.announcement === postedAnnouncement) {
+        if (this._compareRawAnnouncements(postedAnnouncement, payload.announcement)) {
             const timeElapsed = Date.now() - this._retryTimestamp;
 
             if (!payload.success && timeElapsed < RetryTimeout) {
@@ -88,6 +88,11 @@ export class Accessibility extends NativeAccessibility {
                 }
             }
         }
+    }
+
+    private _compareRawAnnouncements(postedAnnouncement: string, payloadAnnouncement: string): boolean {
+        const spacesCommasRegExp = /[s,]/g; // special regex for iOS 12
+        return payloadAnnouncement.replace(spacesCommasRegExp, '') === postedAnnouncement.replace(spacesCommasRegExp, '')
     }
 }
 


src/ios/Accessibility.ts
```diff
@@ -89,6 +89,11 @@ export class Accessibility extends NativeAccessibility {
             }
         }
     }
+
+    private compareRawAnnouncements(postedAnnouncement: string, payloadAnnouncement: string): boolean {
+        const trimSpacesCommas = new RegExp(/[\s,]/);
```
a-tarasyuk: Is it required to replace only first occurrence?

```typescript
'test, test, tets, '.replace(/[\s,]/);
> 'test test, tets, '
```
or need to replace all spaces and commas?

```typescript
'test, test, tets, '.replace(/[\s,]/g);
> 'testtesttets'
```

a-tarasyuk: nit: in this case better to use RegEx literal instead of constructor `new RegExp(/[\s,]/)` -> `/[\s,]/`

src/ios/Accessibility.ts
```diff
@@ -89,6 +89,11 @@ export class Accessibility extends NativeAccessibility {
             }
         }
     }
+
+    private compareRawAnnouncements(postedAnnouncement: string, payloadAnnouncement: string): boolean {
```
a-tarasyuk: I think need to use `_` before name `_compareRawAnnouncements`, just like other private methods

src/ios/Accessibility.ts
thewulf7: As long string is splitting in several with commas with spaces, better to remove them all for comparison

Example: 
```
const postedAnnouncement = "New message from Evgenii Utkin, please write me"
const payloadAnnouncement = "New message from ,  Evgenii Utkin, ,  please write me"
```

thewulf7: @a-tarasyuk We are not using dynamic input here, so why?

src/ios/Accessibility.ts
thewulf7: Agree, fixed

src/ios/Accessibility.ts
thewulf7: fixed

a-tarasyuk: > we are not using dynamic input here

```typescript
new RegExp(/[\s,]/)
```

constructor here is useless because you pass to it regular expression `/[\s,]/` which can be used without constructor

```
const r = /[\s,]/g;
```

Does that make sense?

thewulf7: @a-tarasyuk Ok, I've refactored it a little bit, but I still believe that more strict way makes more sense.

a-tarasyuk: I meant in my previous comment, the following changes: 

```typescript

private _compareRawAnnouncements(postedAnnouncement: string, payloadAnnouncement: string): boolean {
    const spacesCommasRegExp = /[\s,]/g;
    return payloadAnnouncement.replace(spacesCommasRegExp, '') === postedAnnouncement.replace(spacesCommasRegExp, '');
}
```

thewulf7: @a-tarasyuk I got it - still believe this is not TS way to do things)

a-tarasyuk: It is not related to TS, it common JavaScript practice, in which TypeScript will be [_transpired_](http://www.typescriptlang.org/play/index.html#src=const%20spacesCommasRegExp%20%3D%20new%20RegExp('%2F%5B%5Cs%2C%5D%2F'%2C%20'g')%3B)  

[Creating a regular expressionSection](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Creating_a_regular_expression
)
> Regular expression literals provide compilation of the regular expression when the script is loaded. If the regular expression remains constant, using this can improve performance.

> Using the constructor function provides runtime compilation of the regular expression. Use the constructor function when you know the regular expression pattern will be changing, or you don't know the pattern and are getting it from another source, such as user input.

thewulf7: fixed

src/ios/Accessibility.ts
```diff
@@ -89,6 +89,11 @@ export class Accessibility extends NativeAccessibility {
             }
         }
     }
+
+    private _compareRawAnnouncements(postedAnnouncement: string, payloadAnnouncement: string): boolean {
+        const spacesCommasRegExp = /[s,]/g; // special regex for iOS 12
```
berickson1: This regex matches the 's' character and ',' comma

>test, foo

will get transformed to

> tet foo

Meaning tet, foo will incorrectly "match" test, foo

berickson1: @erictraut - this PR requires changes before you publish next reactxp

thewulf7: @berickson1 my bad - mistake during CR fixes - will create a new PR

PR #918, opened 2018-11-15
remove duplicate conditions. add missing ;

a-tarasyuk:


src/native-common/View.tsx
```diff
@@ -343,17 +343,20 @@ export class View extends ViewBase<Types.ViewProps, Types.Stateless> {
...
             accessibilityLiveRegion: AccessibilityUtil.accessibilityLiveRegionToString(props.accessibilityLiveRegion)
         };
-        if (_isNativeMacOs && App.supportsExperimentalKeyboardNavigation && (this.props.onPress ||
-                (this.props.tabIndex !== undefined && this.props.tabIndex >= 0))) {
+
+        if (_isNativeMacOs && App.supportsExperimentalKeyboardNavigation) {
             const macAccessibilityProps: MacComponentAccessibilityProps = accessibilityProps as any;
-            if (this.props.tabIndex !== -1) {
+
+            if (this.props.tabIndex && this.props.tabIndex >= 0) {
```
berickson1: This isn't correct. If tabIndex is 0, this condition will fail and the code below is not run
If tab index is undefined and onPress is defined, the code below should be run but is not

The conditional that was previously above is needed on this block, so I don't think it really optimizes anything by moving it

a-tarasyuk: ok, tabindex: -2

src/native-common/View.tsx
```diff
@@ -343,17 +343,20 @@ export class View extends ViewBase<Types.ViewProps, Types.Stateless> {
...
             accessibilityLiveRegion: AccessibilityUtil.accessibilityLiveRegionToString(props.accessibilityLiveRegion)
         };
-        if (_isNativeMacOs && App.supportsExperimentalKeyboardNavigation && (this.props.onPress ||
-                (this.props.tabIndex !== undefined && this.props.tabIndex >= 0))) {
+
+        if (_isNativeMacOs && App.supportsExperimentalKeyboardNavigation) {
             const macAccessibilityProps: MacComponentAccessibilityProps = accessibilityProps as any;
-            if (this.props.tabIndex !== -1) {
+
+            if (this.props.tabIndex !== undefined && this.props.tabIndex >= 0) {
```
berickson1: This is still a change in behaviour.  If tabIndex is undefined and onPress is defined, the below code should be run.
You're going to end up with the same conditional that was above

a-tarasyuk:
Oh, ok. For the first look, it is a little bit confused :). Have closed PR., In my opinion, this condition is too long and would be better to make it simpler or make a method with a readable name which describes the purpose of the condition. Seems case when tabIndex less than -1 causes the error...

PR #920, opened 2018-11-15
Remove a deep comparison on every button prop change

berickson1:
Instead let _setOpacityStyles handle determining if anything has changed


erictraut:
I'm not opposed to this change, but it's not obvious to me that it's a perf win. Instead of doing a deep comparison, we're now always calling a method to flatten the style, allocating extra objects along the way. What prompted this change in the first place? Do you have some profiler data that shows this was a bottleneck?

PR #922, opened 2018-11-15
Remove undocumented and broken (on RN) API

berickson1:
This API doesn't exist in RN and will crash if these functions are called.
Remove addToScrollLeft
Remove addToScrollTop


samples/RXPTest/src/Tests/ScrollViewEventTest.tsx
```diff
@@ -216,8 +216,6 @@ class ScrollViewView extends RX.Component<RX.CommonProps, ScrollViewState> {
 // focus
 // setScrollLeft
 // setScrollTop
-// addToScrollLeft
```
erictraut: Please remove from the scrollview.md documentation as well.

PR #935, opened 2018-11-20
Expose React.Fragment via ReactXP

berickson1:
This allows for returning multiple components from a render function without needing to wrap them in a View
https://reactjs.org/docs/fragments.html


berickson1:
Currently, you can use React.Fragment directly in RX, but the following syntax will cause runtime exceptions since RX.Fragment does not exist

`
render() {
    return (
        <>
            <RX.Text> Foo </RX.Text>
            <RX.Text> Bar </RX.Text>
        </>
    );
}
`

PR #944, opened 2018-11-27
Rework VirtualListView

berickson1:
- Move the concept of focus/selected into VLV itself - usually we want to select something
- Use React.Fragment to remove one layer of un-needed rendering
- Provide a single object in renderItem callback that contains details about focus/selection state
- Make accessibility in VLV easier for some platforms since we can focus and click on the same items

### DIFF
--- a/docs/docs/extensions/virtuallistview.md
+++ b/docs/docs/extensions/virtuallistview.md
@@ -147,6 +147,10 @@ initialSelectedKey?: string;
 // Ordered list of descriptors for items to display in the list.
 itemList: VirtualListViewItemInfo[];
 
+// Use this if you want to vertically offset the focused item from the 
+// top of the viewport when using keyboard nav
+keyboardFocusScrollOffset?: number;
+
 // Logging callback to debug issues related to the VirtualListView.
 logInfo?: (textToLog: string) => void;
 
@@ -184,6 +188,7 @@ disableBouncing?: boolean; // iOS only, bounce override
 scrollIndicatorInsets?: { top: number, left: number,
     bottom: number, right: number }; // iOS only
 onLayout?: (e: RX.Types.ViewOnLayoutEvent) => void;
+scrollEventThrottle?: number;
 onScroll?: (scrollTop: number, scrollLeft: number) => void;
 scrollXAnimatedValue?: RX.Types.AnimatedValue;
 scrollYAnimatedValue?: RX.Types.AnimatedValue;
--- a/docs/docs/extensions/virtuallistview.md
+++ b/docs/docs/extensions/virtuallistview.md
@@ -142,6 +142,14 @@ interface VirtualListViewItemInfo {
 // Should the list animate additions, removals and moves within the list?
 animateChanges?: boolean;
 
+// Cell recycling is an optimization to re-use similar VLV cells & decrease the cost
+// of render operations. However, in some cases, disabling cell recycling can be
+// benificial, recycled cells continue their regular react lifecycle even when not
+// visible, which can lead to excessive background re-rendering in some cases. When
+//  combining VLV with react libraries (like ReSub) that have subscriptions managed
+// by components can cause this behaviour to manifest.
+disableCellRecycling?: boolean;
+
 initialSelectedKey?: string;
 
 // Ordered list of descriptors for items to display in the list.
--- a/docs/docs/extensions/virtuallistview.md
+++ b/docs/docs/extensions/virtuallistview.md
@@ -86,9 +86,9 @@ class FruitListView extends RX.Component<null, FruitListState> {
         );
     }
 
-    private _renderItem(item: FruitListItemInfo, hasFocus?: boolean) {
+    private _renderItem(details: VirtualListViewCellRenderDetails<FruitListItemInfo>) {
         const viewStyle = RX.Styles.createViewStyle({
-            height: item.height,
+            height: details.item.height,
             backgroundColor: item.template === _headerItemTemplate ?
                 '#ddd' : '#fff',
             alignItems: 'center'
@@ -97,7 +97,7 @@ class FruitListView extends RX.Component<null, FruitListState> {
         return (
             <RX.View style={ viewStyle }>
                 <RX.Text>
-                    { item.text }
+                    { details.item.text }
                 </RX.Text>
             </RX.View>
         );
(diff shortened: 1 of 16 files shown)


extensions/virtuallistview/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "reactxp-virtuallistview",
-  "version": "1.5.0-rc.1",
```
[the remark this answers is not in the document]
erictraut: OK, now that 1.5.0 has been released, I think it makes sense to make this a major version change — a 2.0.0-rc.1

extensions/virtuallistview/src/VirtualListCell.tsx
```diff
@@ -7,19 +7,29 @@
...
+}
+
 export interface VirtualListCellProps<ItemInfo extends VirtualListCellInfo> extends RX.CommonProps {
     // All callbacks should be prebound to optimize performance.
     onLayout?: (itemKey: string, height: number) => void;
     onAnimateStartStop?: (itemKey: string, start: boolean) => void;
-    onCellFocus?: (itemKey: string|undefined) => void;
-    renderItem: (item: ItemInfo, focused: boolean) => JSX.Element | JSX.Element[];
+    onCellFocus?: (itemKey: string | undefined) => void;
+    onItemSelected?: (item: ItemInfo) => void;
```
erictraut: Why is onItemSelected passed the full item but onCellFocus is passed only the key? Seems inconsistent. 

Similarly, I wonder if we should be more consistent with the naming and change onCellFocus to onItemFocus?

extensions/virtuallistview/src/VirtualListCell.tsx
```diff
@@ -323,30 +340,48 @@ export class VirtualListCell<ItemInfo extends VirtualListCellInfo> extends RX.Co
...
         return (
             <RX.Animated.View
-                style={ [_styles.cellView, overflow, this._animatedStylePosition] }
+                style={ [_styles.cellView, overflow, this._animatedStylePosition, this._animatedStyleWidth] }
                 ref={ _virtualCellRef }
                 tabIndex={ this.props.tabIndex }
                 onLayout={ this.props.onLayout ? this._onLayout : undefined }
                 onFocus={ this.props.onCellFocus ? this._onFocus : undefined }
                 onBlur={ this.props.onCellFocus ? this._onBlur : undefined }
+                onPress={this.props.onItemSelected ? this._onPress : undefined}
```
erictraut: This onPress isn't going to properly handle gesture feedback on touch-based devices because you're not using an RX.Button here. With the old VLV, users would typically render an RX.Button if they wanted to make items pressable. By incorporating selection in the VLV, you're taking away this flexibility and removing some functionality in the process. You're also adding functionality, so maybe it's a good tradeoff, but it would be even better if you weren't taking away flexibility. For example, would it make sense to give the user a way to provide his own selection/focus capabilities, overriding the built-in behaviors?

extensions/virtuallistview/src/VirtualListCell.tsx
```diff
@@ -323,30 +340,48 @@ export class VirtualListCell<ItemInfo extends VirtualListCellInfo> extends RX.Co
...
                     renderItem={ this.props.renderItem }
                 />
             </RX.Animated.View>
         );
     }
 
+    private _onKeyPress = (e: RX.Types.KeyboardEvent) => {
+        const isSelectItemKeyPress = e.keyCode === _keyCodeEnter || e.keyCode === _keyCodeSpace;
+        if (isSelectItemKeyPress && this.props.onItemSelected && this.props.item) {
+            this.props.onItemSelected(this.props.item);
```
erictraut: You're not marking the event as handled here. You're also not passing the event to onItemSelected, so the callee has no way of marking the event as handled. So the key press will continue to bubble up through the view hierarchy. Is this what you intend?

extensions/virtuallistview/src/VirtualListCell.tsx
```diff
@@ -323,30 +340,48 @@ export class VirtualListCell<ItemInfo extends VirtualListCellInfo> extends RX.Co
...
+                    isSelected={ this.props.isSelected }
                     item={ this.props.item }
                     renderItem={ this.props.renderItem }
                 />
             </RX.Animated.View>
         );
     }
 
+    private _onKeyPress = (e: RX.Types.KeyboardEvent) => {
+        const isSelectItemKeyPress = e.keyCode === _keyCodeEnter || e.keyCode === _keyCodeSpace;
```
erictraut: What about the return key?

extensions/virtuallistview/src/VirtualListCell.tsx
```diff
@@ -323,30 +340,48 @@ export class VirtualListCell<ItemInfo extends VirtualListCellInfo> extends RX.Co
...
+
     private _onFocus = (e: RX.Types.FocusEvent) => {
         if (this.props.onCellFocus) {
             this.props.onCellFocus(this.props.itemKey);
         }
     }
 
+    private _onPress = (e: RX.Types.SyntheticEvent) => {
+        if (this.props.onItemSelected && this.props.item) {
+            this.props.onItemSelected(this.props.item);
```
erictraut: Same comment as above. You're not marking the event as handled, so it will continue to bubble up the view hierarchy. Is that what you want?

extensions/virtuallistview/src/VirtualListCell.tsx
```diff
@@ -323,30 +340,48 @@ export class VirtualListCell<ItemInfo extends VirtualListCellInfo> extends RX.Co
...
                 />
             </RX.Animated.View>
         );
     }
 
+    private _onKeyPress = (e: RX.Types.KeyboardEvent) => {
+        const isSelectItemKeyPress = e.keyCode === _keyCodeEnter || e.keyCode === _keyCodeSpace;
+        if (isSelectItemKeyPress && this.props.onItemSelected && this.props.item) {
+            this.props.onItemSelected(this.props.item);
+        }
```
erictraut: What if an item is not meant to be selectable, like the static text header for a group of items? Can the user simply not provide an onItemSelected callback in that case? Looks like that's the intent. We should make sure that's documented.

extensions/virtuallistview/src/VirtualListView.tsx
```diff
@@ -63,14 +63,21 @@ export interface VirtualListViewItemInfo extends VirtualListCellInfo {
...
 
     // Ordered list of descriptors for items to display in the list.
     itemList: ItemInfo[];
 
     // Callback for rendering item when it becomes visible within view port.
-    renderItem: (item: ItemInfo, hasFocus: boolean) => JSX.Element | JSX.Element[];
+    renderItem: (renderDetails: VirtualListCellRenderDetails<ItemInfo>) => JSX.Element | JSX.Element[];
+    onItemSelected?: (item: ItemInfo) => void;
+
+    initialSelectedKey?: string;
```
erictraut: I'm not a fan of props that have different semantics based on whether the component was just mounted or not. There are a few other examples of this in RX (notably, the `defaultValue` prop in RX.TextInput), and they tend to cause problems and require the caller to use a `key` to force a mount/unmount in certain cases. I think it would be better to provide a programmatic accessor to set the current selection, and users could call that explicitly after mounting is complete if they want to set the initial selection.

extensions/virtuallistview/src/VirtualListView.tsx
```diff
@@ -1107,17 +1143,25 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
         }
 
         _.each(cellList, cell => {
-            let tabIndexValue: number|undefined;
+            let tabIndexValue = -1;
             let isFocused = false;
+            let isSelected = false;
             if (cell.item) {
                 if (cell.item && cell.item.isNavigable) {
```
erictraut: The first part of this conditional is redundant. (I know this was in the existing code, not in your changes.)

extensions/virtuallistview/src/VirtualListView.tsx
```diff
@@ -1137,13 +1181,16 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
                     top={ cell.cellInfo.top }
                     isVisible={ cell.cellInfo.isVisible }
                     isActive={ cell.item ? true : false }
-                    isFocused={ isFocused }
+                    isFocused={isFocused}
```
erictraut: Nit: Please add back the spaces around `isFocused`.

extensions/virtuallistview/src/VirtualListView.tsx
```diff
@@ -1137,13 +1181,16 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
...
+                    isSelected={ isSelected }
                     tabIndex={ tabIndexValue }
                     onCellFocus={ this._onCellFocus }
+                    onItemSelected={ this._onItemSelected }
                     shouldUpdate={ !this.props.skipRenderIfItemUnchanged || cell.cellInfo.shouldUpdate }
                     showOverflow={ this.props.showOverflow }
                     isScreenReaderModeEnabled={ this._isAndroidScreenReaderEnabled() }
-                    renderItem={ this.props.renderItem }
+                    renderItem={this.props.renderItem}
```
erictraut: Nit: Please add spaces within brace for consistency.

extensions/virtuallistview/src/VirtualListView.tsx
```diff
@@ -1209,6 +1256,24 @@ export class VirtualListView<ItemInfo extends VirtualListViewItemInfo>
         }
     }
 
+    // Sets selection & focus to specified key
+    selectItemKey(key: string) {
+        // Set focus and selection)
```
erictraut: Nit: This comment ends in a weird way)

extensions/virtuallistview/src/VirtualListCell.tsx
berickson1: onCellFocus is a prop only usable by VirtualListView iteself to know when the VirtualListCell is focused
I could rename it & change the signature if you want, but no data besides the key will be used

extensions/virtuallistview/src/VirtualListCell.tsx
berickson1: I went back and forth on this one a little bit. It'll still call back onItemSelected & the listener can choose to ignore it.  I considered changing the isNavigable prop to be isSelectable to bring clickable & focusable elements inline, but I'm not entirely sure if that's desirable

docs/docs/extensions/virtuallistview.md
```diff
@@ -138,54 +138,55 @@ interface VirtualListViewItemInfo {
...
-scrollsToTop?: boolean; // iOS only, scroll to top on status bar tap
-disableBouncing?: boolean; // iOS only, bounce override
-scrollIndicatorInsets?: { top: number, left: number,
-    bottom: number, right: number }; // iOS only
-onScroll?: (scrollTop: number, scrollLeft: number) => void;
-scrollXAnimatedValue?: RX.Types.AnimatedValue;
-scrollYAnimatedValue?: RX.Types.AnimatedValue;
+``` javascript    testId?: string;
+
+    // Ordered list of descriptors for items to display in the list.
```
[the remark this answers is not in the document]
berickson1: The original list seemed to be a direct copy paste from the typescript props definition so I followed that. I think re-ordering makes maintenance harder if someone forgets to update the doc

extensions/virtuallistview/src/VirtualListView.tsx
berickson1: If you don't provide this key, the VLV will automatically make the wrong element the initially selected/focused item. Calling a function to set this would result in an additional render cycle of 2 VLV items, which isn't the end of the world but in some cases could be expensive if the components are heavy.  This is optional, so consumers could also do what you're suggesting

docs/docs/extensions/virtuallistview.md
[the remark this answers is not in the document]
erictraut: It may make maintenance harder, but I think it makes reading the documentation easier, and it's consistent with the other documentation.

extensions/virtuallistview/src/VirtualListCell.tsx
erictraut: For consistency, I think it's worth changing.

extensions/virtuallistview/src/VirtualListView.tsx
erictraut: OK, I'm still not a fan of these types of props, but you make a good argument. I'm fine leaving it as is.

docs/docs/extensions/virtuallistview.md
```diff
@@ -138,54 +138,54 @@ interface VirtualListViewItemInfo {
...
-
+``` javascript    testId?: string;
 // Ordered list of descriptors for items to display in the list.
-itemList: VirtualListViewItemInfo[];
+itemList: ItemInfo[];
 
-// Logging callback to debug issues related to the VirtualListView.
-logInfo?: (textToLog: string) => void;
+// Callback for rendering item when it becomes visible within view port.
+renderItem: (renderDetails: VirtualListCellRenderDetails<ItemInfo>) => JSX.Element | JSX.Element[];
```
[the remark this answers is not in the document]
berickson1: What is the max line length?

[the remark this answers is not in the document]
erictraut: Good question. I'm not sure if it's a set amount. I guess it would be a fixed max length given that we're using a mono-spaced font. I just know that it overflows and makes the docs difficult to read if the lines get too long.

docs/docs/extensions/virtuallistview.md
```diff
@@ -142,6 +142,14 @@ interface VirtualListViewItemInfo {
 // Should the list animate additions, removals and moves within the list?
 animateChanges?: boolean;
 
+// Cell recycling is an optimization to re-use similar VLV cells & decrease the cost
+// of render operations. However, in some cases, disabling cell recycling can be
+// benificial, recycled cells continue their regular react lifecycle even when not
+// visible, which can lead to excessive background re-rendering in some cases. When
+//  combining VLV with react libraries (like ReSub) that have subscriptions managed
+// by components can cause this behaviour to manifest.
+disableCellRecycling?: boolean;
```
erictraut: I don't think we should add this. It's unnecessary. If you don't want to use cell recycling, simply don't provide an "itemTemplate" field.

berickson1: Template is a required field on VLV

erictraut: Hmm, it wasn't previously. We could switch it back to optional, as it was intended. Or you can just specify an empty string for the template, which will also disable recycling for those items.

berickson1: Looks like since it's OSS introduction it's been required.  I'll make it optional but will leave a comment in documentation similar to the above comment

PR #949, opened 2018-11-28
Add onFocus/onBlur/onKeyPress support to GestureView

berickson1:


src/common/Types.ts
```diff
@@ -788,6 +788,10 @@ export interface GestureViewProps extends CommonStyledProps<ViewStyleRuleSet>, C
     onLongPress?: (gestureState: TapGestureState) => void;
     onContextMenu?: (gestureState: TapGestureState) => void;
 
+    onFocus?: (e: FocusEvent) => void;
```
erictraut: These will require updates to the documentation as well.

PR #957, opened 2018-12-04
Adding "context" mode to Popup handling, and consolidated popup code

deregtd:
To add support for context menus using that reuse the RX.Popup handling support, I added a new positioning type, 'context'.

To make this change easier, I extracted a common set of nearly identical duplicate functions/constants from native-common/PopupContainerView and web/RootView into a single implementation in common/PopupContainerViewBase.

### DIFF
--- a/docs/docs/apis/popup.md
+++ b/docs/docs/apis/popup.md
@@ -19,7 +19,9 @@ Popups are identified by a caller-specified ID that should be unique.
 
 ## Types
 ``` javascript
-type PopupPosition  = 'top' | 'right' | 'bottom' | 'left';
+// 'context' mode makes it attempt to behave like a context menu -- defaulting
+// to the lower right of the anchor element and working its way around.
+type PopupPosition  = 'top' | 'right' | 'bottom' | 'left' | 'context';
 
 interface PopupOptions {
     // Returns a mounted component instance that serves as the
--- a/docs/docs/apis/popup.md
+++ b/docs/docs/apis/popup.md
@@ -20,7 +20,8 @@ Popups are identified by a caller-specified ID that should be unique.
 ## Types
 ``` javascript
 // 'context' mode makes it attempt to behave like a context menu -- defaulting
-// to the lower right of the anchor element and working its way around.
+// to the lower right of the anchor element and working its way around.  It is not supported
+// with inner positioning and will throw an exception if used there.
 type PopupPosition  = 'top' | 'right' | 'bottom' | 'left' | 'context';
 
 interface PopupOptions {
--- a/src/common/PopupContainerViewBase.tsx
+++ b/src/common/PopupContainerViewBase.tsx
@@ -146,6 +146,31 @@ export function recalcPositionFromLayoutData(windowDims: Dimensions, anchorRect:
                         constrainedWidth = anchorRect.left - ALLEY_WIDTH;
                     }
                     break;
+
+                case 'context':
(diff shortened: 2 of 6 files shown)


src/common/PopupContainerViewBase.tsx
```diff
@@ -28,6 +29,249 @@ export interface PopupComponent {
...
+            popupX = anchorRect.left + anchorRect.width - popupWidth;
+            anchorOffset = popupHeight / 2;
+            break;
+
+        case 'right':
+            popupY = anchorRect.top + anchorRect.height / 2 - popupHeight / 2;
+            popupX = anchorRect.left;
+            anchorOffset = popupHeight / 2;
+            break;
+    }
```
erictraut: This case doesn't handle the new 'context' value. I presume it's considered a programming error to specify 'context' with inner position. We should make sure this is documented and doesn't crash.

src/common/Types.ts
```diff
@@ -1025,7 +1025,7 @@ export interface WebViewProps extends CommonStyledProps<WebViewStyleRuleSet> {
     sandbox?: WebViewSandboxMode;
 }
 
-export type PopupPosition  = 'top' | 'right' | 'bottom' | 'left';
+export type PopupPosition  = 'top' | 'right' | 'bottom' | 'left' | 'context';
```
erictraut: Please add new test cases for 'context' in the RXPTestApp.

src/common/PopupContainerViewBase.tsx
```diff
@@ -28,6 +29,249 @@ export interface PopupComponent {
     onHide: () => void;
 }
 
+export interface RecalcResult {
```
erictraut: Thanks for moving this code into common. I should have done before.

docs/docs/apis/popup.md
```diff
@@ -19,7 +19,9 @@ Popups are identified by a caller-specified ID that should be unique.
 
 ## Types
 ``` javascript
-type PopupPosition  = 'top' | 'right' | 'bottom' | 'left';
+// 'context' mode makes it attempt to behave like a context menu -- defaulting
+// to the lower right of the anchor element and working its way around.
+type PopupPosition  = 'top' | 'right' | 'bottom' | 'left' | 'context';
```
erictraut: Can 'context' be used in a list with the other positions, or must it always be by itself as the only requested position?

src/common/PopupContainerViewBase.tsx
```diff
@@ -28,6 +29,249 @@ export interface PopupComponent {
...
+                        absY = anchorRect.top - popupRect.height;
+                    }
+                    break;
+            }
+
+            const effectiveWidth = constrainedWidth || popupRect.width;
+            const effectiveHeight = constrainedHeight || popupRect.height;
+
+            // Make sure we're not hanging off the bounds of the window.
+            if (absX < ALLEY_WIDTH) {
```
erictraut: None of these adjustments are made in the 'context' case. Is this intended? What happens if there is no "perfect fit" and the context menu hangs off the edge of the window? Don't you want to nudge it back on and/or resize it so it fits?

erictraut:
@deregtd, looks like you never addressed my code review comments above. Are you still interested in pursuing this PR?

deregtd:
Sorry for the delay, totally forgot about this.  Addressed all PR comments -- added a sample app entry for context mode, made context mode default to falling back to the other 4 positioning modes if it can't find an exact fit, and added docs about it.

PR #960, opened 2018-12-04
Add native support for stickyHeaderIndicies

berickson1:
Add RN support for StickyHeaderIndicies, update documentation & RXPTest

### DIFF
--- a/docs/docs/components/scrollview.md
+++ b/docs/docs/components/scrollview.md
@@ -62,6 +62,11 @@ overScrollMode?: 'auto' | 'always' | 'never';
 pagingEnabled: boolean = false; // Android & iOS only
 snapToInterval: number = undefined; // iOS only
 
+// Native Only: An array of child indices determining which children get docked to the top of the screen when scrolling.
+// For example passing `stickyHeaderIndices={[0]}` will cause the first child to be fixed to the
+// top of the scroll view. This property is not supported in conjunction with `horizontal={true}`.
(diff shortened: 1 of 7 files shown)


docs/docs/components/scrollview.md
```diff
@@ -62,6 +62,11 @@ overScrollMode?: 'auto' | 'always' | 'never';
 pagingEnabled: boolean = false; // Android & iOS only
 snapToInterval: number = undefined; // iOS only
 
+// Native Only: An array of child indices determining which children get docked to the top of the screen when scrolling.
+// For example passing `stickyHeaderIndices={[0]}` will cause the first child to be fixed to the
+// top of the scroll view. This property is not supported in conjunction with `horizontal={true}`.
+stickyHeaderIndices?: number[];
+
```
a-tarasyuk: @berickson1 I see, that in props were added two new options  `stickyHeaderIndices` and `invertStickyHeaders`. Does make sense to add a mention about `invertStickyHeaders` in ScrollView docs?

erictraut:
I'm not convinced we should add this functionality to RX.ScrollView. It's not aligned with the cross-platform philosophy of ReactXP. The feature is specific to iOS and Android.

How are you making use of this feature in your cross-platform app? Are you applying some workaround for the web implementation?

It seems like there's a (relatively) easy workaround that doesn't require this feature. You can simply render the sticky header(s) as a peer to the scroll view.

PR #961, opened 2018-12-04
Fix a crash when calling setScrollTop/Left on ScrollView

berickson1:
The animated implementation of ScrollView in RN doens't have scrollTop/scrollLeft defined


erictraut:
This is a pretty gross hack, but I don't see a better approach unless/until RN exposes this call for the animated version of RN.ScrollView.

PR #964, opened 2018-12-10
Expose ScrollEventThrottle via VirtualListView

berickson1:
This is useful when driving native animations off of the scroll event. The default of 32 events/s can cause stuttering on the animation


berickson1:
No longer needed - This was also included in VLV re-work

PR #965, opened 2018-12-11
Add zIndex support to ReactXP

berickson1:
Facebook already supports this in React Native based on the CSS spec
https://facebook.github.io/react-native/docs/layout-props#zindex

This is useful in situations where component order must be maintained for accessibility but we want a component to appear in front of others.

### DIFF
--- a/docs/docs/styles.md
+++ b/docs/docs/styles.md
@@ -186,6 +186,7 @@ elevation: number; // Android only
 wordBreak: 'break-all' | 'break-word'; // Web only
 appRegion: 'drag' | 'no-drag'; // Web only
 cursor: 'pointer' | 'default'; // Web only
+zIndex: number;
 ```
 
 ## Transform Style Attributes
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -169,6 +169,7 @@ export interface ViewAndImageCommonStyle extends FlexboxStyle, TransformStyle {
 
     backgroundColor?: string;
     opacity?: number;
+    zIndex?: number;
 }
 
 export interface AnimatedViewAndImageCommonStyle extends AnimatedFlexboxStyle, AnimatedTransformStyle {


erictraut:
I spent a bunch of time investigating this and looking at ways to implement it in a way that is consistent across platforms, and I was ultimately unable to do so. We made the decision not to implement it. See [this discussion](https://github.com/Microsoft/reactxp/issues/600) for details. I don't think anything has changed since this discussion, so I'm inclined to reject this PR.

berickson1:
I don't think there's a way to achieve what I want without z-index.
We have a fixed header in (or absolutely positioned above) a VirtualListView that needs to be pinned in place (like stickyheaderindicies does to ScrollView) but still respect accessibility tab order (header gets tab focus before list items).

If I create the header before the scrollview, the scrollView items are drawn overtop of the header.
If I create the header after the scrollview, keyboard tab order is not respected.

Any other suggestions instead of this?

erictraut:
Can't you override the default keyboard tab order using the tabIndex?

berickson1:
I agree, but in practice most items you render match the focus order (or have additional focus capabilities, like VLV).

berickson1:
This leaves us with 2 potential solutions
1) Maintain a custom fork of reactxp that allows the zIndex prop in this PR
2) Cast the style to `as any` and pass the z-index prop anyways. If RX ever chooses to explicitly sanitize props, this will break this solution.

I'm happy to call this something that indicates there may be divergent behaviour, but we require this prop to be functional in order to make our app keyboard accessible 

In the below example, consider that `MyHeaderBar` has tab-accessible items in it, as does `MyContentList`.  `MyHeaderBar` is designed to float above the content.  Without z-index, `MyContentList` will render over `MyHeaderBar`. If I change the order in the render function, visually things look ok, but tab order will be backwards.

```
const _styles = {
    container: RX.Styles.createViewStyle({
        flex: 1,
        alignSelf: 'stretch',
    }),
    fixedHeader: RX.Styles.createViewStyle({
        position: 'absolute',
        top: 0,
        left: 0,
        right: 0,
    }),
    // zIndex is needed here to preserve render order for accessibility AND keep the header on top of the scroll view
    zIndex: RX.Styles.createViewStyle({
        zIndex: 1,
    }),
};
...
...
            <RX.View style={ [_styles.container] }>
                <RX.View style={ [_styles.fixedHeader, _styles.zIndex] }>
                    <MyHeaderBar />
                </RX.View>
                <MyContentList />
            </RX.View>
```

erictraut:
Yeah, I think the best solution for now is to cast to "any" and manually add zIndex. If/when it breaks or doesn't have the intended action on a particular platform, you'll need to deal with it within your source code. I think that's preferable to making it a confusing option for all ReactXP users.

Issue #906, opened 2018-11-09
TodoList fails to build


tsukhu:
While the other samples compile and run , the TodoList fails to compile
```
 npm run start-web
.....
ERROR in C:\workspace\reactxp\samples\TodoList\src\ts\app\AppBootstrapper.tsx
./src/ts/app/AppBootstrapper.tsx
[tsl] ERROR in C:\workspace\reactxp\samples\TodoList\src\ts\app\AppBootstrapper.tsx(33,42)
      TS2345: Argument of type 'Element' is not assignable to parameter of type 'ReactElement<any>'.
  Types of property 'type' are incompatible.
    Type 'string | ComponentClass<any, any> | StatelessComponent<any>' is not assignable to type 'string | ComponentClass<any> | StatelessComponent<any>'.
      Type 'ComponentClass<any, any>' is not assignable to type 'string | ComponentClass<any> | StatelessComponent<any>'.
        Type 'ComponentClass<any, any>' is not assignable to type 'ComponentClass<any>'.
          Type 'Component<any, any, any>' is not assignable to type 'Component<any, ComponentState, any>'.
            Property 'context' is missing in type 'Component<any, any, any>'.

ERROR in C:\workspace\reactxp\samples\TodoList\src\ts\app\AppBootstrapper.tsx
./src/ts/app/AppBootstrapper.tsx
[tsl] ERROR in C:\workspace\reactxp\samples\TodoList\src\ts\app\AppBootstrapper.tsx(59,13)
      TS2605: JSX element type 'RootView' is not a constructor function for JSX elements.
  Property 'context' is missing in type 'RootView'.

ERROR in C:\workspace\reactxp\samples\TodoList\src\ts\controls\SimpleDialog.tsx
./src/ts/controls/SimpleDialog.tsx
[tsl] ERROR in C:\workspace\reactxp\samples\TodoList\src\ts\controls\SimpleDialog.tsx(150,13)
      TS2605: JSX element type 'Modal' is not a constructor function for JSX elements.
  Property 'context' is missing in type 'Modal'.

ERROR in C:\workspace\reactxp\samples\TodoList\src\ts\views\AccountMenuButton.tsx
./src/ts/views/AccountMenuButton.tsx
[tsl] ERROR in C:\workspace\reactxp\samples\TodoList\src\ts\views\AccountMenuButton.tsx(110,21)
      TS2605: JSX element type 'SimpleMenu' is not a constructor function for JSX elements.
[...]

PR #975, opened 2018-12-17
fix TodoList sample android project

amritbhandari:
upgraded gradle wrapper
upgraded android compile sdk, build tools, gradle plugin
see https://github.com/Microsoft/reactxp/issues/906#issuecomment-447640437


samples/TodoList/android/app/build.gradle
```diff
@@ -127,7 +127,7 @@ android {
 
 dependencies {
     compile fileTree(dir: "libs", include: ["*.jar"])
-    compile "com.android.support:appcompat-v7:23.0.1"
+    compile "com.android.support:appcompat-v7:27.1.1"
```
mikehardy: Why not current, e.g. v28, or even androidx?

samples/TodoList/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Sun Apr 09 20:23:19 PDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
mikehardy: Why not current, e.g. 4.10.2 or 5?

samples/TodoList/android/app/build.gradle
dryganets: He synced the versions with other samples. I support him here as it is easier to install a single set of the dependencies to run all of the samples.
We probably should load the versions of tools and libraries from the properties file instead of defining them in every sample.

samples/TodoList/android/gradle/wrapper/gradle-wrapper.properties
dryganets: Same here. I don't see any problem with upgrading to the latest version of Gradle.
Though I don't see any benefits of it. Whatever version we choose users will end up with different in a month or two.

mikehardy: @dryganets having consistent versions across samples is a solid reason. I was mostly just curious. 4.10.2 has the disk space benefit of cache expiry but otherwise I also agree it's more or less a treadmill. It does seem the react native style these days is have a property and then in build.gradle use the property if it exists or use a defined version if not, so developers can control their own dependencies if desired

PR #976, opened 2018-12-17
Make multiline TextInput growth behaviour on web match RN

berickson1:
On RN, TextInput will automatically grow when new lines are added. Match this behaviour on web instead of adding scrollbars

### DIFF
--- a/src/web/TextInput.tsx
+++ b/src/web/TextInput.tsx
@@ -13,10 +13,12 @@ import * as React from 'react';
 import { FocusArbitratorProvider } from '../common/utils/AutoFocusHelper';
 import { applyFocusableComponentMixin } from './utils/FocusManager';
 import { Types } from '../common/Interfaces';
+import { isEmpty } from './utils/lodashMini';
 import Styles from './Styles';
 
 export interface TextInputState {
     inputValue?: string;
+    autoResize?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


src/web/TextInput.tsx
```diff
@@ -295,6 +321,62 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
...
+            // Iterate across the array and see if there's any height value
+            return combinedStyles.some(style => {
+                if (!style || typeof style === 'number') {
+                    return true;
+                }
+                return style.height === undefined;
+            });
+        } else {
+            return combinedStyles.height === undefined;
+
```
erictraut: Nit: extra carriage return can be removed.

src/web/TextInput.tsx
```diff
@@ -295,6 +321,62 @@ export class TextInput extends React.Component<Types.TextInputProps, TextInputSt
...
+            return false;
+        }
+
+        const combinedStyles = Styles.combine(props.style);
+        if (!combinedStyles || typeof combinedStyles === 'number') {
+            // Number-type styles aren't allowed on web but if they're found we can't decode them so assume not fixed height
+            return true;
+        } else if (Array.isArray(combinedStyles)) {
+            // Iterate across the array and see if there's any height value
+            return combinedStyles.some(style => {
```
erictraut: This check is a bit fragile. There are multiple ways to fixing the height of a view other than setting a "height" attribute. For example, you can specify absolute positioning with a top and bottom value. This is probably sufficient in practice, but it deserves a comment at least.

PR #991, opened 2019-01-17
Update Samples in reactxp that utilize VirtualListView

berickson1:


samples/ImageList/src/views/ImageList.tsx
```diff
@@ -106,7 +106,8 @@ export class ImageList extends ComponentBase<{}, ImageListState> {
         };
     }
 
-    private _renderItem = (item: ImageListItemInfo) => {
+    private _renderItem = (details: VirtualListViewCellRenderDetails<ImageListItemInfo>) => {
```
a-tarasyuk: @berickson1 I think we can use, instead of an additional variable. What do you think?

```typescript
private _renderItem = ({ item }: VirtualListViewCellRenderDetails<ImageListItemInfo>) => {

```

PR #993, opened 2019-01-17
Propagete RN WebView props to RXP

zarubond:
Add missing ReactNative WebView props to ReactXP.

### DIFF
--- a/docs/docs/components/webview.md
+++ b/docs/docs/components/webview.md
@@ -63,6 +63,12 @@ injectedJavaScript: string = undefined; // Native only
 // Is JavaScript executed within the control?
 javaScriptEnabled: boolean = true;
 
+// Determines whether HTML5 audio and video requires the user to tap them before they start playing.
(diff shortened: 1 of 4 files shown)


erictraut:
Thanks for the contribution. In general, we try to avoid adding platform-specific props in ReactXP. In this particular case, I think it's justified because the fallback behavior is reasonable on platforms that don't support it. Before I merge, I'd like to see these props added to the RXPTest sample app, which serves as our unit tests for ReactXP. All supported props should be used at least once in the test app. In this case, there's no way to automatically determine whether the props are having the intended effects, but at least we'll know that their use doesn't cause a compiler error or a runtime crash. You can add it to the existing WebView test (https://github.com/Microsoft/reactxp/blob/master/samples/RXPTest/src/Tests/WebViewBasicTest.tsx).

zarubond:
@erictraut These props are required so the YouTube videos can be played inside webview. I have added the props to the testing app

PR #994, opened 2019-01-17
Remove msHyphens css property

zarubond:
Remove msHyphens default text css style as it is experimental feature which is not used in any other browser and it is breaking design.


a-tarasyuk:
_Note:_ The current status of the specification for [`hyphens`](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens) is `Working Draft`. Most browsers [**support**](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#Browser_compatibility) this CSS property with a vendor prefix, including Edge and IE > = 10.

erictraut:
@zarubond, if I remember correctly, the msHyphens attribute was originally added for Skype. Are you confident that its removal won't break something in Skype?

zarubond:
@erictraut  I have tested it in Skype on Edge without problem and I have tried to find someone who knows why it is there without success. Since this feature was only on Edge I am very confident that it will work just fine.

PR #997, opened 2019-01-18
Updates samples with new RN for Windows version (0.57.0-rc.4)

reseul:
Updated the samples. This may also include:
- Fixing RN compilations due to tools change
- Updating solution/project files to latest format
- Some lint and other fixes

I tested all samples for web & iOS (on a mac), and uwp.

Note: The "run-windows" based samples fail to compile at command line due to the embedded pfx file being password protected (I'm at loss how this is supposed to work in a team).
Loading SLN filies from Visual Studio and compiling/running from there does work though.


samples/hello-world/package.json
```diff
@@ -10,6 +10,7 @@
     "start:ios": "npm run rn-cli -- run-ios",
     "start:web": "cross-env platform=web webpack-dev-server --config=web/webpack/dev.js --progress --colors --mode=development",
     "start:rn-dev-server": "npm run rn-cli -- start --reset-cache",
+    "build": "tsc",
```
a-tarasyuk: Is there a case where this command is needed?

reseul: I couldn't find a way to compile to JS. 
I suspect the run-<platform> may do it, but I can't run "run-windows" (it fails miserably).

a-tarasyuk: are you talking about `npm run start:windows`?

reseul: Yes. The msbuild part of it fails due to the pfx certificate being encrypted with a password, I suspect a password of yours since the file comes from you.
I admit a lot of this msbuild/VS is (non fulfilling) magic to me, so I don't know how to fix. 
Running from within VS works fine.

a-tarasyuk: You need to generate your [self-signed certificate](https://docs.microsoft.com/en-us/powershell/module/pkiclient/new-selfsignedcertificate?view=win10-ps). Basically [metro-react-native-babel-preset](https://github.com/facebook/metro/blob/master/packages/metro-react-native-babel-preset/src/configs/main.js#L16) understands TypeScript due to [@babel/plugin-transform-typescript](https://github.com/facebook/metro/blob/master/packages/metro-react-native-babel-preset/src/configs/main.js#L173), and some of example use this feature.

reseul: It's not covered in the readme's, hence my confusion (assuming I'm doing something wrong).
I'll see what I can do on Monday, today I'm past the quota allocated for this. Thanks.

a-tarasyuk: Also, there is command to run dev server - `npm run start:rn-dev-server`

reseul: True, I used that.

reseul: Tried with local self signed certs and all three projects that use "npm run start:windows" (hello-world, hello-world-js, ImageList) work correctly from command line in Windows now.

samples/ImageList/tsconfig.json
```diff
@@ -16,7 +16,7 @@
     "noImplicitAny": true,
     "skipLibCheck": true,
     "declaration": true,
-    "jsxFactory": "React",
+    "jsxFactory": "React.createElement",
```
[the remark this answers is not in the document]
reseul: This change seems to have been necessary when I was using the "tsc" workaround @a-tarasyuk was complaining about. The TS compiler would generate:
    `return React(RootView, null)`
instead of
    `return React.createElement(RootView, null)`

Without the workaround it seems the babel/whatever pipeline doesn't take this jsxFactory property from tsconfig.json into account.
I could change back, but it seems the new version is more correct since it helps those using tsc. I admit I don't know why the setting is there in the first place.

I did test start:ios, start:web, and now start:windows, all with good results. I just didn't do Android cause I'm avoiding that one.

PR #1007, opened 2019-01-28
Perf improvements to VLV

berickson1:
Use ES6 Map & Set instead of raw JS objects, these perform faster (https://jsperf.com/javascript-objects-vs-map-performance)
Use for..of instead of _.each, the transpiled output becomes a for loop, which is also faster than lodash (https://jsperf.com/jquery-each-vs-underscore-each-vs-for-loops/114)


erictraut:
Is this just a theoretical speedup, or does it make a noticeable (or at least measurable) difference? Compared to re-rendering costs, I'd expect the cost of doing map lookups in the VLV should be in the noise.

berickson1:
Running some quick perf tests with long fixed-height VLV, page script time (load + first render) went from avg of 299ms to avg 284ms dropping top and bottommost sample on both. (~5% faster). Scrolling isn't impacted by this (less than 1% change in all experiments)

Additionally Map/Set gives us proper return types, we typically don't annotate |undefined on our dictionary types, so we can pull an undefined value out and typescript would be unaware.  Honestly I find this to be more readable than the previous version. Additionally the usage of the delete on object properties puts the object into hash-mode in some JS engines, which is generally slower. (again, not a huge win either way)

erictraut:
This is a lot of churn (regression risk) for relatively small benefits. I reviewed the code, and I don't see any obvious problems, but it's a big enough change that there's statistically a good chance that there's a regression in there somewhere. I personally don't find this any more readable, but that's subjective. A 5% perf gain isn't nothing, but it's also pretty minimal. I could go either way on this one. @deregtd, any thoughts on this?

deregtd:
After finding out the HUGE perf benefits in tight loops to avoiding using lodash, we're trying to more or less remove most (or possibly all) lodash from all of our codebases.  If we don't remove it from all of our dependencies too, then we don't get the benefit of excising dozens of KB of code from our apps.  Also, while there's technically regression risk, the changes are more or less regex replacements, so I'm not as worried about it as I would otherwise be.

Issue #1013, opened 2019-01-31
Build test fails


dagatsoin:
I am trying to run the test like it is explained in https://github.com/dagatsoin/reactxp/blob/master/CONTRIBUTING.md but I have this error 
```
ERROR in [at-loader] ./node_modules/reactxp/dist/common/Types.d.ts:224:65
    TS2694: Namespace 'React' has no exported member 'RefAttributes'.
```

Here is the full log:
```
MBP-de-Fraktar:RXPTest fraktar$ npm run web-watch

> rxptest@1.3.0 web-watch /Users/fraktar/dev/reactxp/samples/RXPTest
> webpack --progress --colors --watch

 10% building 0/0 modules 0 active
webpack is watching the files…

 10% building 0/1 modules 1 active ...v/reactxp/samples/RXPTest/src/index.tsxℹ ｢atl｣: Using typescript@3.2.4 from typescript
ℹ ｢atl｣: Using tsconfig.json from /Users/fraktar/dev/reactxp/samples/RXPTest/tsconfig.json
 94% after sealℹ ｢atl｣: Checking started in a separate process...
✖ ｢atl｣: Checking finished with 1 errors
Hash: 6325074bcc3f4eed006b
Version: webpack 4.29.0
Time: 11056ms
Built at: 31/01/2019 10:02:41
        Asset      Size  Chunks             Chunk Names
    bundle.js  2.43 MiB    main  [emitted]  main
bundle.js.map  2.89 MiB    main  [emitted]  main
Entrypoint main = bundle.js bundle.js.map
[./node_modules/webpack/buildin/global.js] (webpack)/buildin/global.js 472 bytes {main} [built]
[./node_modules/webpack/buildin/module.js] (webpack)/buildin/module.js 497 bytes {main} [built]
[./src/App.tsx] 2.59 KiB {main} [built]
[./src/CommonStyles.tsx] 616 bytes {main} [built]
[./src/Test.tsx] 317 bytes {main} [built]
[./src/TestContainer.tsx] 8.73 KiB {main} [built]
[./src/TestListView.tsx] 6.17 KiB {main} [built]
[./src/TestRegistry.ts] 4.7 KiB {main} [built]
[./src/Tests/AccessibilityTest.tsx] 5.49 KiB {main} [built]
[./src/Tests/ActivityIndicatorTest.tsx] 4.33 KiB {main} [built]
[./src/Tests/AlertTest.tsx] 6.38 KiB {main} [built]
[...]

PR #1014, opened 2019-01-31
Add missing @types/react in test package

dagatsoin:
Issue #1013


berickson1:
Once the next version of reactxp is published, this will be un-necessary since reactxp defines @types/react as a dependency (in which the version recently changed)
This only occurs since RXPTest relies on possible unpublished reactxp - any time dependencies changes, this may happen.

dagatsoin:
Fair enough. Closing 👍

Issue #1015, opened 2019-01-31
[NATIVE] ViewProps.onStartShouldSetResponder is not trigger when onPress || longPress is set


dagatsoin:
## Description

If `ViewProps.onPress` or `ViewProps.onLongPress` is set as long as `ViewProps. onStartShouldSetResponder`, `onStartShouldSetResponder` will not be triggered.

## Repro

```
   <>
       <RXView
          onStartShouldSetResponder={() => {
            console.log("will be trigger")
            return true
          }}
        >
        <RXView
          onPress={() => {}}
          onStartShouldSetResponder={() => {
            console.log("should be trigger but won't")
            return true
          }}
        />
     </>
```

dagatsoin:
I think it is because, this line does not trigger the passed callback but just the default RN handler
https://github.com/Microsoft/reactxp/blob/master/src/native-common/View.tsx#L379

if should be:
```
onStartShouldSetResponder:  this.props.onStartShouldSetResponder || this.touchableHandleStartShouldSetResponder
```

I wonder if it could solve #1012

PR #1016, opened 2019-01-31
[FIX-1015] View responder events props are now triggered properly

dagatsoin:
## Description

The responder event handlers props passed to a view are not triggered when onPress or onLongPress is also present.

## This PR does the following

- Assign the events handlers props if present or assign the default RN handler.
-  Add tests

More infos: #1015


dagatsoin:
Fixed test, ready for review & QA

PR #1020, opened 2019-02-05
Declare ACCESS_NETWORK_STATE permission

P-Blackburn:
The API - Network test requires that the ACCESS_NETWORK_STATE permission is declared -( without it the test errors.)


mikehardy:
I noticed this too but just manually gave permission instead of sending a PR.

PR #1021, opened 2019-02-05
Build Caveats - for unpublished dependencies

P-Blackburn:
Given RXPTest by design exercises ReactXP it can have transient dependencies on as yet unpublished changes. The proposed changes to the README.md gives a pointer on how to address these.


erictraut:
Thanks for the contribution!

PR #1024, opened 2019-02-07
Add more mouse cursor type into GestureMouseCursor

stdavidchen:
Hi, I would like to add more mouse Cursor type support into ReactXP.

All new cursor is coming from and tested on both electron and chrome.
https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

Thanks,
/Wei


erictraut:
Thanks for your contribution. Looks good to me.

stdavidchen:
Thanks for reviewing.

PR #1041, opened 2019-03-05
Fix textDecorationStyle and textDecorationColor on web

sbeca:
Tested the web version of RXPTest - Components - Text on:
- macOS 10.14.3
- Chrome 72.0.3626.121

Before change:
screen shot 2019-03-05 at 9 48 37 pm

After change:
screen shot 2019-03-05 at 9 48 54 pm


erictraut:
Thanks for the contribution. Looks good to me. Merging.

PR #1042, opened 2019-03-06
Fix transform rotate styles not working on web

sbeca:
React Native platforms expect transform rotate styles to be expressed with a string in either degrees (deg) or radians (rad), e.g. `transform: [{ rotateZ: '45deg' }]`. If you just try to pass a number, e.g. `'45'`, you will get an error like:
`Invariant Violation: Rotate transform must be expressed in degrees (deg) or radians (rad): {"rotateZ":"45"}`

However, the web styles code would always add 'deg' to the end of the string so you would end up with CSS like `transform: rotateZ(45degdeg)` which obviously won't work.

This PR removes the code that adds that extra 'deg'.

### DIFF
--- a/src/web/Styles.ts
+++ b/src/web/Styles.ts
@@ -305,9 +305,7 @@ export class Styles extends RX.Styles {
                         animatedTransforms[key] = t[key];
                     } else {
                         let value: string = t[key].toString();
-                        if (key.indexOf('rotate') === 0) {
-                            value += 'deg';
-                        } else if (key.indexOf('translate') === 0) {
+                        if (key.indexOf('translate') === 0) {
                             value += 'px';
                         }
 


deregtd:
I'm not 100% sure on this, but I believe that all of these are specified in reactxp styles as numbers, and reactxp adds the unit.  This way for react native, it can just use the number.

erictraut:
Yeah, I'm going to need to review the native and web code to make sure this change doesn't introduce a regression. We should also have a test case in the RXPTest app that covers this case.

sbeca:
@deregtd: TransformStyle is defined as this:
```typescript
export interface TransformStyle {
    transform?: {
        perspective?: number;
        rotate?: string;
        rotateX?: string;
        rotateY?: string;
        rotateZ?: string;
        scale?: number;
        scaleX?: number;
        scaleY?: number;
        translateX?: number;
        translateY?: number;
    }[];
}
```

The problem from my testing is the the ReactXP native platforms expect rotate values to include `deg`/`rad` as part of the string whereas web doesn't, so I can only get both native and web platforms to work if I create a style like this:
```typescript
headerBackground: RX.Styles.createViewStyle({
    ...RX.Platform.select({
        web: {
            transform: [
                { rotateZ: '-2' }
            ]
        },
        default: {
            transform: [
                { rotateZ: '-2deg' }
            ]
        }
    })
})
```

erictraut:
I confirmed that this was a long-standing bug in reactxp. Thanks for the fix!

PR #1046, opened 2019-03-09
RXPTest - Adjust UserInterfaceTest number comparison

sbeca:
Running the RXPTest results in an error within the APIs - UserInterface test when run with the following set up:
- macOS 10.14.3
- Xcode 10.1
- Building for iPhone X simulator
- Building using command `cd ./samples/RXPTest/ && npm i && npm run ios`

When running the test I got these errors:
Screenshot1

The first error is expected on some older versions of React Native but the second error isn't. So I expanded the error to see what the actual values where and got this:
Screenshot2

So to me that just looks like a float rounding comparison issue that shouldn't actually fail the test. So this PR adds a `_approxEquals` function that compares the difference against an epsilon value to account for rounding errors. This PR also keeps my expanded error logs as I think they could help for future bug hunting. Using these code changes, I now get this:
Screenshot3

### DIFF
--- a/samples/RXPTest/src/Tests/ImageApiTest.tsx
+++ b/samples/RXPTest/src/Tests/ImageApiTest.tsx
@@ -7,6 +7,7 @@ import RX = require('reactxp');
 
 import * as CommonStyles from '../CommonStyles';
 import { AutoExecutableTest, TestResult, TestType } from '../Test';
+import { approxEquals } from '../Utilities';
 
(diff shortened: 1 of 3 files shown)


sbeca:
FYI, I pushed a second commit to this PR to help identify the problem that I reported in issue #1047.

erictraut:
Thanks, looks good to me. Merging.

PR #1048, opened 2019-03-10
Update dependencies

sbeca:
Update all the dependencies of the main package and the RXPTest sample.

I have pinned `react` to exactly version `16.6.3` and `react-native` to exactly version `0.58.6` for better reassurance of compatibility. `react-native` always seems to specify an exact version of `react` as a dependency, which I assume is so they can have an easier time providing capability between the two libraries. As of writing this, the latest version of `react-native` is `0.58.6`, which has a dependency of exactly `16.6.3` for `react` so that is what I've included. It seems to me like following how `react-native` does dependencies is a good way to go about things but if the core ReactXP maintainers have other thoughts, I'm keen to hear.

I also needed to add `android:usesCleartextTraffic="true"` to the RXPTest AndroidManifest to make Metro Bundler work on Android 9.0+ because Android 9.0+ doesn't allow HTTP by default, only HTTPS, but Metro Bundler uses HTTP.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,19 +5,19 @@
   "requires": true,
   "dependencies": {
     "@types/lodash": {
-      "version": "4.14.121",
-      "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.14.121.tgz",
-      "integrity": "sha512-ORj7IBWj13iYufXt/VXrCNMbUuCTJfhzme5kx9U/UtcIPdJYuvPDUAlHlbNhz/8lKCLy9XGIZnGrqXOtQbPGoQ=="
+      "version": "4.14.122",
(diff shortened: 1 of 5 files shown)


mikehardy:
Nice! the cleartext traffic thing is vital for apps that target API28, yes and this is the way I'd do it (as someone coming from the android side) for an example - just right in the manifest. For real apps you want to have it only in the debug manifest, included for API>=24 but that doesn't apply here. I just checked the corresponding Info.plist for the ios sample RXPTest and it already has the corresponding change (allow insecure loads) - I think because iOS started enforcing the similar requirement sooner. So this looks good for development on both platforms.

Might as well do gradle plugin 3.3.2 while you are in the example taking everything to stable/current?

I haven't had any problems with react versions higher than react-native but I also agree with something that is such basic infrastructure like ReactXP pinning the exact same versions they do.

erictraut:
Thanks! Looks good. Merging.

PR #1055, opened 2019-03-14
[RN59] Make mixin application more robust View, Button

fbartho:
In react-native 59, Facebook added a new arbitrary object into the Touchable mixin.
Reference: https://github.com/facebook/react-native/blame/master/Libraries/Components/Touchable/Touchable.js#L904

This causes the applyMixin methods to crash.

This commit makes the applicators skip the mixin’s attribute if it’s not bindable, restoring support for react-native 59. (This is as far as I've getting debugging, however!)

### DIFF
--- a/src/native-common/Button.tsx
+++ b/src/native-common/Button.tsx
@@ -54,7 +54,7 @@ function noop() { /* noop */ }
 
 function applyMixin(thisObj: any, mixin: {[propertyName: string]: any}, propertiesToSkip: string[]) {
     Object.getOwnPropertyNames(mixin).forEach(name => {
-        if (name !== 'constructor' && propertiesToSkip.indexOf(name) === -1) {
+        if (name !== 'constructor' && propertiesToSkip.indexOf(name) === -1 && typeof mixin[name].bind === 'function') {
             assert(
                 !(name in thisObj),
                 `An object cannot have a method with the same name as one of its mixins: "${name}"`
--- a/src/native-common/View.tsx
+++ b/src/native-common/View.tsx
@@ -44,7 +44,7 @@ function noop() { /* noop */ }
 
 function applyMixin(thisObj: any, mixin: {[propertyName: string]: any}, propertiesToSkip: string[]) {
     Object.getOwnPropertyNames(mixin).forEach(name => {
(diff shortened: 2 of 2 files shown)


mikehardy:
For what it's worth, this seems to work great for me as I move to RN0.59.1 (which also seems to work fine)

I would love to see a new RC cut with this change so I don't have to do odd pre-rendering stuff for types (`cd node_modules/reactxp && npm run build`) while depending on github master to get this change :pray: :smile:

erictraut:
I’ll publish a new RC when I have a spare moment — probably this evening.

mikehardy:
Thanks - As ever, I appreciate the work

fbartho:
As an aside: @mikehardy, That's a super useful workaround. I'm thrilled that ReactXP chooses to ship everything you need to run reactxp straight from github master. Each time I've had to debug & patch transpiled node_modules code has been such a pain with other libraries.

❤️ reactxp, and thank you @erictraut!

erictraut:
Sorry I wasn't able to publish last night. I'm traveling for work right now. I'll get to it this evening. Sorry for the delay.

erictraut:
I just published reactxp 1.6.0-rc.4 with this fix. Enjoy!

mikehardy:
Just for feedback: I have 1.6.0-rc.4 "successfully integrated" with my app now alongside RN0.59.1 and everything seems fine, for what it's worth. I use scare quotes there because it isn't the most complicated app, and it isn't launched yet so my testing pool is small but still on a full test run-through of the app zero regressions is zero regressions. Thanks!

fbartho:
I hope to go to production with this rc & React native 59 in about a week. So far no other problems!

erictraut:
If rc.4 is looking good, this might be a good time to publish the final 1.6.0. Sounds like that would work well for you too @fbartho?

fbartho:
No objections from my side! We had already been on rc.2 in our master trunk for a week or so. No concerns were found there for us!

erictraut:
OK, I'll publish 1.6.0 and create a servicing branch. That way, we can cherry-pick future fixes on a stable branch.

PR #1060, opened 2019-03-19
Fix Ref<T> type

Thristhart:
The previous typing was intended to be:
```typescript
 (instance: T | null) => void |
 RefObject<T> |
 null;
```

But because of a lack of parenthesis, the actual type was:
```typescript
 (instance: T | null) => (void | RefObject<T> | null);
```
which breaks the type if you're using RefObject. 

This could be fixed by adding parenthesis like so:
```typescript
 ((instance: T | null) => void) | RefObject<T> | null;
```
But instead I copied the existing type from @types/react which also includes a hack to enable bivariance on callback refs.

### DIFF
--- a/src/common/Types.ts
+++ b/src/common/Types.ts
@@ -355,7 +355,7 @@ export type ComponentBase = React.Component<any, any>;
 interface RefObject<T> {
     readonly current: T | null;
 }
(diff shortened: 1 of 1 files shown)


erictraut:
Ah, good catch! Thanks for the contribution.

PR #1061, opened 2019-03-20
Guard against secureTextEntry bug in Android React Native

sbeca:
I'm not sure how the ReactXP maintainers feel about guarding against bugs in React/React Native within the ReactXP codebase but in case you are okay with it, here is a PR that guards against a bug I ran into today that is in the React Native codebase.

The React Native bug is: https://github.com/facebook/react-native/issues/20606


erictraut:
I have mixed feelings about this. It could be a slippery slope if we start adding work-arounds to RN bugs in ReactXP. And ReactXP is meant to work with a broad range of RN versions, so it's not clear when it would be feasible to remove such a fix. On the other hand, I see the value to this. I'm interested in other opinions... @deregtd, @berickson1?

deregtd:
I'm not a huge fan of guarding against RN bugs that they will presumably fix someday...  If anything, only enable the guard/warning in dev builds of RXP so it's completely gone in prod?

erictraut:
After some discussion, I think we'll pas on this change.

PR #1062, opened 2019-03-20
Fix TextInput defaultValue not working on native platforms

sbeca:
The TextInput defaultValue prop was never being used on native platforms.

This PR updates the constructor initial state setting code to match the web code that sets the initial inputValue state.


erictraut:
Good fix! Thanks.

PR #1063, opened 2019-03-20
Fix Navigation extension on web after commit 3849346b54357894e681c09c398613b1e979c806

### LINKED ISSUE #1057: reactxp-navigation v1.6.0 broken on web
[v1.6.0](https://github.com/Microsoft/reactxp/commit/3849346b54357894e681c09c398613b1e979c806) of reactxp-navigation made some changes to the web implementation of Navigator that broke navigation for me.

For example, running the hello-world sample on macOS 10.14.3, in Chrome 73.0.3683.75, is broken for me in the following 2 ways:
1. Pressing the `See More Examples` button on the first screen successfully navigates to the second screen but I don't see the `FloatFromRight` transition, instead the second screen just instantly appears.
2. Pressing the `Go Back` button on the second screen doesn't navigate to the first screen, instead just wiping the entire screen and then presenting nothing but an empty page.

sbeca:
Commit 3849346b54357894e681c09c398613b1e979c806 changed some usage of `this.refs` to `this._sceneRefs` but not all, leading to reference tracking getting out of sync, which causes navigation between scenes to not always work correctly.

This PR changes a couple of remaining `this.refs`, which fixes #1057

### DIFF
--- a/extensions/navigation/src/web/Navigator.tsx
+++ b/extensions/navigation/src/web/Navigator.tsx
@@ -435,8 +435,9 @@ export class NavigatorImpl extends NavigatorBase<NavigatorState> {
             (enabledSceneNativeProps.style as any).opacity = 0;
         }
 
-        if (this.refs['scene_' + sceneIndex]) {
-            this._setNativeStyles(this.refs['scene_' + sceneIndex], enabledSceneNativeProps.style);
+        const sceneId = 'scene_' + sceneIndex;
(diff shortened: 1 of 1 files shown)


erictraut:
Thanks for looking into this and fixing it. It was on my to-do list for next weekend. :)

I'll back-port this to the 1.6.x branch and publish an update.

PR #1066, opened 2019-03-21
Fix video extension to support react-native local assets.

### LINKED ISSUE #1065: reactxp-video doesn't support native resources
Hi,

The video extension doesn't support `react-native` local assets which are resolved by `require()`. Assets resolved by `require` resolve to number.

This source line:

```javascript
source={ { uri: this.props.source } })
```
(https://github.com/Microsoft/reactxp/blob/master/extensions/video/src/native-common/Video.tsx#L42)

Naive fix i.e changing it to:
```
source: typeof this.props.source === 'number' ? this.props.source : { uri: this.props.source }
```
solves the issue.

PR coming

zbigg:
In react-native local assets resolved by `require()` resolve to number. They should be passed to `react-native-video` as is.

Attempt to fix #1065.

As for now, tested on iOS target.

### DIFF
--- a/docs/docs/extensions/video.md
+++ b/docs/docs/extensions/video.md
@@ -85,8 +85,8 @@ resizeMode: 'contain'|'cover'|'stretch' = 'contain';
 // Displays controls for play, pause, etc.
 showControls: boolean = false;
 
-// Source of video (URL)
-source: string;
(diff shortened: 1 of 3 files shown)


erictraut:
Thanks for the contribution! Looks good to me. Merging.

PR #1073, opened 2019-03-31
Use react-native template standard for API28 dev bundle

mikehardy:
This solves the problem of the dev bundle for the samples not loading when testing on API28 (Android Pie) devices. Without this, testing on API28 is difficult.

This is the upstream version of the template file

https://github.com/facebook/react-native/blob/master/template/android/app/src/debug/AndroidManifest.xml


erictraut:
Thanks for the contribution.

Issue #1074, opened 2019-04-01
RX.Button TypeError: mixin[name].bind is not a function on RN 0.59


luisnaranjo733:
I am seeing a bunch of app crashing errors from RX.Button when migrating from 0.58.6 to 0.59.2. I can replace all instances of RX.Button with RN.Button and everything else seems to be working fine.

Specifically, it appears that this [applyMixin fn](https://github.com/Microsoft/reactxp/blob/master/src/native-common/Button.tsx#L55) is causing problems

My current theory is that this is related to one of the [bind breaking changes](https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#breaking-changes-) announced in RN 0.58. This wasn't happening previously, when I migrated from 0.57.8 to 0.58.6. It only started happening when I migrated from 0.58.6 to 0.59.2. I guess it could mean that the breaking changes didn't actually go into 0.58 as advertised, but they did go in for 0.59

ExceptionsManager.js:74 TypeError: mixin[name].bind is not a function

[Stack trace part 1](https://i.imgur.com/kWMUzG4.png)
[Stack trace part 2. This one has the reference to applyMixin which aligns with the exception name](https://i.imgur.com/WfgNDWT.png)

luisnaranjo733:
I see that this issue has been fixed in a future version of XP.
I opened a PR and realized that my fix was redundant

https://github.com/Microsoft/reactxp/pull/1075

PR #1075, opened 2019-04-01
Fix RX.Button crash due to failed mixing

luisnaranjo733:
After migrating to RN 0.59.2, I am seeing the issue described [here ](https://github.com/Microsoft/reactxp/issues/1074) crash the app on launch (reactxp@1.3.0-rc.6).

I am seeing that the applyMixin fn is failing to bind when
`name=== "withoutDefaultFocusAndBlur"`
and
`mixin[name] = { "withoutDefaultFocusAndBlur": {} }.`

Checking if the bind fn exists on the mixin[name] prototype before trying to bind fixes the issue


luisnaranjo733:
Abandoning my PR because I see now that this has already been fixed in a future version of XP

PR #1078, opened 2019-04-15
[TS] Improve dev experience on mobile

dagatsoin:
While fixing #1077 I encountered various issues during my development experience:
- can't run index.html on mobile!
- Test app is zoomed out
- Long press triggers magnifying glass on iOS

This PR does the following:
- add `webpack-dev-server` to serves the index.html
- add instruction to run test on mobile
- add meta tag to zoom issue on mobile
- makes app test buttons unselectable

PR #1079, opened 2019-04-15
[FIX] onTouchMove on mobile web does not trigger

dagatsoin:
**Depends on #1078** for testing on mobile.

**Issue:**
View.onTouchMove handler is not available on mobile web.

This commit does the following:
- bind props to div.onTouchMove handler
- add test to ViewTouch

**Note**
Weird lint error forces me to force cast the [TouchEvent handler](https://github.com/Microsoft/reactxp/pull/1079/files#diff-cb2be5fd7674a7551341c483786ef131R361)

Full error: 
```
Type '((e: TouchEvent) => void) | undefined' is not assignable to type '((event: TouchEvent<any>) => void) | undefined'.
  Type '(e: TouchEvent) => void' is not assignable to type '(event: TouchEvent<any>) => void'.
    Types of parameters 'e' and 'event' are incompatible.
      Type 'TouchEvent<any>' is not assignable to type 'TouchEvent'.
        Types of property 'changedTouches' are incompatible.
          Type 'React.TouchList' is not assignable to type 'import("/Users/***/dev/reactxp/src/common/Types").TouchList'.
            Types of property 'item' are incompatible.
              Type '(index: number) => React.Touch' is not assignable to type '(index: number) => import("/Users/***/dev/reactxp/src/common/Types").Touch'.
                Type 'Touch' is missing the following properties from type 'Touch': locationX, locationY
```

### DIFF
--- a/samples/RXPTest/src/Tests/ViewTouchTest.tsx
+++ b/samples/RXPTest/src/Tests/ViewTouchTest.tsx
@@ -46,6 +46,14 @@ const _styles = {
         backgroundColor: '#eee',
         borderColor: 'black',
     }),
+    testContainer4: RX.Styles.createViewStyle({
+        flex: 1,
+        height: 150,
+        margin: 20,        
+        borderWidth: 1,
+        backgroundColor: '#eee',
+        borderColor: 'black',
+    }),
     success: RX.Styles.createViewStyle({
         borderWidth: 2,
         backgroundColor: 'green'
@@ -59,6 +67,10 @@ interface TouchViewState {
     view3TouchResponderTestStart: boolean;
     view3TouchResponderTestGrant: boolean;
     view3TouchResponderTestRelease: boolean;
+    touchPositionOnPage: {
+        x: number | null
+        y: number | null
+    };
     nestedViewTouchTestParent: boolean;
     nestedViewTouchTestChild: boolean;
     pressEvent?: RX.Types.TouchEvent;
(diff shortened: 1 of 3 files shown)


erictraut:
Adding onTouchMove to the public interface of RX.View isn't in keeping with the general philosophy of ReactXP because it's platform-specific (notably, the web). We should consider alternative approaches such as:
1. Subscribing to onTouchMove internally to the web implementation but calling the general onMouseMove at the RX.View layer.
2. Implementing onTouchMove for RX.View in all platforms.

dagatsoin:
Thanks for the precision of the general philosophy.

Regarding the alternative approches:
1- onMouseMove does not fit IMO because it misses important information like `targetTouches`
2- implementing onTouchMove could be redondant with the `responder` system. But that make think to another option

=> Bind onTouchMove to the onResponderMove:
1- the signature is the same `(e: TouchEvent) => void`
2- the API surface won't change.
3- we keep in the general philosophy

dagatsoin:
Just made the modification, waiting for your feedback.

dagatsoin:
To add to the onResponder cause: I am facing a new case where an other missing web touch handler (`onTouchEnd`) is missing.

Here is the current result. It is a Fast Action Button.

image

It works this way:

1- I long press the bottom right FAB
2- its options are displayed.
3- with the same touch I will over one of the options. All the options are included in a view which detects, on each touch move, which option is under the touch.
4- on the touch release, I will trigger the target option onPress handler.

This case works with `onTouchMove` and `onTouchEnd` handlers which are not currently supported.

So I already need to implement two new handlers. And I guess some others in near futur.
Regarding the implementation options, I think a good requirements is to keep the minimum surface API.
So I would argue that binding the web touch handlers to the responder system is a good way to take.

-`onTouchStart` => `onResponderStart`
-`onTouchMove` => `onResponderMove`
-`onTouchEnd` => `onResponderEnd`
-`onTouchCancel` => `onResponderTerminate `

If you are agree I will update this PR this way.

erictraut:
Yeah, this solution looks good to me. Thanks.

erictraut:
I jumped the gun on merging your initial PR for onTouchMove. Your plan above sounds good. Please submit a new PR with the other three responder callbacks bound as well.

PR #1080, opened 2019-04-15
[FIX] delayLongPress has no effect on web

dagatsoin:
**Issue:**
ViewProps.delayLongPress has no effect on web.

This commit does the following:
- assign ViewProps.delayLongPress to the _longPressTimer

**QA**
- change the delay in the ButtonTest long press test and run the test
- the counter should increase after the right delay


erictraut:
Looks good. Thanks.

PR #1081, opened 2019-04-16
[FIX] Web mobile: can't press after a long press

dagatsoin:
## Issue
**Current behavior**
User cannot press button again just after a long press. The user mush do a double tap as a workaround.
This can be reproducted on the current test app on the long press button test on web mobile:
1- do a long press, the "Long press count" increases
2- do a press, the "Press count" does not change
3- do a press again, the "Press count" increases.

**Expected behavior**
1- do a long press, the "Long press count" should increase
2- do a press, the "Press count" should increase

## Fix
This PR does the following.
- reflect the behavior of desktop: the onMouseUp event after a longPress is canceled and the _ignoreClick flag is reset.

## QA
1- do a long press, the "Long press count" should increase
2- do a press, the "Press count" should increase

### DIFF
--- a/src/web/Button.tsx
+++ b/src/web/Button.tsx
@@ -217,6 +217,10 @@ export class Button extends ButtonBase {
     }
 
     private _onMouseUp = (e: Types.SyntheticEvent) => {
+        if (this._ignoreClick) {
+            e.stopPropagation();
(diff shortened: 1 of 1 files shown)


erictraut:
Looks good. Thanks!

PR #1082, opened 2019-04-16
[TS] Bind remaining web touch handlers

dagatsoin:
**Depends on #1083**

**Following PR #1079**

This PR does the following
- bind onTouchStart to ViewProps.onResponderStart
- bind onTouchEnd to ViewProps.onResponderRelease
- bind onTouchCancel to ViewProps.onResponderTerminate
- let the  onTouchEnd of a long press bubbles up
- add a test in ButtonTest.tsx

PR #1083, opened 2019-04-17
[FIX] Web: prevent triggering longPress when pointer is outside the button

dagatsoin:
This PR fixes a bug which allows the long press handler to be triggered while the pointer has left the button.

**Current behavior**
1- The user presses a button and leaves it during the long press delay.
2- The counter still increases.

bug-longpress

**Expected behavior**
1- The user presses a button and leaves it during the long press delay.
2- The counter does not increase.

This PR does the following:
- web desktop: clear the long press timer when the mouse leaves the button
- web mobile: clear the long press timer when the touch leaves the button

### DIFF
--- a/src/web/Button.tsx
+++ b/src/web/Button.tsx
@@ -104,6 +104,7 @@ export class Button extends ButtonBase {
                 aria-checked={ ariaChecked }
                 onClick={ this.onClick }
                 onTouchStart={ this._onMouseDown }
+                onTouchMove={ this._onTouchMove }
                 onTouchEnd={ this._onMouseUp }
                 onContextMenu={ this._onContextMenu }
                 onMouseDown={ this._onMouseDown }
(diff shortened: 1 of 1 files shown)


dagatsoin:
@erictraut ready for review

dagatsoin:
Found a case where this does not work:
- trigger a longpress
- go outside
- trigger the release on another element

I will investigate further.

dagatsoin:
Done @erictraut 

Here some explanations of the new rules to spot eventual regressions or breaking changes.

## Changes for mouse handlers for Web platform

The long press is no longer triggered when the touch has left the button.

## New rules for touch handlers for Web platform

### Case where onPressOut is not triggered and the event propagation is stoped:
- Long press > release

### Case where onPressOut is not triggered and the event propagation continues:
- Press in > leave button > release touch

### Cases where onPressOut is triggered and the event propagation continues:
- Long press > leave button > release touch
- Tap

There is a test in #1082 which demonstrate the rules. I have not put the test in this PR, because it requires some changes which belong to #1082

erictraut:
Thanks for the contribution! This looks good to me.

PR #1091, opened 2019-05-07
[FIX] Web touch: Prevent ButtonProps.onPressIn to trigger twice

dagatsoin:
## Problem: 
On web platform with a touch screen`onPressIn` is called twice.

## Repro:
- test at the **Button with press** on a mobile device
- the PressIn count increase by 2

## This PR does the following 
- prevent `onMouseDown` to trigger after `onTouchStart`

## QA
- test at the **Button with press** on a mobile device
- the PressIn count increase by 1

### DIFF
--- a/src/web/Button.tsx
+++ b/src/web/Button.tsx
@@ -267,6 +267,10 @@ export class Button extends ButtonBase {
      * 4- Press in > leave button > release touch
      */
     private _onTouchEnd = (e: Types.SyntheticEvent | Types.TouchEvent) => {
+        if ('touches' in e) {
+            // Stop the to event sequence to prevent trigger button.onMouseDown
+            e.preventDefault();
+        }
         if (this._isMouseOver && this._ignoreTouchEnd) {
             /* 1 */
             e.stopPropagation();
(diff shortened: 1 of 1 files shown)


src/web/Button.tsx
```diff
@@ -287,6 +287,13 @@ export class Button extends ButtonBase {
         if (this._longPressTimer) {
             Timers.clearTimeout(this._longPressTimer);
         }
+        if ('touches' in e) {
```
berickson1: I think this will cause a regression in button behaviour - touch down -> move -> touch up will trigger an on press, which isn't consistent with previous behaviour or react native

dagatsoin: Sorry for the reply my schedule does not allow me to handle this this week. Let me play with the test app to fully understand your concern. I will reply next week :)

dagatsoin: Got it!
I even caught the bug in https://github.com/dagatsoin/sproutch where a tab is triggered while the cursor is already out.

I just pushed the fix.

dagatsoin:
rebased on master

src/web/Button.tsx
berickson1: I assume you've tried this with both mouse and touch?
The change looks reasonable but I haven't actually tried it locally :)

dagatsoin: Indeed, but do the QA plz, I don't won't to be responsible of the next Skype collapse :p

PR #1059, opened 2019-03-18
Improve support for React Native v0.59.x

sbeca:
EDIT: Some changes have been removed from this PR and I have edited the text below to show those removals with as ~~strikethrough~~.

Changes made in this pull request:
 - Update support for React Native v0.59.x.
 - Update support for React v16.8.3 (which is the version required by React Native v0.59.x).
 - Update other dependencies to latest versions.
 - Update RXPTest sample to match latest React Native project template, as some changes have been made there.
 - ~~Change to the extracted [**AsyncStorage**](https://github.com/react-native-community/react-native-async-storage) package instead of version bundled with React Native. React Native deprecated the bundled version in release v0.59.0 as part of their [Lean Core](https://github.com/facebook/react-native/issues/23313) initiative.~~
 - ~~Change to the extracted [**NetInfo**](https://github.com/react-native-community/react-native-netinfo) package instead of version bundled with React Native. React Native deprecated the bundled version in release v0.59.0 as part of their [Lean Core](https://github.com/facebook/react-native/issues/23313) initiative.~~
 - Change to the extracted [**WebView**](https://github.com/react-native-community/react-native-webview) package instead of version bundled with React Native. React Native deprecated the bundled version in release v0.57.0 as part of their [Lean Core](https://github.com/facebook/react-native/issues/23313) initiative.

Note that it will be important to mention in the changelog that anyone updating to a version including the changes in this PR will need to add ~~AsyncStorage~~, ~~NetInfo~~, and WebView as dependencies and then run the following commands within their project:
```bash
react-native link react-native-webview
```

I will make sure to submit a PR to the [create-rx-app](https://github.com/a-tarasyuk/create-rx-app) tool to update the templates to match this PR after this PR has been merged and a new version published to NPM.

~~This PR is still a **WIP** because I'm having an issue with the JavaScript code that is generated by the TypeScript compiler not playing nice with the NetInfo package, causing an error to be thrown when trying to access the default NetInfo export. So this PR isn't ready to merge yet but I thought it worthwhile to open the PR so people can see my work and potentially save on duplicate work.~~

EDIT: I have not been able to get AsyncStorage or NetInfo to work yet so I have removed them from this PR for now.

PR #1092, opened 2019-05-21
Change to the extracted NetInfo package

sbeca:
This is a follow-up PR to what I started in PR #1059.

The goal of this PR is to change to the extracted [**NetInfo**](https://github.com/react-native-community/react-native-netinfo) package instead of version bundled with React Native. React Native deprecated the bundled version in release v0.59.0 as part of their [Lean Core](https://github.com/facebook/react-native/issues/23313) initiative.

This is still a work-in-progress because this PR doesn't yet support Windows. `react-native-netinfo` doesn't include Windows support and has changed enough from the code that used to be a part of React Native core that it doesn't work with the `react-native-windows` codebase. I've started work on porting and updating the NetInfo code from `react-native-windows` to `react-native-netinfo` but I'm not sure how long that will take. My progress can be tracked here: https://github.com/sbeca/react-native-netinfo/tree/windows-support

I wanted to get this PR out in the public before it was ready in case someone else was also working on the issue. I've noticed that @mikehardy posted the following issue very recently so if we're doubling up on work, it would be good to know: https://github.com/react-native-community/react-native-netinfo/issues/100


mikehardy:
@sbeca I don't think we are doubling up - I'm not doing any heavy-lifting at all, I'm more on the end-user level just trying to test other's work to verify, and smoothing rough edges while you're making the things really work - which I appreciate a lot. @matt-oakes might be interested in this though, I think of him as the goto for netinfo

src/native-common/Network.ts
```diff
@@ -7,7 +7,7 @@
  * Native implementation of network information APIs.
  */
 
-import * as RN from 'react-native';
+import * as NetInfo from '@react-native-community/netinfo';
```
berickson1: @erictraut - thoughts on pulling this out to be a reactxp module instead of inside core (as a breaking change). This will keep ReactXP from requiring module linking and make the module optional (like it now is in React Native)

erictraut: What else would we need to pull out? I'm aware of this and WebView. Is there any other functionality? If that's it, then I'd be in support of breaking out those two components into reactxp extensions. Doing that is going to be quite a bit of work though. And as you said, it will be a significant breaking change. Probably necessitating a version bump to 2.0. Are you volunteering to do it? :)

berickson1: I think all that's needed is to move the code into a new module, it shouldn't be a heavy lift in my opinion.  I agree that a major version bump would be needed
The consumer of the module would react-native link on install or manually link the react-native netinfo module.
I can carve out some time on Friday to tackle this if @sbeca would be open to collaboration on their branch

erictraut: It may be a heavier lift than you expect. We'll need to create new projects for the extensions, update all of the samples (including RXPTest), and update the documentation. If you'd prefer to leave the doc changes to me, I can take care of those.

mikehardy: netinfo is only deprecated in RN0.60 if I recall correctly, but geolocation is fully removed: `Full removal of WebView & Geolocation, you now need to rely on the extracted versions` - https://github.com/facebook/react-native/releases/tag/v0.60.0-rc.0

sbeca: I don't mind if this stays in core or is moved out to its own ReactXP module. If we do move it out into its own module, then we should consider doing the same for WebView and AsyncStorage.

And I'm happy to receive any collaboration from others on my branch.

mikehardy: Likewise, I may not add code but I'll do my best to pull branches and try things clean following updated instructions to make sure they're clean. I have test rigs for everything but windows, and I'm only missing a couple of the oldest android APIs from the full mobile device version skew covered by react-native

sbeca:
FYI, I have just submitted my PR to the NetInfo repo to add Windows support: https://github.com/react-native-community/react-native-netinfo/pull/103

If/when that PR is accepted, I'll update this PR to point to the new version and then this PR should be good to go, pending the decision to keep it in ReactXP core or move it into its own module.

matt-oakes:
The PR to add Windows support to the NetInfo module is ongoing. We'll also need to add web support for you to use it correctly with ReactXP. There is an open PR for this, but it's pretty out of date. I can try to update it soon, unless someone else wants to pick it up.

I've not used ReactXP before, but if we did add Windows and Web support to the NetInfo module, is there any need for this to be in ReactXP? The plugin code seems to just translate the NetInfo API to a different format. Would it not be better for developers to use the NetInfo module directly and they would automatically get new features when they are released? Am I missing something about how ReactXP works that means this is not possible?

deregtd:
In the past, there were several bugs and API quirks with netinfo that we abstracted away between platforms in ReactXP.  I assume some of those still exist to this day, since it's never been a Facebook goal to have uniform behavior across platforms, but I haven't looked at changes recently.

src/native-common/Network.ts
berickson1: Just pushed a full removal of RX.Network to this PR
Creates a new module reactxp-netinfo
Tested in iOS Emulator and Physical Android device

matt-oakes:
@deregtd Thanks for the reply! Can you recall what the issues were? I'm now maintaining the module outside of Facebook.

deregtd:
@matt-oakes I think some of the original nuances have been lost at this point.  I'm recalling stuff from the early days where the state wasn't available until a certain amount of time had passed, so we had to check it later on, and provide some buffering in between.  Looking at the code now, it's more straightforward, though we still shim between the older fetch and the newer promise calls to ensure compat between RN versions:

https://github.com/microsoft/reactxp/blob/master/src/native-common/Network.ts

I think it appears that the newer implementation is pretty good, cross-plat, without any of the nuance we were working around originally.

The other main thing that ReactXP brings to the table is integration with web.  So there's also the simple web abstraction of the same interface:

https://github.com/microsoft/reactxp/blob/master/src/web/Network.ts

matt-oakes:
@deregtd Thanks for the reply! I agree that web support needs to be in before it can be used with ReactXP. As mentioned before, there is an open PR for it [here](https://github.com/react-native-community/react-native-netinfo/pull/47), but it needs some work now as the code was update pretty substantially since it was opened. It mentions `react-native-web` in the PR notes, but the code doesn't seem specific to that at all. If no one else picks it up, then I can take a look at adding it in. Just to note that I don't *need* web support so it will be a little way down my TODO list.

sbeca:
FYI, my PR to `react-native-netinfo` to add Windows support has now been merged and a new version released on npm so I have just updated this PR to point to it. This means that this PR is now ready for final review.

Note, if you want to test the Windows build of the RXPTest sample, I've found using the following local changes to `samples/RXPTest/package.json` gives the best compatibility results:
```diff
-    "react": "16.8.3",
-    "react-dom": "16.8.3",
-    "react-native": "0.59.5",
+    "react": "16.6.3",
+    "react-dom": "16.6.3",
+    "react-native": "0.57.8",
```

matt-oakes:
Just to note, the NetInfo module still doesn't provide support for web. Is this an issue for you with this being merged in now?

deregtd:
Oh, yes, that's bad.  ReactXP needs to maintain support for all platforms and not regress functionality.

matt-oakes:
Sorry. I mentioned it above but should have been clearer. Do you want to revert this until web support is in NetInfo module?

erictraut:
Yes, I'll revert it.

erictraut:
OK, it's reverted. Sorry for jumping the gun on that.

berickson1:
To be fair, the current state of the extracted module is the same as reactxp core was, there are some gaps for NetInfo package.
I don't think the intention of this PR was to add functionality, but to move the package out of the core. I agree that web support should be implemented, but it might be better suited as a separate changeset

sbeca:
Yeah, this PR was fine to merge as it was because web support was still being handled by ReactXP specific code and not relying on non-existent web support in the react-native-netinfo library.

Does ReactXP want to maintain it's own netinfo library or does it want to eventually remove all NetInfo support and just recommend using the react-native-netinfo library?

PR #1094, opened 2019-05-23
Fix bugs in native Image header logic

ladipro:
Addresses two issues with the current implementation:
1. The component keeps in its state a reference to potentially old `props.headers`. It keeps passing the same object down, even after possibly having been mutated by the parent component.
2. `this.props` instead of `nextProps` is used in `componentWillReceiveProps` to compute new state.

### DIFF
--- a/src/native-common/Image.tsx
+++ b/src/native-common/Image.tsx
@@ -66,7 +66,7 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
     protected _mountedComponent: RN.Image | undefined;
     private _nativeImageWidth: number | undefined;
     private _nativeImageHeight: number | undefined;
-    readonly state: ImageState = { forceCache: false, lastNativeError: undefined, headers: this._buildHeaders() };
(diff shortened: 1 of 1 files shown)


src/native-common/Image.tsx
```diff
@@ -215,20 +215,20 @@ export class Image extends React.Component<Types.ImageProps, ImageState> impleme
             return this.props.source;
         }
 
-        const source: RN.ImageSourcePropType = { uri: this.props.source, headers: this.state.headers };
+        const source: RN.ImageSourcePropType = { uri: this.props.source, headers: this.state.headers || this.props.headers };
```
ladipro: Using `this.props.headers` if headers did not require customization. Saves us from cloning the incoming object.

berickson1: FWIW, there's no clone involved in the case where we use the value from this.props.headers, just a ref stuffed into state

ladipro: Yep, understood. What I meant is that we could clone `this.props.headers` and store the copy in state as an alternative way of fixing this.

PR #1095, opened 2019-05-29
Replace external assert dependency with internal assert

sbeca:
This is a follow-up to PR #982.

That PR replaced the external assert dependency with an internal assert function in the main ReactXP code. This PR does the same with the extensions.


sbeca:
Conflicts with master have now been fixed

PR #1098, opened 2019-05-30
[Web] FocusManager: Release reference to _setTabIndexElement

aschultz:
FocusManager was not clearing _setTabIndexElement once the queued set operation was completed or cancelled. This prevented garbage collection if the element was otherwise removed from the page.

This change releases the reference once its no longer needed.


erictraut:
Thanks for the contribution!

PR #1101, opened 2019-06-02
Extract the WebView component out of core and into an extension

sbeca:
This is a followup PR to #1059 and #1092.

In #1092 it was decided to move `RX.Network` out of the core ReactXP codebase and into an extension, because it would require depending on and linking to an external dependency. #1059 made the WebView depend on an external dependency but kept it in core ReactXP so this PR moves WebView out into an extension.

NOTE: While making this move, there were two things that I didn't know how to move that would be good to get feedback on:
1. I didn't know how to move the `WebViewStyle` and everything around that as the creation of styles is tied to some ReactXP internal functions. Seeing as `WebViewStyle` is just an extension of `ViewStyle` without adding anything extra, I have just removed it for now and just used `ViewStyle`. If this is not acceptable, please let me know how I should best include this in the extension.
2. I don't know what `RX.WebViewConstructor` is used for or how best to include it in an extension. Is this needed?


berickson1:
On a first pass, this looks good! Thanks for taking the time to do it. On Friday I'll spend a bit of time testing.
Re: questions above
1) I think the move to ViewStyle is fine, since there's no differing styles between the two
2) Looking at the code WebViewConstructor was used since WebView was defined as an interface, not an abstract class. Your changes bring the behaviour inline with other components, so I believe it's fine to remove `RX.WebViewConstructor`

extensions/webview/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "reactxp-webview",
+  "version": "0.0.1-alpha.1",
+  "description": "Plugin for ReactXP that provides a control that allows the display of an independent web page",
+  "author": "ReactXP Team <reactxp@microsoft.com>",
+  "license": "MIT",
+  "scripts": {
+    "build": "npm run tslint && tsc",
+    "tslint": "tslint -p tsconfig.json -r tslint.json -r ./node_modules/tslint-microsoft-contrib --fix || true"
```
berickson1: nit: the rules directory lookup can be included in tslint.json directly and removed from this command

berickson1:
Validated on Android - fixed one race condition in RXPTest that was causing inaccurate test results. But LGTM

sbeca:
Review nit has been resolved

sbeca:
Conflicts with master have now been fixed

PR #1106, opened 2019-06-07
Fixes #1102 Interpolated Values are not updated properly when tied to…

### LINKED ISSUE #1102: Interpolation jumps back to start value instead of animating
I've encountered a strange issue that is only appearing on the web version when interpolating a value and driving an animation.

I have created a demo of the issue in the following repo: https://github.com/alexstanbury/reactxp-interpolate-bug

Here is the relevant code:
```
import React, {useEffect, useState} from "react";
import RX from "reactxp";

let rotateValue = new RX.Animated.Value(0);

const _styles = {
        main: RX.Styles.createViewStyle({
            flex: 1,
        }),
        box: RX.Styles.createAnimatedViewStyle({
            flex: 1,
            width: 50,
            height: 50,
            backgroundColor : 'red',
            transform: [{rotate: RX.Animated.interpolate(rotateValue, [0, 1], ['0deg', '90deg'])}]
        }),

        button: RX.Styles.createButtonStyle({
            backgroundColor: 'lightgray',
            padding: 20,
            minWidth: 400,
            minHeight: 100,
        })
    }
;


const App = props => {
    const [rotate, setRotate] = useState(false);

    useEffect(() => {
        console.log('rotate', rotate);
        RX.Animated.timing(rotateValue, {
            toValue: rotate ? 1 : 0,
            duration: 2000,
            useNativeDriver: true
        }).start();
    }, [rotate]);
    return (
        <RX.View style={[_styles.main]}>

            <RX.Button style={_styles.button} onPress={() => {
                setRotate(!rotate)
            }}>
                <RX.Animated.View style={_styles.box}>
                    <RX.Text>CLICK ME</RX.Text>
                </RX.Animated.View>
            </RX.Button>
        </RX.View>
    );
};

export default App;
```

- When clicking the button, the box rotates to 90 degrees as expected. 

- If the animation is allowed to complete, then clicking the button causes the box to jump back to it's 0 position instead of animating.

- If the animation is NOT allowed to complete, clicking the button reverses the animation as expected and animates smoothly back to 0 degrees.

This does not happen on Android or iOS, it works as expected on both those platforms.

Am I missing something fundamental here? Thanks

berickson1:
… timed transformations

InterpolatedValue relies on setValue of the base AnimatedValue to determine it's iner state. Directly setting the value in _updateFinalValue prevents the InterpolatedValue from getting the correct value, so animating in the reverse direction fails

### DIFF
--- a/src/web/Animated.tsx
+++ b/src/web/Animated.tsx
@@ -193,7 +193,7 @@ export class Value extends RX.Types.AnimatedValue {
     // After an animation is stopped or completed, updates
     // the final value.
     _updateFinalValue(value: number | string) {
-        this._value = value;
+        this.setValue(value);
     }
 }
 


erictraut:
@berickson1, thanks for tracking this down. IIRC, the reason the implementation didn't call setValue is because it has the side effect of broadcasting the change to all listeners. In reviewing the code, I think it's OK to go through the setValue path, but we should make sure to test it thoroughly to make sure there aren't situations that generate an infinite feedback loop.

berickson1:
I came to the same conclusion when investigating the bug and inspecting the code. The only listener was in InterpolatedValue & not firing was the cause of the bug :)
I can do some additional testing this weekend

berickson1:
I've done some more testing both in RXPTest & with variations of the linked issue. Everything works as expected. I did some more investigation on code usage, also looks ok

PR #1109, opened 2019-06-17
Remove logging from reactxp-navigation

sbeca:
This PR removes logging that the majority of people don't need and can get in the way when trying to read logs for other issues.

If you would prefer the logging to stay then I could modify this PR to instead have the logging off by default, but could be enabled with a prop or something like that.

This PR also includes some automatic code formatting.


erictraut:
Thanks Scott, looks good to me.

Issue #1049, opened 2019-03-11
Disable user interaction with the view


pronebird:
Hi,

Is there any way to disable a user interaction with a view? `ignorePointerEvents` only applies to the specific view but then allows for pointer events in the children sub-views. I'd like to disable the entire sub-tree from receiving the clicks.

What we normally use is `pointer-events: none` but I can't apply it to the `View` because `ViewStyle` does not have such property as `pointerEvents` so I get a corresponding TypeScript error. I could of course `@ts-ignore` the `pointerEvents` property and it would even work on the web but it seems like a hack.

The reason why I need to disable the entire sub-tree is because I run animations when switching between views, that's precisely when I don't want users to interact with the views that are being animated.

By looking at the `blockPointerEvents`, it seems that it does exactly what I need but on mobile only. However, the underlying code for `blockPointerEvents` seems to actually toggle the internal `pointerEvents` flag. I wonder why not to extend the support to the web platform, given that it's about the same functionality.

erictraut:
I presume that you want to prevent not only pointer events but also touch gestures? Your solution wouldn't work with touch gestures, and it wouldn't work on native platforms in general. 

One thought is to temporarily render a transparent child view that covers the entire view (i.e. is rendered last so it's the top-most child to be rendered). It would need to use absolution positioning with left, right, top and bottom set to 0.

That approach still won't prevent keyboard interactions, but it should prevent all mouse and touch interactions.

pronebird:
>I presume that you want to prevent not only pointer events but also touch gestures? Your solution wouldn't work with touch gestures, and it wouldn't work on native platforms in general.

Correct. I'd like no user interaction to be allowed when animating the transition between two views.

>One thought is to temporarily render a transparent child view that covers the entire view (i.e. is rendered last so it's the top-most child to be rendered). It would need to use absolution positioning with left, right, top and bottom set to 0.

That's definitely a solution.

aschultz:
ignorePointerEvents appears to be implemented for both RN and web. blockPointerEvents is currently RN-only, but would be very useful to have on web.

In particular, consider a component structured like so:
`<View>
    {this.props.children}
    <Decorator />
</View>`

where Decorator is absolutely positioned on top of the child elements and fills 100% width/height, but is semi-transparent (say, a gradient). Without blockPointerEvents, there's no easy way to mark the Decorator as transparent to clicks.

[...]

PR #1117, opened 2019-07-10
Add support for blockPointerEvents on web

aschultz:
#1049

### DIFF
--- a/src/web/View.tsx
+++ b/src/web/View.tsx
@@ -66,10 +66,12 @@ const _styles = {
 
 if (typeof document !== 'undefined') {
     const ignorePointerEvents = '.reactxp-ignore-pointer-events  * { pointer-events: auto; }';
+    const blockPointerEvents = '.reactxp-block-pointer-events * { pointer-events: none !important; }';
     const head = document.head;
     const style = document.createElement('style');
     style.type = 'text/css';
     style.appendChild(document.createTextNode(ignorePointerEvents));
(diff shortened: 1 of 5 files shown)


erictraut:
Thanks for the contribution. Looks good to me. @berickson1, if this looks good to you too, please merge.

src/web/View.tsx
```diff
@@ -66,10 +66,12 @@ const _styles = {
 
 if (typeof document !== 'undefined') {
     const ignorePointerEvents = '.reactxp-ignore-pointer-events  * { pointer-events: auto; }';
+    const blockPointerEvents = '.reactxp-block-pointer-events * { pointer-events: none !important; }';
```
berickson1: Is the usage of `!important` needed here?

aschultz: Adding the ".reactxp-block-pointer-events" CSS rule to the document after the ignorePointerEvents rule may be sufficient to have a parent with it applied take precedence over any descendant with the ignorePointerEvents rule set (since both CSS rules otherwise have the same specificity).

However, its a bit fragile. The usefulness of blockPointerEvents is that you can easily apply it to a parent and make the entire subtree transparent to the mouse, regardless of any pointer-events settings on those children, and regardless of whether those children are ReactXP built-ins or external (e.g. SVG). Without !important, its a bit too easy for an inline-style or slightly more specific CSS rule to override that behavior and start making elements opaque again. At least with the !important rule, you need another !important rule to override it.

PR #1118, opened 2019-07-10
Disable drag/drop interference with GestureView panning on web

aschultz:
The mousedown event has a "[default action](https://w3c.github.io/uievents/#mousedown)" that sometimes starts a drag/drop operation, even for elements not marked as draggable. When this occurs while attempting a pan gesture, it changes the mouse cursor to a "not-allowed" icon and prevents further panning.

This change disables the default action when panning listeners are attached to GestureView so that it doesn't enter this state.


erictraut:
Thanks for the contribution. Looks good to me.

PR #1121, opened 2019-07-20
Update animations.md

kg-currenxie:
Missing colon after duration prop. Tiny annoyance when copying ;)


erictraut:
Thanks for the contribution.

PR #1124, opened 2019-07-24
fixed font style documentation

Fulanko:
Fixed documentation for font attribute shortcut.

[Related code](https://github.com/microsoft/reactxp/blob/f0082b3267f1d1d35a5f8c67ffe781afc4d02c26/src/web/Styles.ts#L273)


berickson1:
Thanks for the contribution!

Issue #1039, opened 2019-03-04
React lifecycle method deprecations


luisnaranjo733:
3 react lifecycle methods are being gradually deprecated because they cause confusion and get misused often

componentWillMount
componentWillReceiveProps
componentWillUpdate
There are 2 new lifecycle methods intended to "replace" these methods going forward

getDerivedStateFromProps
getSnapshotBeforeUpdate
Here is the migration path set out by the react team

react@16.3: Add an "UNSAFE_" alias for each of these 3 lifecycle methods (both the original method and the alias will work)
react@16.6+: We will start to get a deprecation warning if we use one of the deprecated methods listed above
react@17: Original deprecated methods will stop working. The UNSAFE alias will continue to work for the foreseeable future, but we should not use them in new code and make an effort to gradually replace them.
More info on why this happened and what it means in [this ](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html)React blog post.

Here is what I think ReactXP should do
* Short term: Find and replace the deprecated methods with the "UNSAFE_" prefixed methods. This is safe for react 16.3 and later. This will allow XP users to not get a deprecation warning from XP, and will ensure longer term support.
* Long term: Replace the UNSAFE_ methods with the new ones as proposed in issue #605

erictraut:
There has already been some discussion of this [here](https://github.com/Microsoft/reactxp/issues/605).

luisnaranjo733:
Yes, the reason I opened a new issue is that the previous discussion proposed going straight from the deprecated methods to the new cool thing (skipping the back compat layer that react provided).

I believe that the UNSAFE_* methods can be a good way to meet in the middle and support as many versions of react as possible.

erictraut:
Yeah, this sounds like a good approach.

fbartho:
Now that we're working on ReactXP 2.0 with some breaking changes re: React-Native extracting some dependencies, maybe we should try to prioritize this work?

Also, related to upgrading, react-native 60 support is going to be rough, maybe that should roll in with this work too? -- Ticket: #1123 

Thoughts?

erictraut:
That's a reasonable point. @deregtd & @berickson1 - any strong opinions on this?

deregtd:
I think this is the right approach.  We still need to figure out what to do about ReSub, though.  Moving to the deprecated names only staves off the inevitable problem over there.

fbartho:
[...]

PR #1127, opened 2019-08-09
Use new/temporary react lifecycle methods (#1039)

mikehardy:
This migration was performed using the recommend codemod tool:
`npx react-codemod rename-unsafe-lifecycles`
https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#renaming-unsafe-lifecycle-methods

The change to tslint.json should be reverted after the second phase of this change
where the deprecated/UNSAFE APIs are removed completely in favor of different lifecycle APIs

The converted code will not work on react versions < v16.3 (react-native v0.55)
The converted code will continue to work on react v17.x

As discussed in #1039 the idea is to have a transitional step to provide wider backwards-compatibility, but with a contemplated breaking change that clips off compatibility to react-native < 0.60, it may be this does not have value here, and the "real" change to the new APIs is desired. That's a much larger change cognitively though, and there was an automated tool for this migration, so I propose this for consideration

### DIFF
--- a/extensions/navigation/src/web/Navigator.tsx
+++ b/extensions/navigation/src/web/Navigator.tsx
@@ -155,7 +155,7 @@ export class NavigatorImpl extends NavigatorBase<NavigatorState> {
         };
     }
 
-    componentWillMount() {
+    UNSAFE_componentWillMount() {
         this.springSystem = new rebound.SpringSystem();
         this.spring = this.springSystem.createSpring();
         this.spring.setRestSpeedThreshold(0.05);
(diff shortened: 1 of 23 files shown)


mikehardy:
I just pushed 3 tiny commits to the RXPTest app

They all seemed necessary to run the app so I could validate these changes - @erictraut if you want I could make the RXPTest changes on a separate branch/PR as they are not really related to this change

deregtd:
@berickson1  This looks good to me.  Any objections?

PR #1140, opened 2019-09-09
fix: upgrade react-native-webview dependency, remove obsolete props

mikehardy:
react-native-webview issued a breaking change release to shed the
UIWebView references on iOS, as Apple indicated they would soon stop
accepting app submissions that used that deprecated API.

react-native-webview is now fully WebKit and without this the webview
extension here references obsolete symbols because the JS layer still has
RNCWKWebView symbols from react-native-webview <7 while the auto-linked
iOS Pod is from react-native-webview >=7

Perhaps react-native-webview should be a peerDependency?

More info: https://github.com/react-native-community/react-native-webview/releases/tag/v7.0.1

### DIFF
--- a/extensions/webview/package.json
+++ b/extensions/webview/package.json
@@ -9,7 +9,7 @@
     "tslint": "tslint -p tsconfig.json -r tslint.json --fix || true"
   },
   "dependencies": {
-    "react-native-webview": "^5.6.3"
+    "react-native-webview": "^7.0.5"
   },
   "peerDependencies": {
     "reactxp": "^2.0.0-rc.1",
--- a/extensions/webview/src/native-common/WebView.tsx
+++ b/extensions/webview/src/native-common/WebView.tsx
@@ -40,11 +40,6 @@ export class WebView extends React.Component<Types.WebViewProps, RX.Types.Statel
         const source = this._buildSource();
         const injectedJavascript = this._buildInjectedJavascript();
 
-        // Force use of webkit on iOS (applies to RN 0.57 and newer only).
-        const extendedProps: RNWebViewProps = {
-            useWebKit: true
-        };
-
         return (
             <RNWebView
                 ref={ this._onMount }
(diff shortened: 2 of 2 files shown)


mikehardy:
also worth noting this is difficult to test in place on my project because it's sort of a sub-package of the main reactxp repo, so I'm not 100% sure I did this right but I'm continuing to work on it locally as well. I'll update here and re-push if I notice anything

mikehardy:
The specific difficulty I have is that in-situ (i.e, in `<project>/node_modules/reactxp-webview`) if I make these changes then run `yarn build` I get this:

```
> reactxp-webview@2.0.0-rc.1 tslint /Users/mike/work/Kullki/ksocialscore/packages/public-app/node_modules/reactxp-webview
> tslint -p tsconfig.json -r tslint.json --fix || true

src/native-common/WebView.tsx:84:15 - error TS2742: The inferred type of '_onMessage' cannot be named without a reference to 'reactxp-webview/node_modules/react-native-webview/lib/WebViewTypes'. This is likely not portable. A type annotation is necessary.

84     protected _onMessage = (e: RNWebViewMessageEvent) => {
                 ~~~~~~~~~~


Found 1 error.

error Command failed with exit code 1.
```

In a clean checkout where I produced this PR it works though:
```
mike@isabela:~/work/react-random/reactxp/extensions/webview (webview-removal) % npm run build

> reactxp-webview@2.0.0-rc.1 build /home/mike/work/react-random/reactxp/extensions/webview
> npm run tslint && tsc


> reactxp-webview@2.0.0-rc.1 tslint /home/mike/work/react-random/reactxp/extensions/webview
> tslint -p tsconfig.json -r tslint.json --fix || true

mike@isabela:~/work/react-random/reactxp/extensions/webview (webview-removal) % 
```

So I think this change is good but I'm not 100% confident, sorry.

mikehardy:
Final comment prior to review: I am now applying the `yarn build` built results of this patch via patch-package (patch attached, just remove the .txt extension) along with a 'cd node_modules/reactxp-webview && yarn add react-native-webview@^7.0.5 && cd ../../' in my package.json::postinstall and it appears to be working now, where it was a crash prior. So I am now about 99% confident this is good. Cheers

mikehardy:
I think the devDependency in the top level package.json needs a version bump too, but before I invest any time, what do you guys think in general? 

Breadcrumb just in case https://github.com/microsoft/reactxp/blob/master/package.json#L31

PR #1141, opened 2019-09-10
feat: enable 'handled' mode of ScrollView['keyboardShouldPersistTaps']

mikehardy:
This enables tap events to go through the ScrollView and trigger onPress
on Touchable things inside the ScrollView. 

So while the keyboard is showing in a TextInput inside a ScrollView and you tap a button, the button will actually tap instead of needing one tap to dismiss keyboard, and one tap on the button

You guys have a pretty concise style and my Typescript Typing skills are not very strong yet, so if there is anything stylistic about this you don't like, just let me know.

Side note: if you disregard the package-lock.json right now (e.g. by using yarn instead of npm) current master fails to compile with tsc. I'm assuming some underlying type package moved and Button and View no longer have compatible Animated Style types. Using `npm install` does work though and after that install a `yarn build` is successful, and I can transplant the src+dist directories to my project's node_modules for patch-package consumption where I can deploy my work.

### DIFF
--- a/src/native-common/ScrollView.tsx
+++ b/src/native-common/ScrollView.tsx
@@ -11,6 +11,7 @@ import * as React from 'react';
 import * as RN from 'react-native';
 
 import * as RX from '../common/Interfaces';
+import { ScrollViewProps } from '../common/Types';
 
 import ViewBase from './ViewBase';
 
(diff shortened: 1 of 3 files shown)


src/native-common/ScrollView.tsx
```diff
@@ -82,7 +83,15 @@ export class ScrollView extends ViewBase<RX.Types.ScrollViewProps, RX.Types.Stat
             scrollHandler = undefined;
         }
 
-        const keyboardShouldPersistTaps = (overrideKeyboardShouldPersistTaps || this.props.keyboardShouldPersistTaps ? 'always' : 'never');
+        // keyboardShouldPersistTaps can be overridden, a deprecated boolean that should translate to 'always' or 'never', or 'handled'
+        let keyboardShouldPersistTaps: ScrollViewProps['keyboardShouldPersistTaps'] = 'never'; // never is the default
+        if (overrideKeyboardShouldPersistTaps ||
+            ((typeof this.props.keyboardShouldPersistTaps === 'boolean') && (this.props.keyboardShouldPersistTaps))) {
```
berickson1: This can be simplified a bit
```
if (overrideKeyboardShouldPersistTaps || this.props.keyboardShouldPersistTaps === true) {
```

src/native-common/ScrollView.tsx
```diff
@@ -82,7 +83,15 @@ export class ScrollView extends ViewBase<RX.Types.ScrollViewProps, RX.Types.Stat
...
+            keyboardShouldPersistTaps = 'always';
+        }
+        if (typeof this.props.keyboardShouldPersistTaps === 'string') {
```
berickson1: I think this should be an `else if` since overrideKeyboardShouldPersistTaps should probably still override even if a string value is provided . (not 100% confident in this though)

mikehardy: I am fairly confident I handle the various cases, but I do not like my structure there either, and any reader should feel 100% confident, so I'll take another cut at it, taking into account that I can also simplify per your note above. Thanks for the review

mikehardy:
Yay! Now we just need a reactxp-2.0.0-rc.2 and I can purge my patch-package patch :-)

PR #1163, opened 2019-11-14
Fix TodoList sample build in release mode

mikehardy:
If a build using this file was attempted in release mode, it attempted to reference variables that did not exist, and unassert() method which similarly did not exist

Yes, it is just a sample, but I finally got around to extending my app to web and as I integrated the gulp part of the build infrastructure to get the module aliasing, I noticed that release builds would fail in this area


berickson1:
Hah, good catch - some projects choose to strip client assertions in production & I guess we never cleaned this up when we adapted the gulpfile for this sample

Issue #1123, opened 2019-07-24
any known issue with upgrading to RN 0.60?


liorJuice:
is it safe to upgrade or things would break? 
as a general note, how much RN upgrades should be took with caution using reactXP ?

mikehardy:
I am doing a RN59 to RN60 upgrade right now, and yes, it will break you.

1- webview is no longer there - you will need to add `reactxp-webview` and `react-native-webview` packages to maintain webview functionality, along with related code changes
2- netinfo is no longer there, add `@react-native-community/netinfo` package and use reactxp@1.7.0-rc.2 (currently unreleased! so you have to depend on master after this PR landed #1119) or later (which handles the netinfo lean-core changes)
3- I still have some sort of javascript issue in Accessibility which is blowing up my mocha unit tests so I had to take `npm test` off my precommit hook for now

There may be other issues but other than those I do have it running on android and ios

This sounds rough but from my perspective, as disruptive as the lean-core / androidx / cocoapods changes have been in android and ios native land, this is a fairly good result. I believe I will have a working app when I'm done with my iOS project conversion - it's likely (for my project) that if current reactxp.git#master and the related reactxp-webview extension master were released as 1.7.0 and 1.0.0 respectively it's correct for my app

As a related note @erictraut - I'm not sure on intended semver fidelity in this repo, but the webview changes require code changes, shouldn't it be reactxp@2.0.0 when it releases? I'm not semver-militant myself but if the intention is semver-compliance it's definitely breaking change, and if the intention is not semver-compliance the recommendation should be `~`-style semver range specifiers vs `^`, and finally, the extensions seem to be 0.0.1 whereas a 1.0.0 would be nice. Cheers

erictraut:
@mikehardy, Yes that's a good point about versioning. We've historically been bumping the second number in the version string when there's a breaking change in the API, but I now realize that we should have been incrementing the first number. That means we should probably call the new version 2.0.0 rather than 1.7.0. @berickson1, and problem with switching to this new approach? I should probably document it as well so we're clear and consistent going forward.

And yes, the netinfo extension will get a real version number when we finally publish it. @berickson1, is this extension ready to be published as an "rc"?

mikehardy:
I want to add for @liorJuice that if you depend on unreleased versions of reactxp (as you must in order to handle the lean-core webview extraction) you have to transpile the typescript before you can use it

[...]

PR #1165, opened 2019-11-15
Changes for 2.0.0-rc.2

deregtd:
* Bumped all dependencies in ReactXP, all extensions, and all samples to latest versions using npm-check-updates
* Got RXPTest building again on web and ios.
* Fixing regression in web scrollview from a couple weeks ago
* Eliminating setMaxContentSizeMultiplier since it's not in public RN
* Fixing the image api test to take pixel ratio into account
* Updated all versions of extensions to rc.2 (or rc.3 in a few spots) and version of reactxp to rc.2


package.json
```diff
@@ -18,21 +18,21 @@
     "subscribableevent": "^1.0.1"
   },
   "peerDependencies": {
-    "react": "^16.0",
-    "react-dom": "^16.0",
+    "react": "^16.3",
+    "react-dom": "^16.3",
     "react-native": ">=0.57 <0.60",
```
mikehardy: My only question about this PR (and the release path to 2.x) is why not loose the react-native dependency band to 0.60+ (I'm using it on 0.61 without issue and don't see anything in 0.62 that will break it now that netinfo and webview are standing up nicely with current lean-core extracted repos)

I see the peerDep set to <0.60 for the main module here and all the extensions and it seems like noise in the package install process when they seemingly work fine up to 0.61 at minimum

This is a separate statement from "the samples should be upgraded to RN0.6x" as that's a not fun chore but is orthogonal from the peerDependency setting I think

mikehardy:
Do I smell a release coming :-) ? FWIW I'm effectively using master (2.0.0-rc.1 plus a patch-package patch built from the diff to master) right now with RN 0.61-current as well as a clean 'yarn outdated' for a web build and it all seems fine. 👍

package.json
deregtd: Good question... I coulda sworn there was some work left to do to make it compatible with 0.60+ but maybe it was just the samples that needed the chore and we're good.

deregtd:
> Do I smell a release coming :-) ? FWIW I'm effectively using master (2.0.0-rc.1 plus a patch-package patch built from the diff to master) right now with RN 0.61-current as well as a clean 'yarn outdated' for a web build and it all seems fine. 👍

Thanks much for trying this out!  Appreciate the validation. :)

package.json
mikehardy: This was my list as I brought my work project up on it - seemed like it was just non-blocking things now (React lifecycle items etc) https://github.com/microsoft/reactxp/issues/1123#issuecomment-520096927

mikehardy:
> Thanks much for trying this out! Appreciate the validation. :)

Integrated rc2 this morning via regular install and was able to evict my patch-package patch. I'm no longer carrying any reactxp-related patches with the current RCs (including ReSub 2.0.0-rc2) with react-native current stable 0.61.4 and I confirm it all seems to work. Have a great weekend

deregtd:
Hoping for a little more testing and feedback from others, then we can do a 2.0.0 final release of everything (reactxp and resub, at least.)

PR #1169, opened 2019-12-08
Replace TSLint on ESLint

a-tarasyuk:
cc @deregtd


src/native-common/ScrollView.tsx
```diff
@@ -20,7 +20,7 @@ import ViewBase from './ViewBase';
 //   order for the UI to acknowledge your interaction.
 const overrideKeyboardShouldPersistTaps = RN.Platform.OS === 'macos' || RN.Platform.OS === 'windows';
 export class ScrollView extends ViewBase<RX.Types.ScrollViewProps, RX.Types.Stateless, RN.ScrollView, RX.ScrollView>
-        implements RX.ScrollView {
```
deregtd: This one's a little weird -- I'd expect this to keep the double-indent since it's a continuation of a line that has encapsulation, just like a wrapped function signature.

src/native-common/UserInterface.tsx
```diff
@@ -34,7 +34,7 @@ export class UserInterface extends RX.UserInterface {
     }
 
     measureLayoutRelativeToWindow(component: React.Component<any, any>):
-        Promise<RX.Types.LayoutInfo> {
+    Promise<RX.Types.LayoutInfo> {
```
deregtd: Same here -- I'd expect this to be double-indented, not 0-indented

src/native-common/UserInterface.tsx
```diff
@@ -47,16 +47,16 @@ export class UserInterface extends RX.UserInterface {
...
-            }
+            },
         );
 
         return deferred.promise();
     }
 
     measureLayoutRelativeToAncestor(component: React.Component<any, any>,
-        ancestor: React.Component<any, any>): Promise<RX.Types.LayoutInfo> {
+            ancestor: React.Component<any, any>): Promise<RX.Types.LayoutInfo> {
```
deregtd: yeah -- exactly like this!

src/native-common/ViewBase.tsx
```diff
@@ -14,7 +14,7 @@ import * as RX from '../common/Interfaces';
 import { isEqual } from './utils/lodashMini';
 
 export abstract class ViewBase<P extends RX.Types.ViewPropsShared<C>, S, T extends RN.View | RN.ScrollView,
-        C extends RX.View | RX.ScrollView> extends RX.ViewBase<P, S> {
+    C extends RX.View | RX.ScrollView> extends RX.ViewBase<P, S> {
```
deregtd: another indent oddity

src/native-common/ScrollView.tsx
a-tarasyuk: @deregtd I'll check maybe need to add changes to `@typescript-eslint/indent` to handle spaces for `implements`.

PR #1184, opened 2020-02-07
Update ScrollView docs

berickson1:
Update to reflect changes in https://github.com/microsoft/reactxp/commit/1a924ac6cd26b0c21eb1c9bbccee8d41241392bd


deregtd:
Only nit: Probably should describe the behavior of horizontal = undefined (is that vertical? both? none?) and then explicitly say what happens when horizontal = true (vertical disabled, etc.)

PR #1200, opened 2020-04-22
Add support for React Native's maxFontSizeMultiplier Text prop

sbeca:
This PR adds support for React Native's [maxFontSizeMultiplier](https://reactnative.dev/docs/text#maxfontsizemultiplier) Text prop, which I need for some work that I'm doing.

ReactXP already has support for the [allowFontScaling](https://reactnative.dev/docs/text#allowfontscaling) Text prop and these two props would be often used together so I feel like having them both available in ReactXP makes sense.

### DIFF
--- a/docs/docs/components/link.md
+++ b/docs/docs/components/link.md
@@ -24,6 +24,12 @@ allowFontScaling: boolean = true; // Android and iOS only
 // property.
 autoFocus: boolean = false;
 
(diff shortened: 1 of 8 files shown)


fbartho:
@sbeca -- I know this PR has lingered a little bit since you made it, are there any additional tweaks you want to make before I review this? I'm looking to make a release in Early September!

sbeca:
Hi @fbartho, this PR should still be good to review and add. I have used this code within an app using a fork and it works well.

PR #1235, opened 2020-08-15
Text selectable

Fulanko:
Added missing Text selection prop for Android.

On iOS, there is no highlighting and partial text selections are not possible (same as react-native implementation)


erictraut:
Looks good to me. Does this need to be added within the RX.Link implementation as well?

Fulanko:
I think selection on RX.Link is only implemented for Web right now.

PR #1248, opened 2020-10-16
Updating readme as it is misleading if using cocoapods. Better to fol…

NickLatin0:
…low the instructions on package getting started page.


mikehardy (commented):
I love the direction of this change as manual linking is so 2019 ;-), my only hesitation is that this then continues the risky (could go out of date) behavior be deep-linking in the external repo. Perhaps best to just link to the repo itself with a note to "follow that module's installation instructions" as a vague generic thing?

fbartho (approved):
Thanks @NickLatin0 -- that's a good catch.

Issue #1251, opened 2020-12-07
Scrollview with buttons in Chromebook triggers an "Intervention" warning


dlgski:
Using touch events on a Chromebook to scroll a view that has buttons inside it, often triggers the button "onPress" event when it should be ignored due to scrolling.  

A warning "Intervention" error is displayed in the console stating:

```
[Intervention] Ignored attempt to cancel a touchend event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.
```

This is pointing to the touchEnd handling in Button.js. 

To fix, seems like just an additional check on e.cancelable is all that is needed here before executing the press event methods.

fbartho:
That sounds like a good fix!

dlgski:
In `src/web/Button.tsx`  line 282 in the `_onTouchEnd` method, add a check for `e.cancelable`:

` if ('touches' in e && e.cancelable) { `

PR #1252, opened 2020-12-07
make sure cancelable is true before calling onPress for touchEnd event

dlgski:
Issue #1251

When you have buttons within a ScrollView, and on a Chromebook you press on one of those buttons and drag your finger up or down to scroll.  It should not trigger the onPress event. 

What is currently happening is the press event is triggered and an Intervention warning is displayed in the console:

`[Intervention] Ignored attempt to cancel a touchend event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.`

What needs to done to fix this is to make sure onTouchEnd the event is cancelable before triggering the onPress event actions.  

This PR simply adds this check.

### DIFF
--- a/src/web/Button.tsx
+++ b/src/web/Button.tsx
@@ -279,7 +279,7 @@ export class Button extends ButtonBase {
             /* 3 */
             this._isMouseOver && !this._ignoreTouchEnd
         ) {
-            if ('touches' in e) {
+            if ('touches' in e && e.cancelable) {
                 // Stop the to event sequence to prevent trigger button.onMouseDown
                 e.preventDefault();
                 if (this.props.onPress) {


fbartho (approved):
This only resolves the console warning, right? There's no user-visible behavior changes other than that?

dlgski:
It fixes the conflict of scrolling and pressing on a button at the same time.  So if you press & scroll the press action should not occur.  Currently the press action does happen and this Intervention error happens.  With this in place, the press action will be suppressed if you are scrolling and no Intervention warning is displayed.

dlgski:
hi @fbartho do you need anything more for this  PR?  Is there a plan to get it merged? ty!

tahins (approved):

Issue #1257, opened 2021-02-16
How to bring popups into the same context as the rest of the app?


anchann:
ReactXP seems to have this sort of top level component layout:

```
RootView
    MainViewContainer
        the top level component for your application goes here, e.g. <App/>
    PopupContainerView
        individual popups go here
```

My App has the following layout:

```
<BrowserRouter>
    ... actual content that depends on current route, etc ...
</BrowserRouter>
```

The BrowserRoute component (from react-router) internally uses react's context infrastructure to make routing/navigation primitives accessible to child components deep down the tree, without explicitly passing them around as props.

If I want to display a popup which has a link for navigating to a different route, the `<Link>` or any routing related component relies on being a descendent of the original `<BrowserRouter>` component which defines the shared context. With the ReactXP setup, however, popups are rendered outside of the `<App>`, and thus don't have access to the shared context established by `<BrowserRouter>`.

I tried to search for a way to wrap the whole RootView in a context, or somehow inject additional context into PopupContainerView, but could not find any APIs that would allow me to do this. What is the recommended way for handling this issue?

Thank you!

anchann:
Is there anything I can do in order to get a response on this issue? Sorry if the issue somehow malformed...

anchann:
I've implemented a way to provide a wrapper function to achieve the desired behaviour. Feel free to take a look:

https://github.com/microsoft/reactxp/pull/1264

PR #1264, opened 2021-03-17
Allow user to wrap the whole app in a context

anchann:
This enables one to use context based features, like react router,
in popups and dialogs.

NOTE: I have NOT TESTED the native changes, so this PR is for discussion purposes only.


anchann:
this is for issue https://github.com/microsoft/reactxp/issues/1257

